PR #2289, opened 2015-01-09
Delete button for entity when you're admin.

fdlk:
Deletes the current entity and reloads the data explorer.

### DIFF
--- a/molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
+++ b/molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
@@ -380,6 +380,12 @@ function($, molgenis, settingsXhr) {
 			$(document).trigger('removeAttributeFilter', {'attributeUri': $(this).data('href')});
 		});
 		
+		$('#delete').on('click', function(e){
+			$.ajax('/api/v1/'+selectedEntityMetaData.name+'/meta', {'type': 'DELETE'}).done(function(){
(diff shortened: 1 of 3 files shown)


molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
```diff
@@ -380,6 +380,12 @@ function($, molgenis, settingsXhr) {
 			$(document).trigger('removeAttributeFilter', {'attributeUri': $(this).data('href')});
 		});
 		
+		$('#delete').on('click', function(e){
+			$.ajax('/api/v1/'+selectedEntityMetaData.name+'/meta', {'type': 'DELETE'}).done(function(){
```
dennishendriksen: How about adding a confirm dialog that says "Are you sure you want to delete all data and meta data for entity <entity>"?

molgenis-dataexplorer/src/main/resources/templates/view-dataexplorer.ftl
```diff
@@ -51,6 +51,13 @@
                     </div>
                 </div>
             </div>
+            <#if isAdmin?has_content && isAdmin>
+	            <div class="row">
+	            	<div class="col-md-1">
+	            		<a id="delete" class="btn btn-danger">delete</a>
```
dennishendriksen: Delete instead of delete

PR #2292, opened 2015-01-09
New class that can be used to ping online services to see whether they are online. Fix for annotator copy function where sometimes the attribute parts of a compound attribute were never added.

Mark-de-Haan:


Mark-de-Haan:
Fixed broken test

PR #2324, opened 2015-01-16
added transactional annotation to method

georgebyelas:


molgenis-compute-ui/src/main/java/org/molgenis/compute/ui/model/JobStatus.java
```diff
@@ -23,7 +23,8 @@ public boolean isFinalState()
 		@Override
 		public EnumSet<JobStatus> getStateTransitions()
 		{
-			return EnumSet.of(SCHEDULED, RUNNING);
+			//FIXME temp
+			return EnumSet.of(SCHEDULED, RUNNING, FAILED, COMPLETED, CANCELLED);
```
dennishendriksen: Is this temporary fix still required after fixing the transactional issue?

molgenis-compute-ui/src/main/java/org/molgenis/compute/ui/model/JobStatus.java
```diff
@@ -41,9 +42,10 @@ public boolean isFinalState()
 		}
 
 		@Override
+		//FIXME temp
 		public EnumSet<JobStatus> getStateTransitions()
 		{
-			return EnumSet.of(RUNNING);
+			return EnumSet.of(RUNNING, COMPLETED, FAILED, CANCELLED);
```
dennishendriksen: Is this temporary fix still required after fixing the transactional issue?

molgenis-compute-ui/src/main/java/org/molgenis/compute/ui/job/JobServiceImpl.java
```diff
@@ -37,7 +39,8 @@ public void updateJobStatus(JobStatusUpdate statusUpdate)
 		}
 
 		job.setStatus(statusUpdate.getStatus());
-		job.setOutputMessage(statusUpdate.getOutputMessage());
+//		job.setOutputMessage(statusUpdate.getOutputMessage());
```
dennishendriksen: Please add a note why you disabled this line.

georgebyelas: Now, the message is too large for the field in database

molgenis-compute-ui/src/main/java/org/molgenis/compute/ui/model/JobStatus.java
georgebyelas: status transition from RUNNING to CREATED is not allowed. Allowed transitions [COMPLETED, FAILED, CANCELLED]

molgenis-compute-ui/src/main/java/org/molgenis/compute/ui/model/JobStatus.java
georgebyelas: status transition from RUNNING to CREATED is not allowed. Allowed transitions [COMPLETED, FAILED, CANCELLED]

pom.xml
```diff
@@ -290,7 +290,7 @@
                     <artifactId>jetty-maven-plugin</artifactId>
                     <version>8.1.12.v20130726</version>
                     <configuration>
-                        <stopPort>${jetty.stop.http.port}</stopPort>
+                        <stopPort>12345</stopPort>
```
dennishendriksen: Please revert change

PR #2328, opened 2015-01-16
Fix #2033 Entity form saves false boolean properties as NULL values

### LINKED ISSUE #2033: Entity form saves false boolean properties as NULL values.
An entity with a boolean attribute with value false gets saved as value NULL by the entity form, probably since the form cannot differentiate between false and NULL values.
I'd argue that the original value of the attribute should be used.

jjettenn:
Fixed: Entity form saves false boolean properties as NULL values. #2033


molgenis-core-ui/src/main/resources/js/jquery.molgenis.table.js
```diff
@@ -407,16 +406,10 @@
                 }
 				break;
             case 'BOOL':
-            	// FIXME refactor formatTableCellValue to accept attribute instead of attribute field type and move nillable boolean code to this function
```
fdlk: This suggestion, to send the entire attribute to the formatTableCellValue instead, sounds a bit nicer to me. Did you consider it?

PR #2333, opened 2015-01-19
Introduced a new algorithm to post process the matched ontology terms

ChaoPang:
The reason is that there are cases where matching term is mapped to multiple synonyms of the same term at the same time leading to a low final score. E.g. protruding eyeball (exophthalmos, proptosis) is mapped to 'protruding eye' by 30%, 'exophthalmos' by 30% and 'proptosis' by 26%. For such cases, the score should re-calculated by combining multiple synonyms together in order to give a fair score


ChaoPang:
Will re-open in the future, because the new algorithm probably does not improve the results

Issue #2321, opened 2015-01-15
language errors in molgenis1.2.2


jjettenn:
1. After successful registration
Message is: 
Success! You have successfully registered, an activation e-mail has been send to your email. 
Should be: Success! You have successfully registered, an activation email has been sent to your email.

fdlk:
This is fixed in the master.

jjettenn:
Fixed

Issue #2318, opened 2015-01-15
Removing in Submit Study Request modal does not work


fdlk:
Log in.
Go to catalog plugin.
Select a couple folders.
Push submit.
Enter a title and add a file
Click one (or two) remove crosses in the dialog
Push Submit.

Unexpected result:
Dialog remains open.
An error box pops up on the main page:
Error! An error occurred. Please contact the administrator.
SyntaxError: JSON Parse error: Unrecognized token '<'

What goes wrong: /cart/remove does not recognize the request, and reports the error in HTML, rather than JSON and then the response gits parsed badly.

PR #2335, opened 2015-01-19
Fixed: Removing in Submit Study Request modal does not work #2318

jjettenn:
Removing in Submit Study Request modal does not work #2318
AND
Edit New issue language errors in molgenis1.2.2 #2321

### DIFF
--- a/molgenis-security/src/main/resources/templates/resetpassword-modal.ftl
+++ b/molgenis-security/src/main/resources/templates/resetpassword-modal.ftl
@@ -70,7 +70,7 @@
 		            url:  '/account/password/reset',
 		            data: form.serialize(),
 		            success: function () {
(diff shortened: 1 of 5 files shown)


molgenis-security/src/main/resources/templates/resetpassword-modal.ftl
```diff
@@ -70,7 +70,7 @@
 		            url:  '/account/password/reset',
 		            data: form.serialize(),
 		            success: function () {
-		            	$(document).trigger('molgenis-passwordresetted', 'Password resetted, an email has been send to you');
+		            	$(document).trigger('molgenis-passwordresetted', 'Password resetted, an email has been sent to you');
```
fdlk: resetted is incorrect too.

PR #2336, opened 2015-01-19
Use iframe uploader to upload study file in order item form. 

fdlk:
Fixes https://github.com/molgenis/molgenis-lifelines/issues/128
Uses https://github.com/cmlenz/jquery-iframe-transport to upload the file with the study data request.

Also: language change: resetted -> reset

### DIFF
--- a/molgenis-omx-protocolviewer/src/main/java/org/molgenis/omx/protocolviewer/ProtocolViewerController.java
+++ b/molgenis-omx-protocolviewer/src/main/java/org/molgenis/omx/protocolviewer/ProtocolViewerController.java
@@ -247,12 +247,22 @@ public class ProtocolViewerController extends MolgenisPluginController
 	// Spring's StandardServletMultipartResolver can't bind a RequestBody or
 	// ModelAttribute
 	@RequestMapping(value = "/order", method = RequestMethod.POST, headers = "Content-Type=multipart/form-data")
-	@ResponseStatus(HttpStatus.NO_CONTENT)
-	public void orderData(@RequestParam Integer catalogId, @RequestParam String name, @RequestParam Part file)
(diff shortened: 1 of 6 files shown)


molgenis-omx-protocolviewer/src/main/java/org/molgenis/omx/protocolviewer/ProtocolViewerController.java
```diff
@@ -247,12 +247,22 @@ public String getOrderDataForm()
...
 	{
 		if (!getEnableOrderAction()) throw new MolgenisDataAccessException("Action not allowed");
-		protocolViewerService.submitStudyDefinitionDraftForCurrentUser(name, file, catalogId.toString());
+		try {
+			protocolViewerService.submitStudyDefinitionDraftForCurrentUser(name, file, catalogId.toString());
+			ModelAndView model = new ModelAndView("order-response");
+			model.addObject("ok", true);
+			model.addObject("message", "Your submission has been received.");
+			return model;
+		} catch (Exception ex){
```
dennishendriksen: did you mean throw RuntimeException ex?

fdlk: The IFrame solution isn't able to access the HTTP status codes, and will always call the complete() callback. So the idea was to put the response in a success message instead.

fdlk: You're absolutely right! thanks!

fdlk: Will update to make clearer.

PR #2340, opened 2015-01-21
Disease Matcher - more bug fixes

tommydeboer:
Added onDataChange setting to molgenis.table 
Changed custom warning system to molgenis.createAlert

Bug fixes:
- When applying a filter with 0 results, the info of the last selected gene/disease stays visible
- Using the search box in data explorer fires the queryChanged event twice
- Clinical synopsis not rendered correctly
- 'PhenoTips offline’ message when actually input was wrong (added input check)
- Data Explorer table doesn’t update when values are edited in DM, and vice versa
- Download button not working in Filter menu (page not available)
- Some Text links point to OMIM pages that don’t exist
- PhenoTips filtered dataset has results even when nothing in query
- “No genes found” when in Disease list menu
- When refreshing the page with DM open: pager not showing + error (undefined is not a function)

### DIFF
--- a/molgenis-dataexplorer/src/main/resources/js/dataexplorer-diseasematcher.js
+++ b/molgenis-dataexplorer/src/main/resources/js/dataexplorer-diseasematcher.js
@@ -53,11 +53,18 @@
 	
 	var phenotipsFilteredQuery;
 	
+	var currentQuery;
+	
 	/**
 	 * Listens for data explorer query changes and updates the selected gene/disease lists.
 	 */
 	$(document).on('changeQuery', function(e, query) {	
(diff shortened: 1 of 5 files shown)


molgenis-dataexplorer/src/main/resources/js/dataexplorer-diseasematcher.js
```diff
@@ -447,8 +467,16 @@
 		$('#btn-filter-phenotips-complete').click(function(e){
 			e.preventDefault();
 			var terms = $('#hpoTermsInput').val();
```
dennishendriksen: improvement: use jquery validation rule instead of regex in js

PR #2344, opened 2015-01-22
Footer now editable via RuntimeProperty

tommydeboer:


molgenis-core-ui/src/main/java/org/molgenis/ui/MolgenisPluginInterceptor.java
```diff
@@ -53,6 +57,7 @@ public void postHandle(HttpServletRequest request, HttpServletResponse response,
 			{
 				modelAndView.addObject(KEY_PLUGIN_ID, molgenisPlugin.getId());
 			}
+			modelAndView.addObject("footerText", molgenisSettings.getProperty("molgenis.footer"));
```
dennishendriksen: use KEY_FOOTER instead of molgenis.footer

PR #2347, opened 2015-01-26
stabilize poms

fdlk:


pom.xml
```diff
@@ -401,7 +398,7 @@
                                         </goals>
                                     </pluginExecutionFilter>
                                     <action>
-                                        <ignore/>
+                                        <ignore />
```
Mark-de-Haan: Fix these, only occuring in this file

PR #2376, opened 2015-02-02
Feature/biobankconnect WIP

fdlk:


molgenis-core-ui/src/main/java/org/molgenis/ui/MolgenisPluginInterceptor.java
```diff
@@ -60,7 +60,8 @@ public void postHandle(HttpServletRequest request, HttpServletResponse response,
 			{
 				modelAndView.addObject(KEY_PLUGIN_ID, molgenisPlugin.getId());
 			}
-			
+
+
```
Mark-de-Haan: Double whitespace

PR #2445, opened 2015-02-09
new relations for UML imports

jmuilu:
New relations added, which are needed for importing UML models.

### DIFF
--- a/molgenis-data/src/main/java/org/molgenis/data/semantic/Relation.java
+++ b/molgenis-data/src/main/java/org/molgenis/data/semantic/Relation.java
@@ -30,7 +30,8 @@ public enum Relation {
 
 	public static Relation forIRI(String string) {
 		for (Relation relation : values()) {
-			if (relation.getIRI().equals(string)) {
+			if (relation.getIRI().equals(string))
+			{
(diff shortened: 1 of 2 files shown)


molgenis-data/src/main/java/org/molgenis/data/semantic/Relation.java
```diff
@@ -1,34 +1,36 @@
...
+			"http://molgenis.org/uml/isRealizationOf"), isGeneralizationOf(
+			"http://molgenis.org/uml/isGeneralizationOf"), hasSourceId(
+			"http://molgenis.org/uml/hasSourceId"), hasSourceName(
+			"http://molgenis.org/uml/hasSourceName");
+	
 	private String iri;
 
-	private Relation(String iri)
-	{
+	private Relation(String iri) {
```
fdlk: Can you fix the formatting? See https://github.com/molgenis/molgenis/blob/master/eclipse-java-molgenis-style.xml for the standard settings.

jmuilu: Sure. Done now !

Issue #2490, opened 2015-02-18
All menu-items from the same plugin are active  


jjettenn:
When having multiple menu items with the same plugin_id, by selecting one all are selected.

Reproduce:

Create one menu item with the next configurations:
plugin: dataexplorer
name: MenuItem1
query string: entity=entityNameOne (een existing entity name)

Create a second  menu item with the next configurations:
plugin: dataexplorer
name: MenuItem2
query string: entity=entityNameOne (een existing entity name)

Go to the homepage and then select the MenuItem1.

Result: both MenuItem1 and MenuItem2 are selected in the menu bar

PR #2491, opened 2015-02-18
Feature/itemplugin Fix related to: All menu-items from the same plugin are active #2490

jjettenn:
Fix related to: All menu-items from the same plugin are active #2490


molgenis-core-ui/src/test/java/org/molgenis/ui/MolgenisPluginInterceptorTest.java
```diff
@@ -117,6 +119,32 @@ public void postHandle_pluginIdExists() throws Exception
 	}
 
 	@Test
+	public void postHandle_pluginid_with_query_string() throws Exception
```
Mark-de-Haan: method names should be camelcase not with '_'

PR #2500, opened 2015-02-19
R Api fix suggestions. URL encoding fix. Fixes #1774

### LINKED ISSUE #1774: RestController breaks when idAttribute needs URLEncoding
`new URI(null, DataConverter.toString(entity.getIdValue()), null)` throws an exception when the idValue contains a character that needs encoding.

`new URI(null, URLEncoder.encode(DataConverter.toString(entity.getIdValue()), "UTF-8"), null)` fixes this, but I can imagine this brings back the original issue for which this got changed?

Mark-de-Haan:


mswertz:
I would expect typing to be automagic

Mark-de-Haan:
The R read.csv will infer its own types on columns. For example an Allele 'T' will be seen as TRUE by R. This can be prevented by setting a ColClass for every column.

mswertz:
Is this an end-user choice?

Mark-de-Haan:
Hmm no, the EMX that users upload is their definition of types.
So we should just use the molgenis metadata for typing

If someone gets an entity that is not uploaded by themselves they should
accept the fact that we assign the typings. They can always change it
within their R script once they have the dataframe locally

On 19 February 2015 at 17:42, Morris Swertz <notifications@github.com>
wrote:

> Is this an end-user choice?
> On 19 Feb 2015 17:42, "Mark de Haan" <notifications@github.com> wrote:
>
> > The R read.csv will infer its own types on columns. For example an Allele
> > 'T' will be seen as TRUE by R. This can be prevented by setting a
> ColClass
> > for every column.
> >
> > --
> > Reply to this email directly or view it on GitHub
> > <https://github.com/molgenis/molgenis/pull/2500#issuecomment-75086488>.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/molgenis/molgenis/pull/2500#issuecomment-75086800>.
>



-- 
Mark de Haan
Genomics Coordination Centre
Genetics Department
University Medical Centre Groningen

Issue #2455, opened 2015-02-10
A deleted entity is not removed from the menu


djvanenckevort:
When I delete an entity this entity is not removed from the menu. When I try to select the entity in the menu I get an Apache Tomcat 500 error page.

dennishendriksen:
Fixed

PR #2502, opened 2015-02-20
Fix for #2455 (Entities menu not updating when Entity is deleted)

tommydeboer:
Fixed Entities menu not updating when an Entity is deleted

### DIFF
--- a/molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
+++ b/molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
@@ -16,7 +16,6 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.ResourceBundle;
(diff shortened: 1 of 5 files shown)


molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
```diff
@@ -391,6 +399,31 @@ private boolean isGenomeBrowserEntity(EntityMetaData entityMetaData)
...
+	/**
+	 * Updates the 'Entities' menu when an entity is deleted.
+	 * 
+	 * @param datasetName
+	 */
+	@RequestMapping(value = "/removeEntityFromMenu/{entityName}", method = POST)
+	public void updateMenuOnDeleteEntity(@PathVariable("entityName") String entityName, HttpServletResponse response)
+	{
+		Menu menu = menuManager.getMenu();
+		MenuItem mEntities = menu.findMenuItem("entities");
```
dennishendriksen: Depending on the app, the form plugin for the requested entity might not be in this menu. Solution is to call menu.findMenuItem with the id of the form plugin for this entity.

molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
```diff
@@ -391,6 +399,31 @@ private boolean isGenomeBrowserEntity(EntityMetaData entityMetaData)
...
+
+		List<MenuItem> items = mEntities.getItems();
+		List<MenuItem> newItems = Lists.newArrayList();
+		for (MenuItem i : items)
+		{
+			if (!i.getLabel().equals(entityName))
```
dennishendriksen: User might rename the plugin thus changing the label, check for id instead of label.

PR #2503, opened 2015-02-20
working with new expanded parameters file (import/export and data model changes)

georgebyelas:


molgenis-compute-ui/src/main/java/org/molgenis/compute/ui/workflow/WorkflowImportService.java
```diff
@@ -206,20 +193,50 @@ private boolean workflowExists(String name)
 
 		List<UIWorkflowParameter> params = Lists.newArrayList();
 		CsvRepository csv = new CsvRepository(f, null);
-		try
+		Iterator<Entity> it = csv.iterator();
+
+		Hashtable<String, List<String>> tmp = new Hashtable<String, List<String>>();
```
dennishendriksen: Map<String, List<String>> ... = new HashMap<...

molgenis-compute-ui/src/main/java/org/molgenis/compute/ui/workflow/WorkflowImportService.java
```diff
@@ -206,20 +193,50 @@ private boolean workflowExists(String name)
...
+
+		while(it.hasNext())
```
dennishendriksen: for(Entity e : csv)

PR #2507, opened 2015-02-23
Feature/pathways

fdlk:
Update to 1.5 and use REST api instead of soap.


Mark-de-Haan:
Issues:
- Some pathways give me an infinite loading spinner, like 'Gap junction trafficking and regulation'
- When I import a VCF file, this is not shown in the dropdown

PR #2573, opened 2015-03-04
Added exception messages and added '()' support when using complicated q...

brammiedema:
...uery statements, changed naming in request and response so these are more explicit. Fixed minor bugs in response naming the type of query that was performed e.g allele or coordinate.


molgenis-vkgl/src/main/java/org/molgenis/vkgl/api/VkglRest.java
```diff
@@ -406,18 +453,22 @@ else if (qEl.equals("|"))
...
 								q = q.unnest();
 							}
 						}
+
 					}
 				}
 
 			}
-			
+			System.out.println(q);
```
tommydeboer: Remove println

PR #2605, opened 2015-03-10
Fix #2581 Cannot import separate EMX data file

### LINKED ISSUE #2581: Cannot import separate EMX data file
Reproduce: Split emx_all_datatypes in one file with meta data and one file with data. 1) Import meta data and then 2) Import data. Expected: Data import is successful. Actual: Data import fails with 'no importer found for' error.

dennishendriksen:


molgenis-data-mysql/src/main/java/org/molgenis/data/importer/EmxImportService.java
```diff
@@ -49,12 +51,28 @@ public boolean canImport(File file, RepositoryCollection source)
...
 		}
 
 		return false;
 	}
 
+	private boolean canImportByHeuristic(String entityName)
+	{
+		// entity is importable if entity name is the simple name of an existing entity,
+		// and only one entity with this simple name exists
+		List<EntityMetaData> entityMetaDatas = new ArrayList<EntityMetaData>();
```
erwinwinder: You can use the diamant operator here: List<EntityMetaData> entityMetaDatas = new ArrayList<>();

PR #2626, opened 2015-03-11
fix findbugs warnings and fix imports for snpEffAnnotator

bartcharbon:


bartcharbon:
Added the fix for a nullpointer and applied the eclipse code formatter

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/impl/SnpEffServiceAnnotator.java
```diff
@@ -121,18 +133,22 @@ public String getDescription()
...
+			if (snpEffpath.exists() && snpEffpath.isFile())
+			{
+				LOG.info("SnpEff found at: " + snpEffpath.getAbsolutePath());
+				result = true;
+			}
+			else
+			{
+				LOG.error("SnpEff not found at: " + snpEffpath.getAbsolutePath());
+			}
 		}
```
fdlk: maybe log that the property is missing?

PR #2630, opened 2015-03-11
Fix bug for adding compound attributes, created util class for annotators, eclipse formatted all annotatos

bartcharbon:


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/AbstractRepositoryAnnotator.java
```diff
@@ -1,8 +1,10 @@
 package org.molgenis.data.annotation;
 
+import org.molgenis.MolgenisFieldTypes;
 import org.molgenis.data.AttributeMetaData;
 import org.molgenis.data.Entity;
 import org.molgenis.data.EntityMetaData;
+import org.molgenis.data.support.DefaultAttributeMetaData;
```
Mark-de-Haan: Code was only removed, but imports were added?

PR #2636, opened 2015-03-13
Fix #2582 and #2645. Make mrefs preserve the order of the referenced entities. 

### LINKED ISSUE #2582: attribute metadata order gets shuffled when the metadata repositories get reindexed
When you throw away the index and restart molgenis, the attributes in an entity get shuffled around.
This is cause attribute references entity instead of the other way around.

fdlk:
This was nontrivial, but I think it turned out quite nice. Could use some feedback now.

* Turns around the xref that used to lie from attribute to entity. Now it is an mref from entity to attributes.
* Does the same for the compound attribute relation, now the compound attribute has an mref to its attribute children

So from old") to new")
* Adds an order column to the mref tables in mysql.
* Changes the DefaultEntity's getEntities() method to sort the retrieved entities into the order that their IDs are stored in.
* Makes the MySQL tests clean up after themselves so that errors will be reported in the test that makes them


molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntity.java
```diff
@@ -275,7 +276,15 @@ public Entity getEntity(String attributeName)
 		EntityMetaData ref = attribute.getRefEntity();
 		ids = ids.stream().map(attribute.getDataType()::convert).collect(Collectors.toList());
 
-		return dataService.findAll(ref.getName(), IN(ref.getIdAttribute().getName(), ids));
+		String entityIdName = ref.getIdAttribute().getName();
+		Map<Object, Entity> entities = new HashMap<Object, Entity>();
```
dennishendriksen: Fails for entities that refer to large number of other entities due to in memory storage, method should stream results back to caller.

molgenis-app/src/main/java/org/molgenis/migrate/MysqlMigrate.java
```diff
@@ -0,0 +1,79 @@
...
+				}
+			}
+		}
+		LOG.info("Migrating backend {} done.", backend);
+	}
+
+	private static String getMrefExistenceSql(EntityMetaData emd, AttributeMetaData att)
+	{
+		return String.format(
+				"SELECT COUNT(*) FROM INFORMATION_SCHEMA.COLUMNS WHERE table_name = '%s_%s' AND column_name = 'order'",
```
dennishendriksen: Can an entity name contain underscores? In that case this select might fail.

dennishendriksen:
Regarding RunAsSystemProxy: Move the annotation to molgenis-security-core which doesn't have dependencies on other molgenis modules?

molgenis-app/src/main/java/org/molgenis/migrate/MysqlMigrate.java
fdlk: Not sure how you mean. This is what the MREF tables are currently called. Entity name, underscore, attribute name. If you'd happened to create an entity with the same name, the upload would've failed because the table was already taken. But the migrate script cannot help with that, I'd think?

molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntity.java
fdlk: In MySql, when selecting the row, we've already retrieved lists of IDs (and up to this pull request used to trunkate the string containing the ID list to 1024 chars). In Elasticsearch, we store the full list of children in a document. So is this an academic point or do we actually also have a backend that wouldn't fail elsewhere for MREF lists in the hundreds of MB size of magnitude?

dennishendriksen: We actually have use cases with mref lists of > 10K entities.

molgenis-app/src/main/java/org/molgenis/migrate/MysqlMigrate.java
dennishendriksen: Ok clear, our importer should validate this.

molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntity.java
fdlk: How about we stream it like this then, preserving order, by turning it into separate lookups?
I'm a bit torn here in the tradeoff between order preservation, memory footprint, query count, and clean code.

The other option, I'd think, is that in MySQL, we could also override the getEntities() method of the DefaultEntity, which would then join the MREF table with the refEntity table, sorting in order, and stream the results efficiently from that last-minute single select.

molgenis-app/src/main/java/org/molgenis/migrate/MysqlMigrate.java
fdlk: Heh. Should is the word, I doubt it actually happens. A couple months ago I filed a story about technical constraints on entity names. (#2948) This would be good in there I think.

fdlk:
RunAsSystem -> Erwin has made the same changes as well, and moved things to the security-core package. So we propose to leave this as is so that it compiles and runs, then merge it with Erwin's changes.

PR #2649, opened 2015-03-16
Fix #2602 StackOverflowError when adding MetaData to an Entity with self-reference

### LINKED ISSUE #2602: StackOverflowError when adding MetaData to an Entity with self-reference
### To reproduce:
1. If issue #2637 is fixed, upload *test_circularReference.xlsx*. If it is not, upload *test_circularReferenceMeta.xlsx* and add some rows with references yourself in the Data Explorer. Both files can be found in Google Drive: *Systeemgenetica > GCC > Molgenis > TestData*.
2. Write some code to add an attribute to the *persons* entity, for example:
```
AttributeMetaData attr = new DefaultAttributeMetaData("test");
		dataService.getMeta().addAttribute(entityName, attr);
```
3. Execute the code.

### Expected result:
The attribute is added without errors.

### Actual result:
The attribute is actually added, but first a StackOverflowError is thrown: an endless loop between DefaultEntityMetaData and DefaultAttributeMetaData. 
Line 81 in DefaultEntityMetaData says:
**// TODO: Fails dramatically for self-referencing entities.**

tommydeboer:


molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntityMetaData.java
```diff
@@ -77,15 +77,10 @@ public DefaultEntityMetaData(EntityMetaData entityMetaData)
...
-		// deep copy attributes
-		// TODO: Fails dramatically for self-referencing entities.
 		Iterable<AttributeMetaData> attributes = entityMetaData.getAttributes();
 		if (attributes != null)
 		{
-			for (AttributeMetaData attributeMetaData : attributes)
-			{
-				addAttributeMetaData(new DefaultAttributeMetaData(attributeMetaData));
-			}
+			addAllAttributeMetaData(attributes);
```
fdlk: So now if an entity has an attribute that refers to itself, what happens if you copy it? The copied attribute refers to the old version, I think?

PR #2653, opened 2015-03-16
Fixed #2608 Annotate button enabled in case no annotators available

### LINKED ISSUE #2608: Annotate button enabled in case no annotators available
Reproduce: Open annotator tab for Script entity, all annotators are in the unavailable list, select "copy before annotating". Expected: The annotator button is disabled. Actual: The annotator button is enabled and annotation can be performed.

Mark-de-Haan:


dennishendriksen:
Can you add the issue title to the pull request title? That way we can more easily generate release notes and know what this is commit is about by only looking at the title.

PR #2654, opened 2015-03-16
Fixed #2619 "Unknown entity [exist]" when merging data with name field left empty

### LINKED ISSUE #2619: "Unknown entity [exist]" when merging data with name field left empty
To reproduce:
Upload either some EMX files with genomic data or VCF files (or both). In the Merge Genomic Data plugin, try to merge without specifying a dataset name.

Result:
Error! Unknown entity [exist]

Expected:
No error, just the "This field is required." message.

Mark-de-Haan:


dennishendriksen:
Can you add the issue title to the pull request title? That way we can more easily generate release notes and know what this is commit is about by only looking at the title.

Issue #2660, opened 2015-03-17
No WRITE permission on entity when importing VCF as normal user


bartcharbon:
a "No WRITE permission on entity " message is shown if you try to import VCF as a non-admin user.

PR #2662, opened 2015-03-18
Fixed: #2660 and Story 3294 (Feature/permission in importer)

bartcharbon:


molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportWizard.java
```diff
@@ -145,4 +151,32 @@ public void setMustChangeEntityName(boolean mustChangeEntityName)
...
+	{
+		this.entityNames = entityNames;
+	}
+
+	public List<String> getImportedEntities()
+	{
+		return this.entityNames;
+	}
+
+	public boolean getAllowPermissions()
```
jjettenn: getAllowPermissions change to isAllowPermission

Issue #2661, opened 2015-03-17
MREFs between the same entity do not get properly imported.


fdlk:
Import molgenis-data-mysql/src/test/resources/example.xlsx
Expected: import_person geoffry has two children
Actual: import_person geoffry has no children

The reason is that the entities to import are created as DefaultEntity.

DefaultEntity's `getEntities` and `getEntity` methods use the DataService to retrieve the referenced entities. Unfortunately, the DataService searches in the MySqlRepository that the data is supposed to be written to, not in the ExcelRepository that the import sheet sits in. 

This only makes a difference, I think, when the XREF/MREF attribute references its own entity. For then the rows have not yet been written from the ExcelRepository to the MysqlRepository.

PR #2672, opened 2015-03-19
Fix for issue #2661: mref's between the same entity do get properly

jjettenn:


molgenis-data-mysql/src/test/java/org/molgenis/data/importer/EmxImportServiceTest.java
```diff
@@ -72,35 +80,35 @@ public void testValidationReport() throws IOException, InvalidFormatException, U
 		EntitiesValidationReport report = importer.validateImport(f, source);
 
 		// SheetsImportable
-		Assert.assertEquals(report.getSheetsImportable().size(), 4);
+		AssertJUnit.assertEquals(report.getSheetsImportable().size(), 4);
```
fdlk: Why the switch? (If you do need to switch, you probably also need to switch the argument order around. testng assert works with (actual, expected) while junit assert uses (expected, actual)

molgenis-data-mysql/src/main/java/org/molgenis/data/importer/ImportWriter.java
```diff
@@ -126,21 +127,45 @@ private void importData(EntityImportReport report, Iterable<EntityMetaData> reso
...
 								{
 									return new DefaultEntity(entityMetaData, dataService, entity);
 								}
 							});
 
 					entities = DependencyResolver.resolveSelfReferences(entities, entityMetaData);
-
 					int count = update(repository, entities, dbAction);
+					
+					// Fix self referenced entities where not imported
```
fdlk: typo: where -> were

PR #2673, opened 2015-03-19
fix saving permission=none at import

bartcharbon:


molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportWizardController.java
```diff
@@ -127,28 +127,38 @@ public void addGroupEntityClassPermissions(@RequestParam String groupId, WebRequ
...
+					authority.setRole(SecurityUtils.AUTHORITY_ENTITY_PREFIX + value.toUpperCase() + "_"
+							+ entityClassId.toUpperCase());
+					if (authority.getId() == null)
+					{
+						authority.setId(UUID.randomUUID().toString());
+						dataService.add(GroupAuthority.ENTITY_NAME, authority);
+					}
+					else dataService.update(GroupAuthority.ENTITY_NAME, authority);
+				}
+				else if (value.equals("none"))
```
jjettenn: Add org.molgenis.security.core.Permission.NONE to Permission?
and use it as value.equals(org.molgenis.security.core.Permission.NONE)

Issue #2687, opened 2015-03-24
Inconsistent encoding in the REST controller


jjettenn:
Reproduce:
Upload an attribute with a question mark ('?'), as a part of the id.
When trying to open the filter dialog via the data-explorer you get an error.

Issue #2637, opened 2015-03-13
Uploading Entity with self-reference (xref) fails


tommydeboer:
### To reproduce:
Upload the dataset *test_circularReference.xlsx* (found in Google Drive: *Systeemgenetica > GCC > Molgenis > TestData*), which contains xrefs to itself.

### Expected result:
Data is uploaded without any problems.

### Actual result:
Error: **"circular_persons with id [1] does not exist"**

tommydeboer:
When changing the dataType to mref, it uploads without any problems

Issue #2495, opened 2015-02-19
Landing on a datae-explorer plugin after login results in errors


jjettenn:
Landing on a data-explorer plugin after login results in errors

To reproduce:
1. Create a new data-explorer menuItem. Make sure it will be the first on the list in the menu manager.
2. Log out and login.

Workaround: 
Move the home plugin to the first position in the menu manager.  

Error 1: 
     Error! An error occurred. Please contact the administrator.

Error 2:
After applying the workaround trying to reproduce the error results in a not correctly loaded data-explorer.

PR #2689, opened 2015-03-24
Fix: inconsistent encoding in the REST controller #2687  and Uploading Entity with self-reference (xref) fails #2637 Fix Landing on a datae-explorer plugin after login results in errors #2495 [0.25]

jjettenn:
Fixing encoding.
Encoding was not consistent for the href attribute.
The encoding util for encoding the href attribute is changed because it
was wrong. Now using UriUtils of spring instead of the previus
URLEncoder that was using application/x-www-form-urlencoded instead of
RFC 3986


molgenis-data-mysql/src/main/java/org/molgenis/data/importer/ImportWriter.java
```diff
@@ -507,4 +510,45 @@ public int update(Repository repo, Iterable<? extends Entity> entities, Database
...
+			super(entityMetaData, dataService, entity);
+		}
+
+		/**
+		 * getEntity returns null when attributeName is not resulting in an entity
+		 */
+		@Override
+		public Entity getEntity(String attributeName)
+		{
+			try {
```
fdlk: formatting :)

molgenis-data-mysql/src/main/java/org/molgenis/data/mysql/MysqlRepository.java
```diff
@@ -949,7 +989,10 @@ public void deleteById(Iterable<Object> ids)
 	{
 		final List<Object> idList = new ArrayList<Object>();
 		for (Object id : ids)
+		{
 			idList.add(id);
+			System.out.println("MysqlRepositoy 956 id: " + id.toString());
```
fdlk: sysout

molgenis-data-mysql/src/main/java/org/molgenis/data/mysql/MysqlRepository.java
```diff
@@ -923,10 +924,49 @@ public void delete(Entity entity)
...
+		for (Entity e : entities)
+		{
+			deleteByIdBatch.add(e.getIdValue());
+		}
+		this.deleteById(deleteByIdBatch);
+	}
+
+	/**
+	 * Use before a delete action of a entity with XREF data type were the entity and refEntity are the same entities.
+	 *
```
fdlk: were -> where

fdlk:
I fixed and then tried to upload the typetest dataset. It fails stating that a foreign key constraint is not met. Could it be that cross-entity references are now broken?

jjettenn:
The xcategorymrefnullable name was te groot
Op 25 mrt. 2015 23:10 schreef "Fleur Kelpin" <notifications@github.com>:

fdlk:
Yes. So when I fixed that by making it smaller (see https://github.com/jjettenn/molgenis/pull/2) I got a foreign key constraint failure.

Mark-de-Haan:
You need to drop the database, I think transactional import might be broken
cause I ran into this problem as well. Import fails -> fix EMX -> foreign
key constraints

On 26 March 2015 at 00:40, Fleur Kelpin <notifications@github.com> wrote:

> Yes. So when I fixed that by making it smaller (see jjettenn#2
> <https://github.com/jjettenn/molgenis/pull/2>) I got a foreign key
> constraint failure.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/molgenis/molgenis/pull/2689#issuecomment-86253916>.
>



-- 
Mark de Haan
Genomics Coordination Centre
Genetics Department
University Medical Centre Groningen

Issue #2762, opened 2015-04-08
OMIM HPO annotator broken


bartcharbon:
try running the omim hpo annotator:

Message:org.molgenis.data.UnknownAttributeException: OMIM_Causal_ID

bartcharbon:
probably related to #2756

Issue #2738, opened 2015-04-08
If a user has "count" permissions on a plugin, it is shown in the menu but gives 403 on click


bartcharbon:
- give "count" permission on plugin for a given user
- login as that user
- click the plugin in the menu

result: 403

expected: plugin is not in the menu

Issue #2596, opened 2015-03-06
"No suitable importer found" when uploading .tsv files


tommydeboer:
The DiseaseMapping.tsv and Disease.tsv files can not be uploaded anymore.

dennishendriksen:
Duplicate #2581

PR #2771, opened 2015-04-09
fixed #2755?,#2756,#2762,#2596 and #2738

### LINKED ISSUE #2755: In IE9 all content under the table is increasingly pushed down on hover of an x/mref
select typeTest

hover over an x/mref link
- the scrollbar will change, and content under the table will be pushed down, an increasing region of whitspace will be shown...

### LINKED ISSUE #2756: Using multiple annotators will result in an error
Multiple annotators results in: "Removing of existing attributes is currently not supported."

Probably because the metadata from the original entity is used instead of the resulted metadata from the previous annotator

bartcharbon:
Discuss for #2738 how to deal with "COUNT" permissions on a plugin, should this even be possible?

### DIFF
--- a/molgenis-data-elasticsearch/src/main/java/org/molgenis/data/elasticsearch/ElasticsearchRepositoryCollection.java
+++ b/molgenis-data-elasticsearch/src/main/java/org/molgenis/data/elasticsearch/ElasticsearchRepositoryCollection.java
@@ -92,11 +92,12 @@ public class ElasticsearchRepositoryCollection implements ManageableRepositoryCo
 	@Override
 	public void addAttribute(String entityName, AttributeMetaData attribute)
 	{
-		EntityMetaData entityMetaData = dataService.getEntityMetaData(entityName);
+		EntityMetaData entityMetaData = dataService.getMeta().getEntityMetaData(entityName);
 		if (entityMetaData == null) throw new UnknownEntityException(String.format("Unknown entity '%s'", entityName));
 
 		try
 		{
+			((DefaultEntityMetaData) repositories.get(entityName).getEntityMetaData()).addAttributeMetaData(attribute);
 			searchService.createMappings(entityMetaData);
 		}
 		catch (IOException e)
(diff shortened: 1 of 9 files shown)


molgenis-data-elasticsearch/src/main/java/org/molgenis/data/elasticsearch/ElasticsearchRepositoryCollection.java
```diff
@@ -92,11 +92,12 @@ public void deleteEntityMeta(String entityName)
 	@Override
 	public void addAttribute(String entityName, AttributeMetaData attribute)
 	{
-		EntityMetaData entityMetaData = dataService.getEntityMetaData(entityName);
+		EntityMetaData entityMetaData = dataService.getMeta().getEntityMetaData(entityName);
```
fdlk: The above change, I understand, but this one I find a little worrisome. Does it mean that after annotating the dataService thinks the repo has different metadata than the metadataservice thinks it has?

fdlk: shouldn't the repository get updated with the new metadata? or does that happen later?

bartcharbon: See 4 lines below, but I'll rewrite that bit anyhow, was planning to but forgot.
The change here is probably unnecessary, a artefact of a attempt to understand/fix what was going on.

molgenis-data/src/main/java/org/molgenis/data/meta/MetaDataServiceImpl.java
```diff
@@ -287,6 +287,7 @@ public void refreshCaches()
 	@Transactional
 	public List<AttributeMetaData> updateSync(EntityMetaData sourceEntityMetaData)
 	{
+
```
Mark-de-Haan: Didnt know java had white space coding ;p

fdlk: The devil is in the details

molgenis-data/src/main/java/org/molgenis/data/support/MapEntity.java
```diff
@@ -61,7 +61,8 @@ public void set(Entity other, EntityMetaData metaData)
 		this.idAttributeName = entityMetaData.getIdAttribute().getName();
 		for (AttributeMetaData attribute : metaData.getAtomicAttributes())
 		{
-			set(attribute.getName(), other.get(attribute.getName()));
+			if (Iterables.contains(other.getAttributeNames(), attribute.getName())) set(attribute.getName(),
```
fdlk: This change is what breaks the annotator unit tests. `other` is a mock and getAttributeNames returns null.

PR #2798, opened 2015-04-13
Fix #2739

### LINKED ISSUE #2739: Merging 2 datasets with the name of an existing entity will overwrite this entity
upload 2 vcfs
upload an emx
merge vcfs using the name of the emx

result: succes, merged result is available as the name of the emx

expected: error, cannot merge using the name of an existing dataset.

Mark-de-Haan:
Check if submitted name already exists as an entity. 
Throw exception if true


fdlk:
I don't know how they use this merger but I can imagine it's a nice feature that the merger overwrites the existing merged dataset.

Issue #2766, opened 2015-04-08
Disabled scrollbar in mapping service


fdlk:
Create a mapping project with target entity HOP.
You'll see a disabled scrollbar to the right of the attributes table.

![screen shot 2015-04-08 at 16 48 24](https://cloud.githubusercontent.com/assets/1523838/7047795/1ffdea92-de0f-11e4-995c-2b702fbc0b43.png)

Mark-de-Haan:
Not reproducible in Chrome

fdlk:
Screenshot is in safari

jjettenn:
Test on your own laptop, not with the external screens

jjettenn:
See PR

PR #2801, opened 2015-04-13
Molgenis/r1.5.0bugfix#2766

jjettenn:
Fixed issue: Disabled scrollbar in mapping service #2766


molgenis-data-mapper/src/main/resources/js/single-mapping-project.js
```diff
@@ -1,6 +1,6 @@
 (function($, molgenis) {	
 	$(function() {
-		$('#attribute-mapping-table').scrollTableBody();
+		$('#attribute-mapping-table').scrollTableBody({rowsToDisplay:10});
```
fdlk: that's the default already, I think?

PR #2818, opened 2015-04-14
Fix #2814

### LINKED ISSUE #2814: As user I get obscure error messages if I try to upload an EMX file that has already been uploaded by someone else
As normal user, upload the all datatypes file that the admin user has also uploaded (but that you have no rights for)

You'll get an error message saying

"Unable to get value of the property 'length': object is null or undefined"
And a somewhat more informative message
"No COUNT permission on entity org_molgenis_test_TypeTestRef"

I'd expect the validation to check this sort of thing before the actual import starts.

Mark-de-Haan:


Mark-de-Haan:
Not the fix we are looking for

PR #2876, opened 2015-04-21
Adding checks to FTL; Bugfix #2859

jjettenn:
Adding checks to FTL; Bugfix #2859

### DIFF
--- a/molgenis-data-import/src/main/resources/templates/ValidationResultWizardPage.ftl
+++ b/molgenis-data-import/src/main/resources/templates/ValidationResultWizardPage.ftl
@@ -51,7 +51,11 @@
 	</thead>
 	
 	<#list wizard.entitiesImportable?keys as entity>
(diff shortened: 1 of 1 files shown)


molgenis-data-import/src/main/resources/templates/ValidationResultWizardPage.ftl
```diff
@@ -51,7 +51,11 @@
 	</thead>
 	
 	<#list wizard.entitiesImportable?keys as entity>
-		<#if wizard.entitiesImportable[entity] == true && (wizard.fieldsDetected[entity]?size gt 0 || wizard.fieldsRequired[entity]?size gt 0 || wizard.fieldsAvailable[entity]?size gt 0 || wizard.fieldsUnknown[entity]?size gt 0) >
+		<#if wizard.entitiesImportable[entity] == true 
+			&& ((wizard.fieldsDetected[entity]?? && wizard.fieldsDetected[entity]?size gt 0)
```
Mark-de-Haan: you can do these checks with only ?? (checks if > 0  and if it is not null

PR #2882, opened 2015-04-21
Fix/2873

fdlk:
What happened was
1. TypeTestRef gets created
2. Creation of TypeTest fails
3. Importer begins to roll back table creations
4. Rollback of creation of TypeTest fails because it didn't get created
5. Importer gives up rolling back table creations

Changed 2 things:
* MysqlRepositoryCollection no longer considers it a good idea to throw an exception if it is asked to delete a repository that doesn't exist.
* Importer tries to rollback all table creations even if a couple of the rollbacks fail

### DIFF
--- a/molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportWriter.java
+++ b/molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportWriter.java
@@ -331,7 +331,16 @@ public class ImportWriter
 	private void dropAddedEntities(List<String> addedEntities)
 	{
 		// Rollback metadata, create table statements cannot be rolled back, we have to do it ourselves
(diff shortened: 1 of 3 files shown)


molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportWriter.java
```diff
@@ -558,7 +567,7 @@ public Entity getEntity(String attributeName)
 		@Override
 		public Iterable<Entity> getEntities(String attributeName)
 		{
-			return from((Iterable<Entity>) super.getEntities(attributeName)).filter(notNull());
+			return from(super.getEntities(attributeName)).filter(notNull());
```
fdlk: My eclipse does this automatically cause it considers this a pointless cast.

PR #2893, opened 2015-04-23
Feature/semanticsearch

jjettenn:
Semantic search is implemented, fixed and tested


molgenis-data-mapper/src/main/resources/templates/view-attribute-mapping.ftl
```diff
@@ -34,7 +34,7 @@
 						</tr>
 					</thead>
 					<tbody>
-						<#list entityMapping.sourceEntityMetaData.attributes as source>
+						<#list suggestedAttributes as source>
 							<tr>
```
fdlk: Maybe if one list is empty write down the other? Or allow the user to switch between the two?

molgenis-data-semanticsearch/src/main/java/org/molgenis/data/semanticsearch/service/impl/SemanticSearchServiceHelper.java
```diff
@@ -0,0 +1,144 @@
...
+				new Function<Entity, String>()
+				{
+					public String apply(Entity attributeEntity)
+					{
+						return attributeEntity.getString(AttributeMetaDataMetaData.IDENTIFIER);
+					}
+				}).toList();
+	}
+
+	public List<OntologyTerm> findTagsSync(String description, List<String> ontologyIds)
```
fdlk: I'd name the Async method Async, if it exists, but otherwise all methods are Sync by default, so I'd not put that in the name.

molgenis-data-semanticsearch/src/test/java/org/molgenis/data/semanticsearch/service/impl/OntologyTagServiceTest.java
```diff
@@ -135,7 +133,7 @@ public void testgetTagsForAttribute()
 		when(ontologyService.getOntologyTerm("http://edamontology.org/data_0987")).thenReturn(chromosomeName);
 		when(ontologyService.getOntologyTerm("http://edamontology.org/data_0919")).thenReturn(geneAnnotation);
 
-		Multimap<Relation, OntologyTerm> expected = ArrayListMultimap.<Relation, OntologyTerm> create();
```
fdlk: Why not keep the interface, just initialize it differently?

PR #2903, opened 2015-04-23
Feature/semanticsearch :

ChaoPang:
-  broke createDisMaxQuery method into 2 new methods; 1. createTargetAttributeQueries to extract all the information (label, description, tags) about targetAttribute; 2. createDisMaxQueryRule to create disMaxQuery based on a given list of strings
- fixed the bug that description of attributes were not used in the semanticsearch. Now both of description and label are used in semanticsearch.


fdlk:
Can you merge with molgenis:feature/semanticsearch branch?

PR #2904, opened 2015-04-23
Fix #2902

### LINKED ISSUE #2902: Ace editor is unable to find worker-javascript.js
See https://github.com/ajaxorg/ace/issues/655

fdlk:
Updated the ace version to 1.1.9.
Added the config item that sets the base url to the ace.js script:
```
window.ace.config.set("basePath", "/js/ace/src-min-noconflict/");
```
It will *not* work on a machine that doesn't have molgenis as ROOT webapp.
(Any suggestions on how to fix this there?)


dennishendriksen:
Alternatively you could use Ace.js (which is a React component) that already solves this problem for you. Creating an instance of Ace and listening to value changes should be quite easy.

PR #2927, opened 2015-04-28
add specific datatable for col7a1

bartcharbon:


Mark-de-Haan:
New data table does not have a row click event listener, fix please

PR #2965, opened 2015-04-30
Fix #2935 Forms: error selecting categoricals with int ids (master)

### LINKED ISSUE #2935: Forms: error when selecting categoricals that have int as identifier
### To reproduce:
1. Upload an EMX that has categoricals with int as identifier
2. Edit one of the entities and select one of the categorical options

### Expected:
No error.

### Actual:
```Uncaught TypeError: Cannot read property 'label' of undefined```

dennishendriksen:


tommydeboer:
Doesn't work yet with decimal or categorical_mref

PR #2971, opened 2015-04-30
Feature/bootstrap theme select

Mark-de-Haan:
## Features
* New Menu Manager layout
* Now contains a new drop down allowing you to select several bootstrap themes
* Bootstrap themes are applied to Menu Manager as a preview and are only applied to the entire application when saved

## Bugfixes
* Removed some CSS files that were duplicated in minified and non minified versions
* Some minor fixes to existing html element classes which were not 100% bootstrap


Mark-de-Haan:
Adding newest bootstrap version first

PR #2990, opened 2015-05-05
Migration step for changing storage engine to InnoDB for all MySQL tables (r1.6.0)

tommydeboer:


molgenis-data-migrate/src/main/java/org/molgenis/data/version/MigrationUtils.java
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	public static String getVersion()
+	{
+		return getServerProperty(VERSION_KEY);
+	}
+
+	public static Properties getMolgenisServerProperties()
+	{
+		try (InputStream in = new FileInputStream(getMolgenisServerPropertiesFile()))
```
dennishendriksen: Close stream in finally

dennishendriksen: Use InputStreamReader with UTF-8 encoding instead of InputStream, since Properties.load will use the platform default encoding (different on Mac, Windows and Linux)

molgenis-data-migrate/src/main/java/org/molgenis/data/version/MigrationUtils.java
```diff
@@ -0,0 +1,76 @@
...
+			throw new UncheckedIOException(e);
+		}
+	}
+
+	public static File getMolgenisServerPropertiesFile()
+	{
+		if (propertiesFile == null)
+		{
+			// get molgenis home directory
+			String molgenisHomeDir = System.getProperty("molgenis.home");
```
dennishendriksen: Also used in Molgenis web app properties, store value in public static String somewhere

molgenis-data-migrate/src/main/java/org/molgenis/data/version/MigrationUtils.java
erwinwinder: Try-With-Resources statement is used, to close is not needed (is implicit)

tommydeboer: It's a try-with-resources statement, the reader will be closed automatically.

Issue #3026, opened 2015-05-13
Mapping dates are not working correctly


jjettenn:
Mapping dates are not working correctly

To reproduce: try to test a date value

jjettenn:
Already done

PR #3027, opened 2015-05-13
Fix for convert date in mapping service

jjettenn:
Issue #3026


fdlk:
Let's add a unit test for this cause it's the second time it breaks already.

PR #3046, opened 2015-05-15
Fix #3009 Chosing non-existing css file breaks MOLGENIS

### LINKED ISSUE #3009: Chosing non-existing css file breaks MOLGENIS
Morris:"I managed to break molgenis01 because I chose a non-existent css file."

Mark-de-Haan:
* Fix website crashing when submitting a non existing css
* Fix theme manager showing duplicates in the dropdownlist


molgenis-core-ui/src/main/resources/css/themes/bootstrap-molgenis.css
```diff
@@ -1,20 +0,0 @@
...
-	padding: 5px;
-}
-
-label.horizontal-inline {
-	margin-right: 5px;
-}
-
-label.horizontal-inline.inbetween {
-	margin-left: 5px;
-}
```
fdlk: I think I'd like to keep the source around in case we want to change it? :)

Mark-de-Haan: http://unminify.com/?
I was thinking the same thing but then i thought how little effort it is to unminify it.
Reason for removing is the ResourcePath object we use to generate a list of styles uses the molgenis.css as well as the generated molgenis.min.css

Creating two molgenis options in the user interface

pom.xml
```diff
@@ -529,11 +529,14 @@
                                 <webappTargetDir>${project.build.directory}/generated-resources</webappTargetDir>
                                 <cssSourceIncludes>
                                     <cssSourceInclude>**/*.css</cssSourceInclude>
+                                    <cssSourceInclude>**/**/*.css</cssSourceInclude>
                                 </cssSourceIncludes>
                                 <cssSourceExcludes>
                                     <!-- changes to excludes must be applied to molgenis-core-ui/src/main/resources/templates/resource-macros.ftl -->
                                     <cssSourceExclude>**/*.min.css</cssSourceExclude>
                                     <cssSourceExclude>**/*-min.css</cssSourceExclude>
+                                    <cssSourceExclude>**/**/*.min.css</cssSourceExclude>
```
fdlk: Above comment line suggests that resource-macros.ftl also needs to be changed then?

molgenis-core-ui/src/main/resources/css/themes/bootstrap-molgenis.css
fdlk: I think I got the deduplication running in a different way, plus added unit tests for the style service. What do you think of https://github.com/Mark-de-Haan/molgenis/pull/10 ?

PR #3115, opened 2015-06-02
Feature/offline styles

Mark-de-Haan:
Remove the dependency on the bootswatch API by having the bootstrap themes locally


molgenis-core-ui/src/main/java/org/molgenis/ui/style/StyleServiceImpl.java
```diff
@@ -75,7 +45,11 @@
 	@Override
 	public void setSelectedStyle(String styleName)
 	{
-		molgenisSettings.setProperty(CSS_THEME_KEY, getStyle(styleName).getLocation());
+		// Pressing save in the UI without doing a selection returns undefined
+		if (!styleName.equals("undefined"))
+		{
+			molgenisSettings.setProperty(CSS_THEME_KEY, getStyle(styleName).getLocation());
+		}
```
fdlk: While this is nice, the javascript should save the currently selected item in the dropdown instead of undefined.

PR #3117, opened 2015-06-03
Document property to activate dev profile.

fdlk:


README.md
```diff
@@ -78,6 +78,7 @@ Create the file ~/.molgenis/omx/molgenis-server.properties and add user and data
     db_uri=jdbc:mysql://localhost/omx
     admin.password=admin
     user.password=admin
+    environment=development
```
Mark-de-Haan: I would not place this here, the sentence below mentions that without these properties the build will fail. But you do not necessarily need the environment property

PR #3122, opened 2015-06-04
Fix:3072 Unable to delete entities with self-reference

bartcharbon:


molgenis-data-mysql/src/main/java/org/molgenis/data/mysql/MysqlRepository.java
```diff
@@ -113,7 +113,7 @@ public void drop()
 		// Find out if the entity is referenced and if it is, report those entities
 		List<Pair<EntityMetaData, List<AttributeMetaData>>> referencingEntities = EntityUtils
 				.getReferencingEntityMetaData(getEntityMetaData(), dataService);
-		if (!referencingEntities.isEmpty())
+		if (!referencingEntities.isEmpty() && !isSelfReferencing(referencingEntities))
 		{
```
fdlk: what if it's self-referencing AND other entities reference it?

bartcharbon:
this request is obsolete after last pull request

PR #3124, opened 2015-06-04
M3498 Package/entity/attribute names now validated

tommydeboer:
- Allowed chars: [a-zA-Z0-9_#]
- Max length: 30
- Can't start with a digit

### DIFF
--- a/molgenis-data/src/main/java/org/molgenis/data/meta/MetaValidationUtils.java
+++ b/molgenis-data/src/main/java/org/molgenis/data/meta/MetaValidationUtils.java
@@ -96,7 +96,7 @@ public class MetaValidationUtils
 
 		if (name.length() > MAX_ATTRIBUTE_LENGTH)
 		{
-			throw new MolgenisDataException("Attribute name [" + name + "] is too long: maximum length is "
+			throw new MolgenisDataException("Name [" + name + "] is too long: maximum length is "
 					+ MAX_ATTRIBUTE_LENGTH + " characters.");
 		}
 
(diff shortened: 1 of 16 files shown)


molgenis-data/src/main/java/org/molgenis/data/meta/MetaValidationUtils.java
```diff
@@ -0,0 +1,67 @@
...
+
+		if (!name.matches("[a-zA-Z0-9_#]+"))
+		{
+			throw new MolgenisDataException("Invalid characters in: [" + name
+					+ "] Only letters (a-z, A-Z), digits (0-9) and underscores (_) are allowed.");
+		}
+
+		if (Character.isDigit(name.charAt(0)))
+		{
+			throw new MolgenisDataException("Invalid name: [" + name + "] Names must start with a letter.");
```
dennishendriksen: Starting with underscore is also allowed.

dennishendriksen:
Extend the validation with a reserved word list: 1) reserved words for Ecmascript 5 and 6 variable names, 2) reserved words for java variable names 3) reserved words for molgenis: login, logout, csv (will lead to collisions in RestController).

molgenis-data/src/main/java/org/molgenis/data/meta/MetaValidationUtils.java
dennishendriksen: Missing hash in exception message.

tommydeboer: We have to allow starting with hash because of VCF (#CHROM). Underscore, I'm not sure.

fdlk: Underscores anywhere will clash with the underscore as a package/entity name delimiter, I think?

tommydeboer: It's allowed now and I've seen it in practice. Doesn't seem to break anything. Next sprint we'll look at the package separators (will probably become a dot).

PR #3135, opened 2015-06-08
fix #3104 Tag Wizard: Run the Automatic Tagging multiple times and it will add the same tags multiple times

### LINKED ISSUE #3104: Tag Wizard: Run the Automatic Tagging multiple times and it will add the same tags multiple times
Tag Wizard: Run the Automatic Tagging multiple times and it will add the same tags multiple times

bartcharbon:


Mark-de-Haan:
Works, 
Although If you try to manually add the same tag you get:
```Uncaught TypeError: Cannot read property 'IRI' of undefined```

PR #3136, opened 2015-06-09
fix #1698 MySql does not support the unique attribute correctly.

### LINKED ISSUE #1698: xrefmref select incorrectly uses lookup attributes when no unique attributes are available
This is needed for the mref select2 functionality when reopening a filter from the filter list

The current work around is:
When no unique attributes are available the lookup attribute is used to search for the right entity.

bartcharbon:


molgenis-core-ui/src/main/resources/js/jquery.molgenis.xrefmrefsearch.js
```diff
@@ -147,7 +147,7 @@
 			},
 			initSelection: function(element, callback) {
 				//Only called when the input has a value
-				var attrNames = (uniqueAttrNames.length ? uniqueAttrNames : lookupAttrNames);
+				var attrNames = uniqueAttrNames;
```
fdlk: I don't understand, why does it use uniqueAttrNames to begin with, instead of the ID column?

PR #3138, opened 2015-06-09
Fix #3099 MappingService: after running the automatic tagging service trying to edit results in error, Fix #1698 MySql does not support the unique attribute correctly

### LINKED ISSUE #1698: xrefmref select incorrectly uses lookup attributes when no unique attributes are available

### LINKED ISSUE #3099: MappingService: after running the automatic tagging service trying to edit results in error (Mappings for the <Name project> project)
To reproduce:
1. Upload the hp.owl.zip ontology
2. Upload the small_test_data.owl.zip file
3. Create a project with target model Person
4. Use the Person also as source
5. Run tag wizard
6. Run Automatic Tagging use the hp ontology
7. Mortality/Aging   Middle age onset   Age-dependent penetrance  will be added
8. Go back to mapping project and click on pencil
9. You will get an error: {"errors":[{"message":"An error occurred. Please contact the administrator.<br />.........

bartcharbon:


molgenis-data-semanticsearch/src/main/java/org/molgenis/data/semanticsearch/service/impl/SemanticSearchServiceHelper.java
```diff
@@ -102,7 +102,7 @@ public QueryRule createDisMaxQueryRule(EntityMetaData targetEntityMetaData, Attr
 			for (String synonym : synonyms)
 			{
 				disMaxQuery.getNestedRules().add(
-						new QueryRule(AttributeMetaDataMetaData.LABEL, Operator.FUZZY_MATCH, synonym));
+						new QueryRule(AttributeMetaDataMetaData.LABEL, Operator.FUZZY_MATCH, QueryParser.escape(synonym)));
 			}
```
fdlk: Is this related?

PR #3140, opened 2015-06-09
Fix #3068

### LINKED ISSUE #3068: A REST API request with a logged out token gives error message in HTML instead of JSON
Reproduce: 
Execute the automated REST API test. Last step will fail.

fdlk:
Seems fair to me, treat invalid tokens as if no token at all is present.


fdlk:
Checked that the REST controller will send that unauthorized statuscode anyways.

PR #3141, opened 2015-06-09
Fix #2528

### LINKED ISSUE #2528: Importing VCF allows NaN as a valid Float value, but data explorer hangs on it
When encountering NaN, rendering of the table stops and this error is thrown:

10:41:00.822 [qtp2022171123-27] ERROR o.molgenis.data.rest.RestController - 
java.lang.IllegalArgumentException: NaN is not a valid double value as per JSON specification. To override this behavior, use GsonBuilder.serializeSpecialFloatingPointValues() method.
	at com.google.gson.Gson.checkValidFloatingPoint(Gson.java:296) ~[gson-2.2.4.jar:na]
	at com.google.gson.Gson.access$000(Gson.java:97) ~[gson-2.2.4.jar:na]
	at com.google.gson.Gson$3.write(Gson.java:264) ~[gson-2.2.4.jar:na]
	at com.google.gson.Gson$3.write(Gson.java:250) ~[gson-2.2.4.jar:na]

Mini VCF file to reproduce bug:
```
##fileformat=VCFv4.1
##INFO=<ID=TDT,Number=1,Type=Float,Description="Test statistic from Wittkowski transmission disequilibrium test.">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
1	880639	rs113796517	TC	T	1553.49	.	TDT=NaN;
```

fdlk:


molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/VcfRepository.java
```diff
@@ -150,6 +160,10 @@ public String apply(Allele allele)
 							// TODO support list of primitives datatype
 							val = StringUtils.join((List<?>) val, ',');
 						}
+						if (val instanceof Float && Float.isNaN((Float) val))
```
Mark-de-Haan: so values in a VCF can never be floats??

PR #3162, opened 2015-06-15
Fix #3161 Max file upload size cannot exceed 2047MB

### LINKED ISSUE #3161: Max file upload size cannot exceed 2047 (mb)
Higher values will internally roll over max int when computing the max file upload size in bytes since it gets stored in an int.

fdlk:


fdlk:
Let's first get palga running properly before merging to the master

PR #3165, opened 2015-06-16
Set upload limit to 4GB.

fdlk:


dennishendriksen:
Increasing the limit to 4GB will result in our average VM running MOLGENIS running out of disk space after two uploads. How to deal with this?

PR #3173, opened 2015-06-18
stream small files, store large files on disk

fdlk:


dennishendriksen:
Could you explain what this change does?

fdlk:
It changes the fileSizeThreshold of the [MultipartConfigElement](http://docs.oracle.com/javaee/6/api/javax/servlet/MultipartConfigElement.html) from 2GB (my previous attempt) to 10MB.
Files smaller than 10 MB will be treated the current way, i.e. streamed in-memory. Larger files will be stored in the tmp folder. Without this change, the heap space overflows if I upload a large file.

dennishendriksen:
Ok, thanks.

PR #3204, opened 2015-06-23
Refactor in importer for Selenium-Tests

jjettenn:


molgenis-core-ui/src/main/java/org/molgenis/ui/wizard/WizardButton.java
```diff
@@ -2,17 +2,24 @@
 
 public class WizardButton
 {
+	private final String id;
 	private final String title;
 	private final boolean enabled;
 	private final String targetUri;
 
-	public WizardButton(String title, boolean enabled, String targetUri)
+	public WizardButton(String id, String title, boolean enabled, String targetUri)
```
fdlk: That means wizards can no longer be nested, right?

PR #3206, opened 2015-06-24
M3485 As data mapper I want to clone a mapping project

dennishendriksen:
- Add button that clones mapping project to mapping overview
- Remove setting permissions on forms


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
```diff
@@ -46,6 +55,15 @@
...
 	@Autowired
+	private MappingTargetRepository mappingTargetRepository;
+
+	@Autowired
+	private EntityMappingRepository entityMappingRepository;
+
+	@Autowired
+	private AttributeMappingRepository attributeMappingRepository;
+
+	@Autowired
```
fdlk: The idea was to hide this implementation from the service and let the mappingprojectrepo delegate the incoming work.

PR #3207, opened 2015-06-24
Fix #3183 UML diagram not scrollable in Chrome

### LINKED ISSUE #3183: UML diagram not scrollable in Chrome
Niek just showed me that if you upload a relatively large UML diagram, it becomes unnavigable in chrome.
In safari you can only navigate when you go fullscreen and use a trackpad, but in chrome you cannot zoom and then scroll to the rightmost part of the diagram.

jjettenn:
1. Zoom in out buttons are fixed.
2. You can scale the graph to the size of your screen.
3. Fixed chrome bug of not showing the scroll-bar


fdlk:
The print preview looks a little off now.

jjettenn:
I hope this helps :)

PR #3240, opened 2015-06-26
Fix #3161 (master)

### LINKED ISSUE #3161: Max file upload size cannot exceed 2047 (mb)

fdlk:


molgenis-core-ui/src/main/java/org/molgenis/ui/MolgenisWebAppInitializer.java
```diff
@@ -20,6 +20,9 @@
 
 public class MolgenisWebAppInitializer
 {
+	private static final int MB = 1024 * 1024;
```
jjettenn: Should his not be also a long?

fdlk: No, it cannot be a long. The fileSizeThreshold parameter of the MultipartConfigElement constructor is an int cause it shouldn't be higher than maxint.

PR #3256, opened 2015-06-29
Specify stream encoding to fix findbugs warning

fdlk:


Mark-de-Haan:
Duplicate findbugs fix with  #3255

PR #3269, opened 2015-06-30
M3564 Mapping service error feedback

dennishendriksen:


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
```diff
@@ -244,6 +251,88 @@ public String removeEntityMapping(@RequestParam String mappingProjectId, String
...
+		Long offset = mappingServiceRequest.getOffset();
+		Long num = mappingServiceRequest.getNum();
+		Query query = new QueryImpl().offset(offset.intValue()).pageSize(num.intValue());
+		String sourceEntityName = mappingServiceRequest.getSourceEntityName();
+		Iterable<Entity> sourceEntities = dataService.findAll(sourceEntityName, query);
+
+		long total = dataService.count(sourceEntityName, new QueryImpl());
+		long nrSuccess = 0, nrErrors = 0;
+		Map<String, String> errorMessages = new LinkedHashMap<String, String>();
+		for (AlgorithmEvaluation evaluation : algorithmService.applyAlgorithm(targetAttr, algorithm, sourceEntities))
```
Mark-de-Haan: If an attribute does not exist it will through a nullpointer

fdlk: But it got checked on line 264, no?

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
```diff
@@ -650,8 +739,21 @@ public String advancedMappingEditor(@RequestParam(required = true) String mappin
 		AttributeMetaData targetAttribute = targetEntityMetaData != null ? targetEntityMetaData
 				.getAttribute(mappingServiceRequest.getTargetAttributeName()) : null;
 		Repository sourceRepo = dataService.getRepository(mappingServiceRequest.getSourceEntityName());
-		List<Object> calculatedValues = algorithmService.applyAlgorithm(targetAttribute,
+
+		Iterable<AlgorithmEvaluation> algorithmEvaluations = algorithmService.applyAlgorithm(targetAttribute,
```
fdlk: Can (and will) currently throw an exception at times.
What's the desired interface, should it be caught in the service or here?

molgenis-data-mapper/src/main/resources/js/attribute-mapping.js
```diff
@@ -274,7 +355,10 @@
 			// update algorithm
 			algorithm = editor.getSession().getValue();
 
-			// update result
+			// validate mapping
+			validateAttrMapping(algorithm);
```
fdlk: This doesn't trigger validation when the map gets updated.

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
Mark-de-Haan: Line 264 checks if the targetAttribute is null i.e. not there. The null pointer that is thrown is caused by an attribute that does not exist in the source entity

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/AlgorithmServiceImpl.java
```diff
@@ -66,59 +64,33 @@ public void autoGenerateAlgorithm(EntityMetaData sourceEntityMetaData, EntityMet
...
-						}
-					}
-					catch (RuntimeException e)
-					{
-						LOG.error("error converting result", e);
-					}
+					return algorithmResult.errorMessage(e.getMessage());
 				}
+
+				Object derivedValue = convert(result, targetAttribute);
```
fdlk: conversion should happen inside the try, I think

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
dennishendriksen: MappingServiceRequest.targetAttributeName can't be null according to annotation.
EntityMetaData.getAttribute(String) returns null if not found according to documentation.

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
dennishendriksen: Method should not throw an exception in case of an invalid algorithm (which doesn't happen at the moment). Other RuntimeExceptions should be handled as generic RuntimeExceptions.

molgenis-data-mapper/src/main/resources/js/attribute-mapping.js
dennishendriksen: I don't understand your comment. The validation is not triggered when displaying the current algorithm by design. Do you mean this behavior should be different?

Mark-de-Haan: The map tab being changed will update the result, but will not activate the validateAttrMapping. This is not a problem for now but we should remember it for future reference

PR #3276, opened 2015-06-30
Feature for highlighting words that were matched by Lucene

Mark-de-Haan:
Merge conflicts are a result from cherrypicking
Will get merge conflicts with #3269 because we worked in the same files


molgenis-data-mapper/src/main/resources/js/attribute-mapping.js
```diff
@@ -201,11 +286,23 @@
 	 * Hides rows of the table if atrribute source labels, names, descriptions
 	 * and tags have nothing to do with the query, hide the row
 	 */
-	function filterAttributeTable() {
-		var searchQuery = $('#attribute-search-field').val().toLowerCase(), attrLabel, attrName, attrDescription;
+	function filterAttributeTable(explainedAttributes, attributes) {
+		var searchQuery = $('#attribute-search-field').val().toLowerCase(), attrLabel, attrName, attrDescription, explainedQueryStrings;
 		if (searchQuery === '') {
 			$('#attribute-mapping-table>tbody').find('tr').each(function() {
-				$(this).show();
```
fdlk: happens always so can stay up here

molgenis-data-mapper/src/main/resources/js/attribute-mapping.js
```diff
@@ -201,11 +286,23 @@
...
+				if (attributes !== null) {
+					if (attributes.indexOf($(this).attr('class').toLowerCase()) > -1) {
+						explainedQueryStrings = explainedAttributes[$(this).attr('class')];
+						$(this).show();
+						$(explainedQueryStrings).each(function() {
+							$('td.source-attribute-information').highlight(this.matchedWords);
```
fdlk: I'd expect a selector that's more specific to this particular attribute.

PR #3345, opened 2015-07-03
Enhance the highlight function in algorithm editor

ChaoPang:
- change the css of highlight
- extend the highlight to the end of the word because the matched words provided by ExplainAPI are stemmed
- combine highlighted words that are sitting next to each other in the attributeLabel and highlight them as one unit


molgenis-data-mapper/src/main/resources/js/attribute-mapping.js
```diff
@@ -362,6 +479,7 @@
 
 				// Call the filterAttributeTable to only show the attributes
 				// that are explained
+				rankAttributeTable(explainedAttributes);
 				filterAttributeTable(explainedAttributes, attributes);
```
Mark-de-Haan: explainedAttributes arent used anymore in the filterAttributeTable method, you can remove it from the function and call

PR #3364, opened 2015-07-08
Enhance the attribute mapping screen

ChaoPang:
- implemented all the feature requirements for enhancing the attribute mappings screen story #3603
- Changed the EnglishStemmer to PorterStemmer so it`s in sync with ElasticSearch


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
```diff
@@ -496,6 +496,17 @@ public String viewAttributeMapping(@RequestParam(required = true) String mapping
 			attributeMapping = entityMapping.addAttributeMapping(targetAttribute);
 		}
 
+		FieldType targetAttributeDataType = dataService.getEntityMetaData(target).getAttribute(targetAttribute)
+				.getDataType();
+
+		if (targetAttributeDataType instanceof XrefField)
```
Mark-de-Haan: This can all be done in the ftl because we have targetAttributeMetaData in the model

PR #3384, opened 2015-07-10
Refactored the annotator module. Added a result.filter package. Added a query.generator package

Mark-de-Haan:
Some simple refactoring of class names (removed service from CGD and SnpEff annotators) and moved query generator classes and result filter classes from entity.impl to their own packages.


bartcharbon:
what else do you expect in the result package, or the query package? In other words, why are there two layers of package added?

PR #3387, opened 2015-07-13
Revert "Print compound attribute values in entity toString". The toString is causing null pointers in the abstract entity tests. Quick revert to fix the build.

Mark-de-Haan:
Reverts molgenis/molgenis#3383


bartcharbon:
Besides 3383 is replacing one problem with another, it still does not reflect the attribute structure because it lacks the compounds...

PR #3388, opened 2015-07-13
Feature/1000g

brammiedema:
Added and tested the following annotators DANN, GoNL, 1000G and fitcon.


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/MultiFileResource.java
```diff
@@ -93,6 +93,8 @@ public String getName()
 	@Override
 	public Iterable<Entity> findAll(Query q)
 	{
+		// initialize after autowiring is complete and resources is empty
+		if (resources.isEmpty()) initializeResources();
```
bartcharbon: use isAvailable instead?

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/ResourcesImpl.java
```diff
@@ -61,4 +62,10 @@ public boolean hasRepository(String entityName)
 		}
 		return dataService.findAll(name, q);
 	}
+
+	public Set<String> getResourcesNames()
```
bartcharbon: What is this function for?

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/Resources.java
```diff
@@ -24,4 +27,11 @@
 	 */
 	Iterable<Entity> findAll(String name, Query q);
 
+	/**
```
bartcharbon: What is this function for? Comment mismatches with the function.

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/CmdLineAnnotator.java
```diff
@@ -157,13 +182,20 @@ else if (annotatorName.equals("exac"))
 			RepositoryAnnotator annotator = annotators.get("exac");
 			annotate(annotator, inputVcfFile, outputVCFFile);
 		}
-		else if (annotatorName.equals("1kg"))
+		else if (annotatorName.equals("thousand_genomes"))
```
jjettenn: Remove old annotator

bartcharbon:
Please add unit tests for new annotators

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/ExacAnnotator.java
```diff
@@ -25,7 +25,7 @@
 @Configuration
 public class ExacAnnotator
 {
-
+	// TODO Write test
```
bartcharbon: remove this line.

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/MultiFileResource.java
brammiedema: resources does not have a function is available

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/ResourcesImpl.java
brammiedema: It was needed to figure out of the resource is empty or the files were wrongly configured.
The error message thrown is the same in both conditions.

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/ExacAnnotator.java
brammiedema: The old one still has functionality that the new does not, are you sure we need to remove this?

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/Resources.java
brammiedema: To be honest I don't know mark did some work on this too might be his code. But I suspect that it has something to do with more then one hit found in the resource files.

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/ResourcesImpl.java
bartcharbon: Where is it used?

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/MultiFileResource.java
bartcharbon: yes it does

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/ResourcesImpl.java
brammiedema: no where any more i can remove it

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/MultiFileResource.java
brammiedema: oops

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/Resources.java
bartcharbon: To be clear, this is about the "getResourcesNames()" not the "findAll" that one we obviously need.
Github shows it a bit confusing because I create the line note on the first line of the comment.

brammiedema: ok, well when we needed it, it wasn't in the interface so we had to add it there as well.
It is just to check if the configuration is going smoothly when debugging, there is no other way to get a list of available loaded resource(s) entities. And the errors do not show a clear distinction between missing resources or failed to load files. if the resource file is not available for any reason what so ever it will just try to look the data up in the dataservice. This will then throw an entity not found exception.

jjettenn:
We must first fix stuff

PR #3391, opened 2015-07-14
Offer the option to override the filename for specific Chromosomes, i…

bartcharbon:
…n case they don't fit the pattern

### DIFF
--- a/molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/MultiResourceConfigImpl.java
+++ b/molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/MultiResourceConfigImpl.java
@@ -18,18 +18,36 @@ public class MultiResourceConfigImpl implements MultiResourceConfig
 	private String chromosomesProperty;
 	private String filePatternProperty;
 	private String folderProperty;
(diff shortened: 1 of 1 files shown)


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/resources/impl/MultiResourceConfigImpl.java
```diff
@@ -18,18 +18,36 @@
...
 
+	/**
+	 * overrideChromosomeFiles contains the chromosome that do not comply with the pattern.
+	 * those chromosomes and corresponding filenames should be stated comma separated per chromosome,
+	 * and the chromosome and filename itself should be colon separated.
+	 * e.g. 1:file1.vcf,2:file2.txt,X:XYZ.vcf
+	 * */
 	public MultiResourceConfigImpl(String chromosomesProperty, String filePatternProperty, String folderProperty,
-			MolgenisSettings molgenisSettings)
+			MolgenisSettings molgenisSettings, String overrideChromosomeFiles)
```
bartcharbon: Should be updated to using the property here instead of the value, I will update

jjettenn:
The changes will be pushed in to the branch feature/annotators_sprint70.

PR #3392, opened 2015-07-14
Feature/catalogue [Master]

Mark-de-Haan:


molgenis-data-migrate/src/main/java/org/molgenis/data/version/MolgenisVersionService.java
```diff
@@ -39,7 +39,7 @@
 @Service
 public class MolgenisVersionService
 {
-	public static final int CURRENT_VERSION = 13;
+	public static final int CURRENT_VERSION = 12;
```
tommydeboer: Are we absolutely sure nobody will be using 1.8.0? Decreasing the version here will cause migration problems in the future for these users.

Mark-de-Haan: No versions of 1.8 have been deployed yet. And 1.8.1 will be released with the same change. I do not expect any bad things to happen

PR #3406, opened 2015-07-15
Feature/1000g

brammiedema:
Fixed issues with the clinvar annotation when it did not pair the right clinical significant value to the corresponding allele in some cases.

### DIFF
--- a/molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/filter/ClinvarMultiAllelicResultFilter.java
+++ b/molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/filter/ClinvarMultiAllelicResultFilter.java
@@ -1,0 +1,139 @@
+package org.molgenis.data.annotation.filter;
+
+import com.google.common.collect.FluentIterable;
+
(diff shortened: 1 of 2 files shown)


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/filter/ClinvarMultiAllelicResultFilter.java
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	@Override
+	public Collection<AttributeMetaData> getRequiredAttributes()
+	{
+		return Arrays.asList(VcfRepository.REF_META, VcfRepository.ALT_META);
+	}
+
+	@Override
+	public com.google.common.base.Optional<Entity> filterResults(Iterable<Entity> results, Entity annotatedEntity)
```
Mark-de-Haan: Put cursor right before optional and press mac+shift+m

brammiedema: done

PR #3414, opened 2015-07-17
Feature/1000g

brammiedema:
Fixed filter issues in clinvar and gonl annotators.


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/GoNLAnnotator.java
```diff
@@ -99,7 +126,7 @@ protected Object getResourceAttributeValue(AttributeMetaData attr, Entity entity
 	Resource gonlresources()
 	{
 		MultiResourceConfig goNLConfig = new MultiResourceConfigImpl(GONL_CHROMOSOME_PROPERTY,
-				GONL_FILE_PATTERN_PROPERTY, GONL_ROOT_DIRECTORY_PROPERTY, GONL_OVERRIDE_CHROMOSOME_FILES_PROPERTY,
+		GONL_FILE_PATTERN_PROPERTY, GONL_ROOT_DIRECTORY_PROPERTY, GONL_OVERRIDE_CHROMOSOME_FILES_PROPERTY,
```
bartcharbon: are you sure that the new formatted code is correct? lines out strange @ github

PR #3467, opened 2015-08-13
Fix #3487 OntologyTermRepository.getChildren low performance

### LINKED ISSUE #3487: OntologyTermRepository.getChildren low performance
the getChildren method in OntologyTermRepository works as follows,

get the nodePath of the current ontologyTerm
get all related nodePaths that are similar to the current nodePath from OntologyTermNodePath table
get all the ontologyTerms which have references to those related nodePath entities
However the previous implementation was excruciatingly slow, it is because when getting related nodePaths from the table, the wrong operator 'Like' was used in the query, it`s translated into the ElasticSearch query such that it found everything from the table because of ngram query.

The problem has been solved by changing the Operator from 'Like' to 'Fuzzy_Match', because in Fuzzy_Match implementation, it uses query string syntax...

ChaoPang:
the getChildren method in OntologyTermRepository works as follows,
- get the nodePath of the current ontologyTerm
- get all related nodePaths that are similar to the current nodePath from OntologyTermNodePath table
- get all the ontologyTerms which have references to those related nodePath entities

However the previous implementation was excruciatingly slow, it is because when getting related nodePaths from the table, the wrong operator 'Like' was used in the query, it`s translated into the ElasticSearch query such that it found almost everything from the table because of ngram query. 

The problem has been solved by changing the Operator from 'Like' to 'Fuzzy_Match', because in Fuzzy_Match implementation, it uses query string syntax...

### DIFF
--- a/molgenis-ontology-core/src/main/java/org/molgenis/ontology/core/repository/OntologyTermRepository.java
+++ b/molgenis-ontology-core/src/main/java/org/molgenis/ontology/core/repository/OntologyTermRepository.java
@@ -17,6 +17,8 @@ import org.molgenis.data.DataService;
 import org.molgenis.data.Entity;
 import org.molgenis.data.MolgenisDataAccessException;
 import org.molgenis.data.Query;
+import org.molgenis.data.QueryRule;
+import org.molgenis.data.QueryRule.Operator;
 import org.molgenis.data.support.QueryImpl;
 import org.molgenis.ontology.core.meta.OntologyTermMetaData;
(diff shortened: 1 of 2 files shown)


dennishendriksen:
Is this issue only related to performance? Or does it change mapping results as well? In that case please create a GitHub issue (so that it will end up in the release notes), attach it to the current milestone and refer to it in the comments of this PR.

ChaoPang:
It is only related to performance and the matching results will not change.

PR #3474, opened 2015-08-17
Fix/3458 Annotators strip VCF flag fields

bartcharbon:


molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/VcfRepository.java
```diff
@@ -177,6 +186,15 @@ public String apply(Allele allele)
 			{
 				val = null;
 			}
+			// if a flag field exists in the line, than this field is true, although the value is null
```
fdlk: than -> then :)

fdlk: But more to the point: A little bit up I see that it's possible to check the VcfMetaInfo for type FLAG. Are there also non-flag booleans that would need to be written as NULL rather than TRUE when absent?

bartcharbon: In this part of the function we have an vcfInfo in our hands instead of a vcfMetaInfo. So additional code would be needed to find the correct metainfo to check for flag...

bartcharbon:
once again in a fight with git
I'll close this request and make a new one

PR #3475, opened 2015-08-17
Fix/1810 (defaultValues)

### LINKED ISSUE #1810: EMX misses DefaultValue
AttributeMetaData has an method getDefaultValue().
This is not (yet?) supported by EMX

fdlk:
Fixes #1810 (defaultValues).
After asking Tommy, I've numbered the migration step nr 15. Number 14 was already taken. I strongly suspect that both migrations are independent.


bartcharbon:
Github marks a lot of unchanged lines as "changed" might be worth looking into sometime...

bartcharbon:
Let's discuss future features for the default value @ Grooming or EOS.

molgenis-data-migrate/src/main/java/org/molgenis/data/version/v1_9/Step15AddDefaultValue.java
```diff
@@ -0,0 +1,139 @@
...
+	{
+		return String.format("UPDATE attributes SET defaultValue = '%s' WHERE name = '%s' "
+				+ "AND identifier IN (SELECT attributes FROM entities_attributes ea WHERE ea.fullName = '%s');",
+				defaultValue, attributeName, entityName);
+	}
+
+	@Override
+	public void upgrade()
+	{
+		LOG.info("Updating metadata from version 14 to 15");
```
bartcharbon: I understand from the comments that the number is discussed, however it will now break continues integration on molgenis01 if merged right away. Step 15 will be executed before the not yet merged 14, so step 14 will never execute when merged.

PR #3503, opened 2015-08-20
VcfToEntity. UnitTest flag fields

fdlk:


molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/format/VcfToEntity.java
```diff
@@ -0,0 +1,327 @@
...
+			{
+				entity.set(VcfRepository.getInfoPrefix() + info.getId(), false);
+			}
+		}
+
+		for (VcfInfo vcfInfo : vcfRecord.getInformation())
+		{
+			System.out.println(vcfInfo.getKey());
+			System.out.println(entityMetaData.getAtomicAttributes());
+			System.out.println(entityMetaData.getAttribute("INFO_DF"));
```
joerivandervelde: Maybe printlns these should not be here?

fdlk: Yikes! You're right of course.

PR #3513, opened 2015-08-21
Feature/#3676/next uncurated button

jjettenn:


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/FirstAttributeMappingInfo.java
```diff
@@ -0,0 +1,49 @@
+package org.molgenis.data.mapper.controller;
+
+public class FirstAttributeMappingInfo
```
ChaoPang: Use AutoValue according to GCC developer guidelines....

PR #3514, opened 2015-08-21
Annotators: Fixed HPO annotator not working in UI

tommydeboer:
- Exposed right attributes
- Set attributes to type TEXT, because the phenotype strings can get very long
 - **Please add this edge case to the integration test by annotating a gene with loads of HPO terms**


bartcharbon:
I have local merge to fix this problem, I will push it if the other stuff on my branch is ready for it (partial commit not allowed during merge...)

PR #3519, opened 2015-08-25
Test TabixRepository + Fix HPO integration test

fdlk:
It should filter out rows that don't match the query and it should behave nicely when the TabixReader returns null.


molgenis-data-annotators/src/main/java/org/molgenis/data/annotator/tabix/TabixRepository.java
```diff
@@ -92,7 +99,11 @@ public EntityMetaData getEntityMetaData()
 			String line = iterator.next();
 			while (line != null)
 			{
-				builder.add(toEntity(line));
+				Entity entity = toEntity(line);
+				if (entity.getLong(VcfRepository.POS) == pos)
```
bartcharbon: I checked the tabix documentation and you can specify the index of the pos column. So it might be something else than "POS".
However, the tabixreader we use doesn't have a function to return which column it is as far as I can tell... so no better alternative available.

fdlk: Can add a posAttributeName field, would that work?

bartcharbon: Yes that would be a good solution I think, probably same story for CHROM...

fdlk:
Added the configurable chrom and pos attribute names as discussed.

PR #3521, opened 2015-08-25
fix SnpEffAnnotator + improve and extend unit test

bartcharbon:


molgenis-data-annotators/src/test/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotatorTest.java
```diff
@@ -129,5 +292,25 @@ public DataService dataService()
 			return mock(DataService.class);
 		}
 
+		@Bean
```
fdlk: Is that really necessary?

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotator.java
```diff
@@ -111,7 +119,7 @@ public static String getGeneNameFromEntity(Entity entity)
 		}
 		if (geneSymbol == null)
 		{
-			String annField = entity.getString(VcfRepository.getInfoPrefix() + "ANN");
+			String annField = entity.getString("ANN");
```
fdlk: Why does it end up in ANN now instead of INFO_ANN?

molgenis-data-annotators/src/test/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotatorTest.java
```diff
@@ -129,5 +292,25 @@ public DataService dataService()
...
+				when(jarRunner.runJar(SnpEffAnnotator.NAME, params, ResourceUtils.getFile("test-snpeff.vcf")))
+						.thenReturn(ResourceUtils.getFile("snpEffOutput.vcf"));
+
+			}
+			catch (Exception e)
+			{
+				e.printStackTrace();
+			}
+			return jarRunner;
+		}
```
fdlk: Is that really necessary, does a normal Mock not work?

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotator.java
bartcharbon: Annotators are not specific for VCF, the CmdLineAnnotator and VCFUtils should handle the prefix. The annotator should return a logical attributename

molgenis-data-annotators/src/test/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotatorTest.java
bartcharbon: ? this is a "normal" mock, right? try/catch is needed for the exceptions that can be caused by files

fdlk: What I mean, a normal untrained mock.
Can't you reset and train it in the beforeMethod or in the test itself?

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/CommandLineAnnotatorConfig.java
```diff
@@ -43,6 +45,12 @@ AnnotationService annotationService()
 		return new AnnotationServiceImpl();
 	}
 
+	@Bean
+	JarRunner jarRunner()
+	{
+		return new JarRunnerImpl();
+	}
+
```
fdlk: This one should sit in the shared config, not in the commandline-only config. Now the application no longer starts up.

bartcharbon:
Somehow my other pull request got mixed up in this one... should I spend time to fix this, or will you review the resulting version?

fdlk:
That's fine!

PR #3522, opened 2015-08-25
Use request.session for all requests, connection has to be made with 'with' statement

npklein:
to make sure that the password file gets cleaned up after use


molgenis-python/src/main/resources/python/security.py
```diff
@@ -1,29 +1,35 @@
 from getpass import getpass
 from pbkdf2 import PBKDF2
-import crypto
-import sys
-sys.modules['Crypto'] = crypto
-from Crypto.Cipher import AES
 import os
+import platform
+if platform.system() == 'Darwin':
```
fdlk: Bit specific, this

fdlk:

fdlk:
add to whitelist

dennishendriksen:
Could you merge with master? Lets have a chat about when to merge your stuff.

PR #3531, opened 2015-08-28
Feature/#3710/questionnaire

jjettenn:
Added some conditions to decide when the form must resets its state. Related to story "#3710"


molgenis-core-ui/src/main/resources/js/component/Form.js
```diff
@@ -49,14 +51,44 @@
...
 		componentWillReceiveProps : function(nextProps) {
-			var newState = {
-				errorMessages : {},
-				validate: false,
-				showModal: true,
-				submitMsg: null
-			};
+			var entity = this.props.entity;
+			var entityInstance = this.props.entityInstance;
+			var resetProps = false;
```
fdlk: maybe resetState is better name?

fdlk: Heh, I meant for the resetProps variable, not for newState variable. Cause it's the state that gets reset, not the props. Anyways, what do you think?

PR #3555, opened 2015-09-03
Fix #3368 Importing an entity with the same name but different casing will result in a locked importer

### LINKED ISSUE #3368: Importing an entity with the same name but different casing will result in a locked importer
## Reproduce
- Upload an EMX containing the entity Gender
- Upload an EMX containing the entity gender
## Expected

FAIL with good error message: there is already a entity with name 'Gender' that conflicts with 'gender'
(while our front-end is case-sensitive, our backend may not be hence we are strict in entity/attribute name checking).
## Observed

Importer is locked
Entity gender already exists
Entity Gender is unknown
Entity gender already exists
Entity Gender is unknown

and it goes on and on...

Looks like the part that checks if an entity exists is not case sensitive, and the part that wants to get the entity is case sensitive.

Mark-de-Haan:
I fixed it through a validation check before we insert parsed data into repositories. This prevents the casing collision from ever happening, preventing roll backs and such.

Note: The casing issue needs to be fixed but its a story on its own and too big for just a bug fix.


molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportedDataToBackendCompatibilityValidator.java
```diff
@@ -0,0 +1,45 @@
...
+{
+	@Autowired
+	DataService dataService;
+
+	public ImportedDataToBackendCompatibilityValidator(DataService dataService)
+	{
+		this.dataService = dataService;
+	}
+
+	public void validate(ImmutableCollection<EntityMetaData> immutableCollection)
```
fdlk: can relax the interface a bit here, to for instance Iterable<EntityMetaData>, I think?

molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportedDataToBackendCompatibilityValidator.java
```diff
@@ -0,0 +1,45 @@
...
+	{
+		immutableCollection.forEach(entity -> checkIfEntityExists(entity));
+	}
+
+	private void checkIfEntityExists(EntityMetaData entity)
+	{
+		if (dataService.hasRepository(entity.getName()))
+		{
+			System.out.println("Test if get here");
```
fdlk: sysout!

molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportedDataToBackendCompatibilityValidator.java
```diff
@@ -0,0 +1,45 @@
...
+			throw new MolgenisDataException("Entity with name [" + entity.getName()
+					+ "] already exists. Casing is ignored!");
+		}
+		else if (dataService.hasRepository(entity.getName().toLowerCase()))
+		{
+			System.out.println("Test if get here");
```
fdlk: sysout!

molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportWriter.java
```diff
@@ -78,6 +78,9 @@ public ImportWriter(DataService dataService, PermissionSystemService permissionS
 	@Transactional
 	public EntityImportReport doImport(EmxImportJob job)
 	{
+		ImportedDataToBackendCompatibilityValidator validator = new ImportedDataToBackendCompatibilityValidator(dataService);
+		validator.validate(job.parsedMetaData.getEntities());
```
fdlk: Not entirely sure, but doesn't the validator need privileges to check ALL metadata for the presence of one of these entities. I.e. shouldn't this check be in the runAsSystem part?

molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportedDataToBackendCompatibilityValidator.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	public void validate(ImmutableCollection<EntityMetaData> immutableCollection)
+	{
+		immutableCollection.forEach(entity -> checkIfEntityExists(entity));
+	}
+
+	private void checkIfEntityExists(EntityMetaData entity)
+	{
+		if (dataService.hasRepository(entity.getName()))
```
fdlk: runAsSystem would be needed a little higher if dataService.hasRepository answers only true for those you can actually read.

molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportWriter.java
Mark-de-Haan: Agreed, you need to check entities existing in the database, some of them not owned by you probably

molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportedDataToBackendCompatibilityValidator.java
Mark-de-Haan: I think the dataservice holds all repositories, not only those you have rights to?

fdlk:
Nicely done!

PR #3562, opened 2015-09-03
Fix #3520 TabixReader is not thread-safe 

### LINKED ISSUE #3520: TabixReader is not thread-safe
If two methods use the same TabixReader concurrently, they'll get each other's results.
The TabixReader's Iterators share one and the same InputStream. Therefore if you read from them concurrently, the results will influence each other.

We already synchronize access to the query() methods of the tabix Repository classes which means only one query() method is ever active per repository. But I am not sure if that's completely fool proof.
One mistake that I do see is that TabixRepository's iterator() method calls readLine() on the repository which will eat up results from people who are annotating. (Plus I don't see where the iterator is reset to begin with! I suspect it just feeds you increasing pages from the repository even if you use it single-user mode)

The TabixReader checks if the position matches so most often you get less results than you should or duplicate results.
But I *do* at times see results that are combinations of two rows, so some of the columns are invalid!

bartcharbon:
Besides fixing the issue I also changed a lot of names in the TabixReader in an attempt to increase readability


molgenis-data-annotators/src/main/java/org/molgenis/data/annotator/tabix/TabixReader.java
```diff
@@ -369,38 +364,38 @@ public String next() throws IOException
...
-					if (intv.tid != tid || intv.beg >= end) break; // no need to proceed
-					else if (intv.end > beg && intv.beg < end) return s; // overlap; return
+					if (intv.internalChromosomeID != internalChromosomeID || intv.beginPosition >= endPosition) break; // no need to proceed
+					else if (intv.end > beginPosition && intv.beginPosition < endPosition) return s; // overlap; return
 				}
 				else break; // end of file
 			}
-			iseof = true;
+			isEndOfFile = true;
+			blockCompressedInputStream.close();
```
fdlk: There's no guarantee that the iterator will actually be fully read right up to the end.
Until it is, the newly created stream will remain open.

PR #3564, opened 2015-09-03
Fix #3516 Questionnair does not submit

### LINKED ISSUE #3516: Chr6 submit questionnaire does not work
Actual
When submitting a chromosome6 questionnaire after filling in all the required fields, nothing happens.

Expected
I get a 'thank you' page. 

Reproduce
- login to Molgenis69 as admin
- fill in all the required fields of the questionnaire 
- press submit button

erwinwinder:


molgenis-data/src/main/java/org/molgenis/data/transaction/TransactionConfig.java
```diff
@@ -16,10 +17,14 @@
 	@Autowired
 	private MolgenisTransactionManager transactionManager;
 
+	// This cannot be a app setting because apps settings are not loaded yet when we use it
+	@Value("${use.transaction.log:true}")
```
dennishendriksen: Why have an option to disable transaction log? Transaction log is required in order for the rollback mechanism to work, right? What problem were you trying to solve?

erwinwinder: Transaction log logs wich transactions per user and is used for checking concurrent modifications, is not always needed and takes a lot of time

dennishendriksen: It is not used for rollback in case something goes wrong?

erwinwinder: Nope

dennishendriksen: If it takes a lot of time then we should improve the performance of logging transactions (which doesn't sound like a heavy job). This will improve the all over performance of the web application. This solution feels like a workaround for the actual problem.

erwinwinder: Yep, but that is another story

PR #3579, opened 2015-09-04
M3504 Add VCF validation step to VCF writing functions

Mark-de-Haan:
CmdLineAnnotator class is expanded to call the vcf tools vcf-validator if the user enters the --validate parameter. Perl configuration is default molgenis server style but can also be set manually.


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/CmdLineAnnotator.java
```diff
@@ -40,9 +40,17 @@
...
+
 	@Autowired
 	private ApplicationContext applicationContext;
 
+	private Boolean validate = false;
+
+	// Default settings for running vcf-validator
+	private String userHome = System.getProperty("user.home");
+	private String perlDirectory = "/usr/bin/perl";
+	private String vcfToolsDirectory = userHome + "/.molgenis/vcf-tools/";
```
fdlk: *nix specific glueing of directory names

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/CmdLineAnnotator.java
```diff
@@ -63,10 +71,12 @@ public void run(String[] args) throws Exception
...
 							+ "\n"
 							+ "Typical usage to annotate a VCF file:\n"
-							+ "\tjava -jar CmdLineAnnotator.jar [Annotator] [Annotation source file] [input VCF] [output VCF] [output attributes (optional, default:all attributes)].\n"
-							+ "\tExample: java -Xmx4g -jar CmdLineAnnotator.jar gonl GoNL/release5_noContam_noChildren_with_AN_AC_GTC_stripped/ Cardio.vcf Cardio_gonl.vcf GoNL_GTC GoNL_AF\n"
+							+ "\tjava -jar CmdLineAnnotator.jar [Annotator] [Annotation source file] [input VCF] [output VCF] [validate flag]=[perl location]|[vcf-tools directory] [output attributes (optional, default:all attributes)].\n"
+							+ "\tExample: java -Xmx4g -jar CmdLineAnnotator.jar gonl GoNL/release5_noContam_noChildren_with_AN_AC_GTC_stripped/ Cardio.vcf Cardio_gonl.vcf --validate=/usr/bin/perl,~/.molgenis/vcf-tools/ GoNL_GTC GoNL_AF\n"
 							+ "\n"
 							+ "Help:\n"
+							+ "\tThe default expected location of the perl executable is: /usr/bin/perl\n"
+							+ "\tThe default expected location of thevcf tool directory is: ~/.molgenis/vcf-tools\n\n"
```
fdlk: thevcf -> the vcf

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/CmdLineAnnotator.java
```diff
@@ -77,6 +87,37 @@ public void run(String[] args) throws Exception
 			return;
 		}
 
+		Integer nrOfArguments = 4;
+		if (args.length > 4)
+		{
+			nrOfArguments = 5;
```
fdlk: I think it's now time to start using a proper command line parser

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/VcfValidator.java
```diff
@@ -0,0 +1,106 @@
...
+	 * Validation method that calls the perl executable of the vcf-validator. Logs vcf validation output into a log file
+	 * 
+	 * @param vcfFile
+	 * @return Message
+	 */
+	public String validateVCF(File vcfFile)
+	{
+		try
+		{
+			String vcfValidator = vcfToolsDirectory + "perl/vcf-validator";
```
fdlk: File.separator

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/VcfValidator.java
```diff
@@ -0,0 +1,106 @@
...
+				Process proc = processBuilder.start();
+
+				InputStream inputStream = proc.getInputStream();
+				Scanner scanner = new Scanner(inputStream);
+
+				String line = "";
+				Integer errorCount = null;
+				Pattern p = Pattern.compile("(\\d+)\\s*errors\\s*total");
+
+				File logFile = new File("vcf-validation.log");
```
fdlk: Maybe use the output file's name here, so that the validation logs don't overlap

fdlk:
Web version now refuses to build cause it doesn't have the perlLocation. Should use settings, probably? (Is VcfWriter used at all by the web version)

Mark-de-Haan:
I can build but I cant run jetty because VcfValidator has no qualified bean ..

fdlk:
Which is odd cause the only one who needs it is the CmdLineAnnotator.
I think the cause is that it sits in a scanned package and has the `@Component` annotation so the bean CmdLineAnnotator will always get created.

Can probably solve it by removing the @Component annotation from CmdLineAnnotator and adding a `@Bean` method in the CmdLineAnnotatorConfig.

Mark-de-Haan:
Yeah I fixed it by removing autowire with new VcfValidator. But then I changed to what you commented @fdlk 

It works now with the autowiring

fdlk:
Still cannot build:

```
Failed tests: 
  chain(org.molgenis.data.annotation.entity.impl.AnnotatorChainingIT): Error creating bean with name 'commandLineAnnotatorConfig': Injection of autowired dependencies failed; nested exception is org.springframework.beans.factory.BeanCreationException: Could not autowire field: private java.lang.String org.molgenis.data.annotation.cmd.CommandLineAnnotatorConfig.perlLocation; nested exception is java.lang.IllegalArgumentException: Could not resolve placeholder 'perl-location' in string value "${perl-location}"
  fitcon(org.molgenis.data.annotation.cmd.CmdLineAnnotatorIT): File 'target/out-fitcon-test.vcf' does not exist
  gonl(org.molgenis.data.annotation.cmd.CmdLineAnnotatorIT): File 'target/out-gonl-test.vcf' does not exist
  hpo(org.molgenis.data.annotation.cmd.CmdLineAnnotatorIT): File 'target/out-hpo-test.vcf' does not exist
  hpoTermsOnly(org.molgenis.data.annotation.cmd.CmdLineAnnotatorIT): File 'target/out-hpo-test.vcf' does not exist
  exac(org.molgenis.data.annotation.cmd.EdgeCaseExacCmdLineAnnotatorIT): File 'target/out-exac-test-edgecases.vcf' does not exist
```

fdlk:

fdlk:
add to whitelist

tommydeboer:
Few nit-picks about the new help texts and error messages:

- No installation instructions for any annotator, even though the manual promises it:

```
To get detailed description and installation instructions for a specific annotator:
java -jar CmdLineAnnotator.jar -a [Annotator]
```

- Might be better to put the -v flag somewhere other than at the end in the example, because now it is confusing: are GoNL_GTC GoNL_AF the arguments for -v?
    
``` Example: java -Xmx4g -jar CmdLineAnnotator.jar -a gonl -s     GoNL/release5_noContam_noChildren_with_AN_AC_GTC_stripped/ -i Cardio.vcf -o Cardio_gonl.vcf -v GoNL_GTC GoNL_AF ```

- These two commands show exactly the same information, but the manual says otherwise:
```
To get detailed description and installation instructions for a specific annotator:
java -jar CmdLineAnnotator.jar -a [Annotator]
To check if an annotator is ready for use:
java -jar CmdLineAnnotator.jar -a [Annotator] -s [Annotation source file]
```

- Unclear which file to use for each annotator (cadd/, cadd.tsv.gz.tbi, cadd.tsv.gz)
- Unclean message when using wrong file:

```java.io.FileNotFoundException: /Users/tommy/.molgenis/omx/data/annotation_resources/cadd/cadd.tsv.gz.tbi.tbi (No such file or directory)```

- Errors & Stacktraces when queries fail on a resource. Also see #3662

```
Now starting to process the data.
09:07:09.093 [main] ERROR o.m.d.a.tabix.TabixRepository - Error reading from tabix reader for query: 1:10050000-10050000
09:07:09.110 [main] ERROR o.m.d.a.tabix.TabixRepository - Error reading from tabix reader for query: 1:10050001-10050001
09:07:09.111 [main] ERROR o.m.d.a.tabix.TabixRepository - Error reading from tabix reader for query: 1:10050002-10050002
09:07:09.112 [main] ERROR o.m.d.a.tabix.TabixRepository - Error reading from tabix reader for query: 1:10050003-10050003
09:07:09.113 [main] ERROR o.m.d.a.tabix.TabixRepository - Error reading from tabix reader for query: 1:10050004-10050004
09:07:09.114 [main] ERROR o.m.d.a.tabix.TabixRepository - Error reading from tabix reader for query: 1:10050005-10050005
09:07:09.115 [main] ERROR o.m.d.a.tabix.TabixRepository - Error reading from tabix reader for query: 1:10050006-10050006
All done!
```
  
- Nice warning, but I don’t get the chance to stop the annotation and now I lose my 20Gb file that took 12 days to annotate!
```
WARNING: Output VCF file already exists at /Users/tommy/git/molgenis/molgenis-data-annotators/target/testcadd.vcf
```

Mark-de-Haan:
@tommydeboer the description on which file should be uploaded SHOULD be a part of the annotator info IMO

PR #3580, opened 2015-09-07
M3454 Mapping service auto template-based algorithms

dennishendriksen:


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/AlgorithmServiceImpl.java
```diff
@@ -87,51 +87,53 @@ public void autoGenerateAlgorithm(EntityMetaData sourceEntityMeta, EntityMetaDat
...
+			String algorithm = null;
 
-			// determine source unit
-			Unit<? extends Quantity> sourceUnit = unitResolver.resolveUnit(source, sourceEntityMeta);
+			AttributeMetaData source = entry.getKey();
 
-			String algorithm = null;
-			if (sourceUnit != null)
+			// use existing algorithm template if available
+			AlgorithmTemplate algorithmTemplate = algorithmTemplateService
```
ChaoPang: The local variable of this method (matches) has already contained all the suggested matches from the source entity, Iterable<ExplainedQueryString>> matches = semanticSearchService
 				.findAttributes(sourceEntityMeta, targetEntityMeta, targetAttr);		 				.findAttributes(sourceEntityMeta, targetEntityMeta, targetAttr);

It would be redundant to call semanticSearchService method in the AlgorithmTemplateService again, would that be possible to pass the local variable matches to AlgorithmTemplateService directly?

PR #3592, opened 2015-09-08
Fix #3585 OntologyTermRepository returns almost all ontology terms if the search term is a pure number

### LINKED ISSUE #3585: OntologyTermRepository returns almost all ontology terms if the search term is a pure number
OntologyTermRepository is responsible for talking with dataService to return typed ontology terms. If the search term is a number only, almost all of the ontology terms will be retrieved. 

How to reproduce:
- login at molgenis01
- load the unit ontology that can be found at https://drive.google.com/open?id=0B7qHi41THTs_eVYwQWZ6QnFRY3M
- Go to RestClient in Chrome
- Type in the url https://molgenis01.target.rug.nl/menu/main/tagwizard/getontologyterms
- Method POST
- request body (you might want to adjust the ontology ID accordingly on molgenis01)
{
    "ontologyIds": ["AAAACT5L5ZDNYXRFAN43UOAAAE"],
    "searchTerm": "1"
}
- almost all the ontology terms from unit ontology will be retrieved. (372 out 373 will be retrieved)

ChaoPang:
- Updated the findOntologyTerm method to limit searching in ontologyTermSynonym field because it already contains all relevant information
- Created a findExactOntologyTerm method to filter out all the ontology terms that are not exact matches to the search terms

### DIFF
--- a/molgenis-ontology-core/src/main/java/org/molgenis/ontology/core/repository/OntologyTermRepository.java
+++ b/molgenis-ontology-core/src/main/java/org/molgenis/ontology/core/repository/OntologyTermRepository.java
@@ -95,7 +95,7 @@ public class OntologyTermRepository
 				return true;
 			}
 		}
-		if (lowerCaseSearchTerms.contains(ontologyTerm.getLabel()))
+		if (lowerCaseSearchTerms.contains(ontologyTerm.getLabel().toLowerCase()))
 		{
 			return true;
 		}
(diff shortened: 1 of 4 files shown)


molgenis-ontology-core/src/main/java/org/molgenis/ontology/core/repository/OntologyTermRepository.java
```diff
@@ -67,6 +68,41 @@ public OntologyTermRepository(DataService dataService)
...
+	{
+		Set<String> lowerCaseSearchTerms = terms.stream().map(term -> term.toLowerCase()).collect(Collectors.toSet());
+		for (String synonym : ontologyTerm.getSynonyms())
+		{
+			if (lowerCaseSearchTerms.contains(synonym.toLowerCase()))
+			{
+				return true;
+			}
+		}
+		if (lowerCaseSearchTerms.contains(ontologyTerm.getLabel()))
```
jjettenn: To lower case

PR #3629, opened 2015-09-10
[Master] M3677 As a biobankconnect user I would like to change the 'key words' on the mapping screen so the suggested attributes can be updated at real time

ChaoPang:


fdlk:

fdlk:
add to whitelist

PR #3640, opened 2015-09-10
[1.9] fix #2768: When CHROM / POS columns are hidden and you click a geneti…

### LINKED ISSUE #2768: When CHROM / POS columns are hidden and you click a genetic data row, you get "viewFeatures called with silly scale"
Happened right now after deleting TypeTest entity on molgenis62, when data explorer responded by returning to testSamples_KEGG entity.

Also seen it during the automated annotator test.

No clear way to reproduce it yet.

bartcharbon:
…c data row, you get "viewFeatures called with silly scale"

### DIFF
--- a/molgenis-dataexplorer/src/main/resources/js/dataexplorer-data.js
+++ b/molgenis-dataexplorer/src/main/resources/js/dataexplorer-data.js
@@ -64,7 +64,7 @@
 			onRowDelete: onDataChange,
 			onRowEdit: onDataChange,
 			onRowInspect: onRowInspect,
-			onRowClick: doShowGenomeBrowser() ? onRowClick : null,
+			onRowClick: (doShowGenomeBrowser() && isGenomeBrowserAttributesSelected()) ? onRowClick : null,
 			onSort: function(e) {
 				tableSort = {
 					'orders' : [ {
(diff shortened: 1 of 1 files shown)


molgenis-dataexplorer/src/main/resources/js/dataexplorer-data.js
```diff
@@ -148,6 +148,12 @@
 		return molgenis.ie9 !== true && genomebrowserStartAttribute !== undefined && genomebrowserChromosomeAttribute !== undefined && molgenis.dataexplorer.settings["data_genome_browser"] !== false;
 	}
 
+	//used to determine if the rowclick should be available
+	function isGenomeBrowserAttributesSelected(){
+    		var attributes = molgenis.dataexplorer.getSelectedAttributes();
+            return (attributes.indexOf(genomebrowserChromosomeAttribute)!=-1) && (attributes.indexOf(genomebrowserStartAttribute)!=-1);
+    }
+
```
fdlk: cunning! :)

fdlk:
But still broken I fear.

fdlk:

fdlk:
add to whitelist

bartcharbon:
Seen it work locally, rowclick should get disabled when you deselect one of the needed attributes(chrom or pos)... 
can you describe what you did to reproduce the error still?

bartcharbon:
reviewing the solution I think the fix does not cover the situation where you start out without the POS and CHROM attributes selected. The solution "fires" on the change of the attributeselection, can this be the case for your test?

fdlk:
that would make sense, yes. I think I reload the page with the chrom/pos attributes deselected

bartcharbon:
updated the fix: added same code to the initial properties for the table

PR #3642, opened 2015-09-10
[1.9] M3504 Add VCF validation step to VCF writing functions

fdlk:


Mark-de-Haan:
FAILED: gonl
java.io.FileNotFoundException: File 'target/out-gonl-test.vcf' does not exist

Tests still fail :(

fdlk:
That test is broken in 1.9, we broke it when I accepted your pull request #3648 . Gonl annotator no longer works in 1.9.

Mark-de-Haan:
Then we remove @test annotation from gonl and accept?

bartcharbon:
Why does this need to be added to 1.9? This is functionality right? not a bugfix.

Mark-de-Haan:
Reasoning was that; we want to make 1.9 cmdline annotators usable for analysis team. This feature makes the cmdline annotators 10x more user friendly, and less prone to bugs because of proper argument parsing

PR #3651, opened 2015-09-11
[1.9] Fix #3643 VCF writer should place GT field first when writing samples

### LINKED ISSUE #3643: VCF writer should place GT field first when writing samples
The VCF writer should write out the GT field first in each sequence of 'FORMAT' fields per sample. So instead of, for example:

"AB:AD:DP:GQ:GT:MIN_DP:MQ0:PGT:PID:PL:RGQ:SB" followed by sample data "0.750:6,2:6:61:0/1:.:0:.:.:61,0,152:.:."

must instead be "GT:AB:AD:DP:GQ:MIN_DP:MQ0:PGT:PID:PL:RGQ:SB" followed by "0/1:0.750:6,2:6:61:.:0:.:.:61,0,152:.:.".

This behaviour is expected by other tools that will throw errors when GT is not placed first. It is also described as mandatory by both VCF specification for 4.1 and 4.2.

See:
https://samtools.github.io/hts-specs/VCFv4.1.pdf
https://samtools.github.io/hts-specs/VCFv4.2.pdf

dennishendriksen:


fdlk:
Will this also move the GT format header to the top?

dennishendriksen:
No, this will not move the GT format header, this is not required according to the VCF specification.

PR #3669, opened 2015-09-14
[master] Fixed menu migration not migrating old menu

tommydeboer:


tommydeboer:
Tested in 1.9 and it works

PR #3676, opened 2015-09-15
fix #3655: The SnpEff annotator contains a field that fails with larg…

### LINKED ISSUE #3655: The SnpEff annotator contains a field that fails with large inserts or deletes
The SnpEff annotator contains a field that fails with large inserts or deletes:
DefaultAttributeMetaData HGVS_c = new DefaultAttributeMetaData(HGVS_C, STRING);

This should be TEXT.

bartcharbon:
…e inserts or deletes


bartcharbon:
Also the new Eclipse formatter was applied to this class

Mark-de-Haan:
@bartcharbon Conflicts with master

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotator.java
```diff
@@ -370,42 +368,43 @@ else if (ann_field[2].startsWith("NMD="))
...
 
 			DefaultAttributeMetaData rank_total = new DefaultAttributeMetaData(RANK_TOTAL, STRING);
-			rank_total
-					.setDescription("Exon or Intron rank / total number of exons or introns(source:http://snpeff.sourceforge.net)");
+			rank_total.setDescription(
+					"Exon or Intron rank / total number of exons or introns(source:http://snpeff.sourceforge.net)");
 			attributes.add(rank_total);
 
-			DefaultAttributeMetaData HGVS_c = new DefaultAttributeMetaData(HGVS_C, STRING);
+			DefaultAttributeMetaData HGVS_c = new DefaultAttributeMetaData(HGVS_C, TEXT);
```
fdlk: And this is where it happens, right?

PR #3679, opened 2015-09-15
M3698 As annotator user I want to configure annotator settings in UI

erwinwinder:


bartcharbon:
I get:
------------ 
org.molgenis.data.validation.MolgenisValidationException: The attribute 'filepattern' of entity 'settings_gonl' can not be null. (row 2).The attribute 'overrideChromosomeFiles' of entity 'settings_gonl' can not be null. (row 2).The attribute 'chromosomes' of entity 'settings_gonl' can not be null. (row 2)
	at org.molgenis.data.validation.RepositoryValidationDecorator.validate(RepositoryValidationDecorator.java:132) ~[molgenis-data-validation-1.10.0-SNAPSHOT.jar:1.10.0-SNAPSHOT]
----------

Maybe the migration doesn't cover the case where the old setting is not present?

erwinwinder:
Ok, made the attribiutes nillable

PR #3682, opened 2015-09-15
Fix #3157 After deleting entities plus metadata, they still remain visible in the model registry

### LINKED ISSUE #3157: After deleting entities plus metadata, they still remain visible in the model registry
Upload an EMX file, delete the entities in the data explorer.
Then access the package in the model registry.

Expected:
The package is empty

Actual:
The entities that got deleted, and their attributes, are still listed.
And an error message is shown in the alert: Error! An error occurred. Please contact the administrator.

Mark-de-Haan:
- Now updating packageCache after deleting entities from repository
- Formatting


fdlk:
When I delete a single entity from the base package, the model registry no longer crashes but now incorrectly claims that the base package is empty

Mark-de-Haan:
Refreshing the caches does not work if executed within a transaction. Bug within the TransactionManager?? @erwinwinder @fdlk

fdlk:
Fixed in #3738

PR #3689, opened 2015-09-16
Fix #3599 Annotating as a normal user will result in no permission on attributes and entities. Using the copy function will fail on no write permissions for packages

### LINKED ISSUE #3599: Annotating as a normal user will result in no permission on attributes and entities. Using the copy function will fail on no write permissions for packages
Can be worked around with permissions.
RunAsSystem would be nice

erwinwinder:


fdlk:
retest this please

PR #3696, opened 2015-09-16
Fix #3637 Clicking the 'map'-tab gives an error when no xref attribute has been selected yet

### LINKED ISSUE #3637: Clicking the 'map'-tab gives an error when no xref attribute has been selected yet
## To reproduce:
1. Make a mapping project with at least one xref in it
2. Edit the xref mapping
3. Don't select an attribute yet in the mapping screen, but go the the 'map'-tab

## Expected:
Empty tab is shown or tab is not clickable.

## Actual:
image

Mark-de-Haan:
Added IDs to the script and map tab of the mapping column
Added a JS check that disables the map tab when there are no attributes selected.

This prevents errors when the map tab is trying to load drop down values but there is not attribute selected to get the values from


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/VcfValidator.java
```diff
@@ -60,7 +61,7 @@ public String validateVCF(File vcfFile)
 				logFile.createNewFile();
 			}
 
-			FileWriter fileWriter = new FileWriter(logFile.getAbsoluteFile(), true);
+			FileWriterWithEncoding fileWriter = new FileWriterWithEncoding(logFile.getAbsoluteFile(), "UTF-8", true);
```
dennishendriksen: Why not use standard Java IO classes?
BufferedWriter out = new BufferedWriter(new OutputStreamWriter(new FileOutputStream("myfile.txt"), "UTF-8"));

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/VcfValidator.java
```diff
@@ -90,6 +91,7 @@ public String validateVCF(File vcfFile)
 			bufferedWriter.close();
 
 			scanner.close();
+			errorScanner.close();
```
dennishendriksen: Put close calls in finally block such that streams are closed in case an exception occurs.

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/VcfValidator.java
```diff
@@ -4,7 +4,6 @@
 
 import java.io.BufferedWriter;
```
dennishendriksen: Next time, please put this commit in a separate pull request.

dennishendriksen:
Fix tested successfully.

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/VcfValidator.java
Mark-de-Haan: I thought I did >.<

Mark-de-Haan:
@dennishendriksen Fixed the two pull request on one issue

PR #3697, opened 2015-09-16
M3741 As dataexplorer user I would like to load more rows

tommydeboer:


dennishendriksen:
You solution is data explorer specific, why not add the dropdown to the Table component so the row selection can used everywhere where you use a table (e.g. compute UI will benefit from this).

molgenis-core-ui/pom.xml
```diff
@@ -174,6 +174,7 @@
 	                            <jsSourceFile>/component/Spinner.js</jsSourceFile>
 	                            <jsSourceFile>/component/Table.js</jsSourceFile>
 	                            <jsSourceFile>/component/TextArea.js</jsSourceFile>
+	                            <jsSourceFile>/component/SelectBox.js</jsSourceFile>
```
dennishendriksen: Would you mind keeping the list in alphabetic order?

molgenis-core-ui/src/main/resources/js/component/SelectBox.js
```diff
@@ -0,0 +1,42 @@
...
+	var option = React.DOM.option;
+	
+	/**
+	 * @memberOf component
+	 */
+	var SelectBox = React.createClass({
+		mixins: [molgenis.ui.mixin.DeepPureRenderMixin],
+		displayName: 'SelectBox',
+		propTypes: {
+			options: React.PropTypes.array,
```
dennishendriksen: .arrayOf(React.PropTypes.object)

molgenis-core-ui/src/main/resources/js/component/SelectBox.js
```diff
@@ -0,0 +1,42 @@
...
+			onChange: React.PropTypes.func
+		},
+		getDefaultProps: function(){
+			return {
+				options: [{value: "val1", text: "option1"}]
```
dennishendriksen: by default empty array

molgenis-core-ui/src/main/resources/js/component/SelectBox.js
```diff
@@ -0,0 +1,42 @@
...
+			}
+		},
+		render: function() {
+			return (
+				div(null,
+					select({className: "form-control", onChange: this.props.onChange}, this.props.options.map(function(item, i){
+						return(
+							option({value: item.value, key: i}, item.text)
```
dennishendriksen: I have a vague recollection that the value of key should be a string. If that is the case then i.toString().

molgenis-core-ui/src/main/resources/js/component/Table.js
```diff
@@ -221,6 +234,9 @@
 		},
 		_handlePageChange: function(e) {
 			this._refreshData(this.props, _.extend({}, this.state, {start: e.start}));
+		},
+		_handleRowsPerPageChange: function(e) {
+			this.setProps({maxRows: e.target.value});
```
dennishendriksen: maxRows should be moved to component state (see https://facebook.github.io/react/docs/interactivity-and-dynamic-uis.html#what-should-go-in-state). Transfer the props passed to the component on creation to state.

molgenis-core-ui/src/main/resources/js/component/SelectBox.js
tommydeboer: You can use either.

PR #3708, opened 2015-09-18
Fix #3705 VCF import slow

### LINKED ISSUE #3705: VCF import slow
Mail from Jouri:

I was building a nice dataset for the BMB workshop. The idea: 3 individuals from 1000G (a Brit, a Finn and a Chinese), 2 of them having a very pathogenic variant 'spiked in', chromosome 7 data only. Up to the participants to find out who is ill, and which disease they have.

So! genomic chr. 7 variants of the 3 individuals was a 691 MB VCF file so I decided to filter the data down dramatically using CGD panel BED file (=a clinical subset of the exome only). Result: a 3 MB VCF file, and when gzipped 500 kB. (see attached) Nice and small, eh?

Two findings:

On MOLGENIS release 1.9 importing takes 2 minutes which is (I think) embarrassingly slow... but for the workshop, work-able.

16:39:37.992 [pool-11-thread-1] INFO  org.molgenis.data.importer.ImportJob - Import started
16:41:36.870 [pool-11-thread-1] INFO  org.molgenis.data.importer.ImportJob - Import finished in 118878 msec.

On MOLGENIS "master" version this file won't import at all - it just keeps going and I quit after 30 minutes. Apparently some recent changes cause a massive slow down.

Data file:

https://drive.google.com/open?id=0B4_AOpOI8V7GVTJyWVpzWFRrOFE

erwinwinder:


dennishendriksen:
Are there unit tests for all methods for both below threshold and above threshold cases?

PR #3714, opened 2015-09-21
Fix #3663 Importing two questionnaires after each other results in a dead data-explorer

### LINKED ISSUE #3663: Importing two questionnaires after each other results in a dead data-explorer
To reproduce:

1. Import the "test questionnaire" file
2. Try to import the latest version of the Chromosome 6 questionnaire (Can also use an old one if it contains a Gender entity).
3. Wait until you think that it is taking to long. (~3 minutes)
4. Push the "Restart" button.

Expected: import is correctly rolled back

Actual: data-explorer is returning a 500 exception saying that the Gender Entity does not exists.

jjettenn:


molgenis-data-import/src/main/java/org/molgenis/data/importer/OptionsWizardPage.java
```diff
@@ -56,14 +56,8 @@ public String getTitle()
 	@Override
 	public String handleRequest(HttpServletRequest request, BindingResult result, Wizard wizard)
 	{
-		if (!(wizard instanceof ImportWizard))
-		{
-			throw new RuntimeException("Wizard must be of type '" + ImportWizard.class.getSimpleName()
-					+ "' instead of '" + wizard.getClass().getSimpleName() + "'");
-		}
-
+		ImportWizardUtil.validateImportWizard(wizard);
```
fdlk: Why not simply cast it to ImportWizard here?

molgenis-data-import/src/main/java/org/molgenis/data/importer/PackageWizardPage.java
```diff
@@ -22,8 +53,54 @@ public String getTitle()
...
+				ImportService importService = importServiceFactory.getImportService(importWizard.getFile(),
+						repositoryCollection);
+
+				// Do integration test only if there are no previous errors found
+				if (!importWizard.getEntitiesImportable().containsValue(false))
+				{
+					// The package name that is selected in the "package selection" page
+					String defaultPackage = request.getParameter("defaultEntity");
+
+					LinkedHashMap<String, Boolean> entitiesImportable = importService.integrationTestMetaData(
```
fdlk: Should we not ask the MetaDataService directly?

PR #3752, opened 2015-09-24
[1.10] Fix #3737 Infinite transaction loop

### LINKED ISSUE #3737: Infinite transaction loop
Molgenis is generating an infinite loop of transactions (and elasticsearch indices, I think?)
See the MolgenisTransactionLog repository or set the log level to trace.

fdlk:


fdlk:
Wrong target!

PR #3755, opened 2015-09-25
Fix FindBugs warning, very carefully close the resources in the VcfValidator

fdlk:


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/VcfValidator.java
```diff
@@ -105,9 +106,24 @@ public String validateVCF(File vcfFile) throws IOException
...
+	{
+		if (closeable != null)
+		{
+			try
+			{
+				closeable.close();
+			}
+			catch (Exception ignore)
+			{
+
```
jjettenn: Missing exception? I would like to know with of them caused the exception.

Issue #3754, opened 2015-09-24
Inputs are very slow in large forms


tommydeboer:
## To reproduce:
1. Upload *Chr6_vragen_emx_testimport10 (2).xlsx* (see IssueData on Google Drive)
2. Go to the bottom of the survey
3. Type 'molgenis' in the text field
4. Wait
5. Wait
6. Wait
7. Conclude that it's slow

## Extra's:
The problem is way worse in development mode, but still very noticeable in production.
Profiled (in development mode) with *React.addons.Perf* (while typing the word 'molgenis'):

![image](https://cloud.githubusercontent.com/assets/5362890/10075729/a5f03982-62d7-11e5-851b-87727272f926.png)
Wasted time = time spent on components that did not change their rendering in any way.

tommydeboer:
The performance improves when I remove all the visibility expressions. There are a lot of those in this survey and it might be that they get checked too often (at every key press?). However, even then it still is laggy enough to be bothersome.

fdlk:
Forms are quite a bit slower in Safari than in Chrome by the way.

Mark-de-Haan:
Wontfix: Will be adressed in the [VueForms](https://github.com/Mark-de-Haan/molgenis-vue-forms)

dennishendriksen:
> 
> 
> Wontfix: Will be adressed in the [VueForms](https://github.com/Mark-de-Haan/molgenis-vue-forms)

Was it addressed?

PR #3760, opened 2015-09-25
[1.10] Improve #3754 Inputs are very slow in large forms

fdlk:
This significantly improves the situation in #3754


tommydeboer:
I've done multiple benchmarks but I'm not seeing a difference:

### Without fix:
image
image

### With fix:
image
image

PR #3798, opened 2015-09-29
[1.10] Remove entity locking

erwinwinder:


fdlk:
If an annotator fails, I end up in an inconsistent state where the model registry knows the entity but the tag wizard cannot retrieve its attributes. Not sure if this is due to the fix or an existing bug.

PR #3840, opened 2015-10-05
M3642 As Rest API user I want to batch update/create entities

jjettenn:
1. Batch create entities
2. Batch update entities
3. Batch update entities (Only one attribute)


molgenis-data-rest/src/main/java/org/molgenis/data/rest/Href.java
```diff
@@ -119,4 +121,30 @@ public static String concatMetaEntityHref(String baseUri, String qualifiedEntity
...
+	 * @param entityIdValue
+	 * @return
+	 * @throws UnsupportedEncodingException
+	 */
+	public static String concatEntityCollectionHref(String baseUri, String qualifiedEntityName,
+			String qualifiedIdAttributeName, List<String> entitiesIds)
+	{
+		try
+		{
+			String ids = String.join(",", entitiesIds);
```
fdlk: shouldn't they be somehow encoded?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/ResourcesResponseV2.java
```diff
@@ -0,0 +1,18 @@
+package org.molgenis.data.rest.v2;
+
+import static java.util.Objects.requireNonNull;
+
+public class ResourcesResponseV2
```
fdlk: Can be an AutoValue value object, I think.

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -129,6 +183,215 @@ public EntityCollectionResponseV2 retrieveEntityCollectionPost(@PathVariable("en
...
+			}
+
+			response.addHeader("Location", Href.concatEntityCollectionHref(RestControllerV2.BASE_URI, entityName, meta
+					.getIdAttribute().getName(), ids));
+			response.setStatus(HttpServletResponse.SC_CREATED);
+			return responseBody;
+		}
+		catch (Exception e)
+		{
+			response.setStatus(HttpServletResponse.SC_NO_CONTENT);
```
fdlk: Smells of success

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -129,6 +183,215 @@ public EntityCollectionResponseV2 retrieveEntityCollectionPost(@PathVariable("en
...
+
+		try
+		{
+			final List<Entity> entities = request.getEntities().stream().map(e -> this.restService.toEntity(meta, e))
+					.collect(Collectors.toList());
+
+			int count = 0;
+			for (Entity entity : entities)
+			{
+				this.getEntityId(entity, count);
```
fdlk: Nothing happens with the ID?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -129,6 +183,215 @@ public EntityCollectionResponseV2 retrieveEntityCollectionPost(@PathVariable("en
...
+	}
+
+	/**
+	 * Get entity id and perform a check, throwing an MolgenisDataException when necessary
+	 * 
+	 * @param entity
+	 * @param count
+	 * @return
+	 */
+	private String getEntityId(Entity entity, int count)
```
fdlk: ah, it does! perhaps rename to checkForEntityId then or something

molgenis-data-rest/src/test/java/org/molgenis/data/rest/v2/RestControllerV2Test.java
```diff
@@ -344,6 +368,208 @@ public void retrieveResourceCollection() throws Exception
...
+
+	/**
+	 * Test all the exception that the method 'updateEntitiesExceptions' can throw
+	 * 
+	 * @throws Exception
+	 */
+	@Test
+	public void testUpdateEntitiesExceptions() throws Exception
+	{
+		this.testUpdateEntitiesExceptions("entity", "{}", RestControllerV2.EXCEPTION_NO_ENTITIES);
```
fdlk: I really like the utility method, but I think the tests are worth separate test methods to make sure all tests get run even if one fails.

molgenis-data-rest/src/main/java/org/molgenis/data/rest/service/RestService.java
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	/**
+	 * Creates a new MapEntity based from a HttpServletRequest
+	 * 
+	 * @param meta
+	 * @param request
+	 * @return
+	 */
+	public Entity toEntity(final EntityMetaData meta, final Map<String, Object> request)
```
fdlk: Would this also work as a spring data converter thingy? That you register it somewhere as a converter from Request to entity and that it then allows you to rewrite the REST method as one that accepts Entity parameter requests rather than HttpRequest

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -47,17 +58,60 @@
...
+	static UnknownAttributeException createUnknownAttributeException(String entityName, String attributeName)
+	{
+		return new UnknownAttributeException("Operation failed. Unknown attribute: '" + attributeName
+				+ "', of entity: '" + entityName + "'");
+	}
+
+	static MolgenisDataAccessException createMolgenisDataAccessExceptionReadOnlyAttribute(String entityName,
+			String attributeName)
+	{
+		return new MolgenisDataAccessException("Operation failed. Attribute '" + attributeName + "' of entity '"
```
fdlk: I think the dataservice can throw this exception for you?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/Href.java
jjettenn: They are encoded: UriUtils.encodePathSegment(ids.toString(), "UTF-8"));

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -47,17 +58,60 @@
 class RestControllerV2
 {
 	private static final Logger LOG = LoggerFactory.getLogger(RestControllerV2.class);
+	static final int MAX_ENTITIES = 1000;
```
fdlk: How long does the Location url get for the standard ID generator and 1000 entities? Do we expect that to fit in all browsers?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/service/RestService.java
fdlk: We found the answer: The converter would need to know the metadata of the entity to convert to so it cannot be done that easily.

jjettenn:
Use this code to generate test entities.


package org.molgenis.data.rest.v2;

import java.util.UUID;

public class GenerateTestDataTool
{

	public static void main(String[] args)
	{
		System.out.println(generatePersonNewEntities(6, RestControllerV2.MAX_ENTITIES));
	}

	public static String generatePersonNewEntities(int start, int amount)
	{
		StringBuilder sb = new StringBuilder();
		sb.append("{entities:[");
		for (int i = start; i < amount + start; i++)
		{
			if (i > start)
			{
				sb.append(',');
			}
			sb.append("{value:\"" + UUID.randomUUID() + "\", label:\"label" + i + "\"}");
		}
		sb.append("]}");
		return sb.toString();
	}
}

jjettenn:
package org.molgenis.data.rest.v2;

import java.util.UUID;

public class GenerateTestDataTool
{

	public static void main(String[] args)
	{
		System.out.println(generatePersonNewEntities(6, RestControllerV2.MAX_ENTITIES));
	}

	public static String generatePersonNewEntities(int start, int amount)
	{
		StringBuilder sb = new StringBuilder();
		sb.append("{entities:[");
		for (int i = start; i < amount + start; i++)
		{
			if (i > start)
			{
				sb.append(',');
			}
			sb.append("{value:\"" + UUID.randomUUID() + "\", label:\"label" + i + "\"}");
		}
		sb.append("]}");
		return sb.toString();
	}
}

jjettenn:
fixed \\\"p1\\\"

PR #3844, opened 2015-10-06
M3796 Retrieve entities by id from backend instead of index + improvements

dennishendriksen:
- Update Elasticsearch 1.4.4 to 1.7.2
- Do not store entities in Elasticsearch unless the backend is Elasticsearch
- Elasticsearch decorator retrieves entities by id from decorated repository instead of index
- Add optimizeIndex to SearchService, execute after each migration


dennishendriksen:
Cannot reproduce "All checks have failed" error reported by bot.

PR #3851, opened 2015-10-07
Fix #3839 The mapping category editor does not show up for categorical target attributes

### LINKED ISSUE #3839: The mapping category editor does not show up for categorical target attributes
How to reproduce:
- go to molgenis01
- create a new mapping project with HOP as the target and add the finrisk biobank as the source
- in the mapping overview screen, click on the cell that is located in the row "History of Stroke" and column 'finrisk'
- the mapping category editor is not in the mapping screen
- unselect the top suggested attribute "Cerebrovascular accident" and select it again, the mapping category editor shows up

Mark-de-Haan:


molgenis-data-mapper/src/main/resources/js/attribute-mapping.js
```diff
@@ -739,7 +739,7 @@
 		});
 
 		$('a[href=#script]').on('shown.bs.tab', function() {
-		// when users switch back to the script tab, the value in the editor is not updated
+		// when users switches back to the script tab, the value in the editor is not updated
```
fdlk: regression!! ;-)

PR #3856, opened 2015-10-07
Fix #3576 No error message when session is expired

### LINKED ISSUE #3576: No error message when session expired
Reproduce:
- Login and wait until session expires (workaround: restart server after login)
- Select a plugin in the menu

Expected:
- Login dialog displayed with "session expired" message

Actual:
- Login dialog displayed with no message

erwinwinder:


molgenis-security/src/main/java/org/molgenis/security/MolgenisWebAppSecurityConfig.java
```diff
@@ -139,7 +142,7 @@ protected void configure(HttpSecurity http) throws Exception
 
 		.formLogin().loginPage("/login").failureUrl("/login?error").and()
 
-		.logout().addLogoutHandler((req, res, auth) -> {
+		.logout().deleteCookies("JSESSIONID").addLogoutHandler((req, res, auth) -> {
```
fdlk: Somewhat magical string. I think the name of the session cookies can be fished up out of the ServletContext but I don't see that available here.

erwinwinder: All containers must use JSESSIONID.
Needed because when you sign out your sessionid is invalid, so when you go to the molgenis login form it will shoe the message 'Your login session is expired', that is a bit odd.

PR #3860, opened 2015-10-08
Fix #3842 Entity class permissions not deleted when an entity metadata is deleted

### LINKED ISSUE #3842: Entity class permissions not deleted when an entity metadata is deleted
- Click delete entitymetadata + data in the dataexplorer
- Goto Admin > Permisison Manager > Entity Class Permissions

Expected: The Permissions for the deleted entity are gone.
Observed: Permissions stil exist.

If you then upload an entity with the same name it automatically gets the permissions set for the deleted entity

erwinwinder:


molgenis-data/src/main/java/org/molgenis/data/meta/MetaDataServiceImpl.java
```diff
@@ -138,17 +140,38 @@ public void deleteEntityMeta(String entityName)
...
 	}
 
+	private void deleteEntityPermissions(String entityName)
+	{
+		List<String> authorities = SecurityUtils.getEntityAuthorities(entityName);
+
+		// User permissions
+		if (dataService.hasRepository("UserAuthority"))
+		{
+			Iterable<Entity> userPermissions = dataService.query("UserAuthority").in("role", authorities).findAll();
```
Mark-de-Haan: do we want to use UserAuthority.ENTITYNAME and UserAuthority.ROLE?

molgenis-data/src/main/java/org/molgenis/data/meta/MetaDataServiceImpl.java
```diff
@@ -138,17 +140,38 @@ public void deleteEntityMeta(String entityName)
...
+			dataService.delete("UserAuthority", userPermissions);
+		}
+
+		// Group permissions
+		if (dataService.hasRepository("GroupAuthority"))
+		{
+			Iterable<Entity> groupPermissions = dataService.query("GroupAuthority").in("role", authorities).findAll();
```
Mark-de-Haan: do we want to use GroupAuthority.ENTITYNAME and GroupAuthority.ROLE?

molgenis-data/src/main/java/org/molgenis/data/meta/MetaDataServiceImpl.java
erwinwinder: In principle yes.
But here we get dependency problems.
That is also the reason that the method is not added to the PermissionManagerService

molgenis-data/src/main/java/org/molgenis/data/meta/MetaDataServiceImpl.java
erwinwinder: In principle yes.
But here we get dependency problems.
That is also the reason that the method is not added to the PermissionManagerService

Mark-de-Haan:
Merge with master please

PR #3879, opened 2015-10-09
Fix #3878 MySqlRepository returns entities out of order

### LINKED ISSUE #3878: MySqlRepository returns entities out of order
```Iterable<Entity> MySqlRepository.findAll(Iterable<Object> ids)``` returns entities out of order. This is unexpected, the desired behavior is that the returned Iterable has the same order as the Iterable of ids.

dennishendriksen:


fdlk:
We run into an issue where the group_concat for mref attributes gets truncated due to the additional ORDER BY FIELD clause.

PR #3880, opened 2015-10-11
Common HttpClient, improve performance, add meta data

dennishendriksen:


molgenis-app/src/main/java/org/molgenis/rdconnect/IdCardOrganization.java
```diff
@@ -0,0 +1,49 @@
+package org.molgenis.rdconnect;
+
```
jjettenn: Can it be an auto_value object?

fdlk: Value objects don't have setters so that'd depend on if the setters are being used.

dennishendriksen: Setters are in use, so no value object.

PR #3938, opened 2015-10-15
[1.11] Fix #3919 Manually Deleting mapping project related records will break ElasticSearch index

### LINKED ISSUE #3919: Manually Deleting mapping project related records will break ElasticSearch index
## Reproduce
- Create a mapping project using the AttributeMapping entity as the target and EntityMapping entity as the source
- Go to DataExplorer
- Delete all the data records (without metadata) in MappingProject
- Delete all the data records (without metadata) in MappingTarget
- Delete all the data records (without metadata) in EntityMapping
- Delete all the data records (without metadata) in AttributeMapping
- Create a mapping project using the Disease entity as the target and DiseaseMapping entity as the source

## Expected
Succeed in creating a mapping project

## Observed
Index being broken, the error message is followed

```
{
  "errors": [
    {
      "message": "An error occurred. Please contact the administrator.<br />Message:Failed to execute phase [query_fetch], all shards failed; shardFailures {[Sq2Df-rrRce4-tH5SUENcw][molgenis][0]: SearchParseException[[molgenis][0]: from[0],size[1000]: Parse Failure [Failed to parse source [{\"from\":0,\"size\":1000,\"query\":{\"filtered\":{\"query\":{\"match_all\":{}},\"filter\":{\"nested\":{\"filter\":{\"term\":{\"mappingtargets.identifier.raw\":\"AAAACUDFWGJSBXVKUEIVWEAAAI\"}},\"path\":\"mappingtargets\"}}}}}]]]; nested: QueryParsingException[[molgenis] [nested] failed to find nested object under path [mappingtargets]]; }"
    }
  ]
}
```

Mark-de-Haan:


molgenis-data-elasticsearch/src/main/java/org/molgenis/data/elasticsearch/AbstractElasticsearchRepository.java
```diff
@@ -178,6 +178,7 @@ public void deleteById(Iterable<Object> ids)
 	public void deleteAll()
 	{
 		elasticSearchService.delete(getEntityMetaData().getName());
+		create();
```
fdlk: TODO: Check if delete entity + metadata also passes through here.

Mark-de-Haan: Checked, delete entity + metadata does not use this function

PR #3945, opened 2015-10-15
[1.11] Fix #3849 Molgenis validation exception count a row too much.

### LINKED ISSUE #3849: Molgenis validation exception count a row too much.
To reproduce:
Try to add a row via the RestAPIV2
1. Upload the emx_all_datatypes.xlsx https://drive.google.com/open?id=0B902kZCFbOw1ZkRITGlJaUVDZGM
2. Try to create a new entity via the "Advanced Rest Client" tool (Chrome):  org_molgenis_test_TypeTestRef
a. URL: http://localhost:8080/api/v2/org_molgenis_test_TypeTestRef
b. Payload: {entities:[{label:"label333"}]}

Error message: 
message: "The attribute 'value' of entity 'org_molgenis_test_TypeTestRef' can not be null. (row 2)"

Expected: 
message: "The attribute 'value' of entity 'org_molgenis_test_TypeTestRef' can not be null. (row 1)"

Tip:
In ConstraintViolation.java: the violation.getMessage(); increases to row number "return message + " (row " + (rownr + 1) + ")";"
and in the  MolgenisValidationException generator gives it already the number 1.

Important: Look on the side effects for the system when solving this issue!

jjettenn:


molgenis-data-jpa/src/test/java/org/molgenis/data/jpa/JpaImportTest.java
```diff
@@ -158,7 +158,7 @@ public void testMrefWrong() throws IOException
 		{
 			assertEquals(e.getViolations().size(), 1);
 			assertEquals(e.getViolations().iterator().next().getViolatedAttribute().getName(), "children");
-			assertEquals(e.getViolations().iterator().next().getRownr(), 3);
+			assertEquals(e.getViolations().iterator().next().getRownr(), Long.valueOf(3));
```
fdlk: But the JPA unit tests sill check for the same row numbers. Were their row numbers incorrect as well?

PR #3948, opened 2015-10-15
[master] Fix #3849 Molgenis validation exception count a row too much.

### LINKED ISSUE #3849: Molgenis validation exception count a row too much.

jjettenn:


molgenis-data-jpa/src/test/java/org/molgenis/data/jpa/JpaImportTest.java
```diff
@@ -76,7 +76,7 @@ public void testMissingRequiredValue() throws IOException
 		{
 			assertEquals(e.getViolations().size(), 1);
 			assertEquals(e.getViolations().iterator().next().getViolatedAttribute().getName(), "lastName");
-			assertEquals(e.getViolations().iterator().next().getRownr(), 2);
+			assertEquals(e.getViolations().iterator().next().getRownr(), Long.valueOf(2));
```
fdlk: I think you can also write 2L

jjettenn: There is a difference between the notation:
2L Results to primitive while Long.valueOf() results to a Long object.

PR #3968, opened 2015-10-16
[1.11] Fix #3878 MySqlRepository returns entities out of order; Fix #3887 When clicking the dropdown in the settingsscreen I get a error

### LINKED ISSUE #3878: MySqlRepository returns entities out of order

### LINKED ISSUE #3887: When clicking the dropdown in the settingsscreen I get a error
screen shot 2015-10-13 at 14 54 35

fdlk:
Alternative solution that does not run into the group_concat / order by bug.

### DIFF
--- a/molgenis-data-mysql/src/main/java/org/molgenis/data/mysql/MysqlRepository.java
+++ b/molgenis-data-mysql/src/main/java/org/molgenis/data/mysql/MysqlRepository.java
@@ -1,5 +1,7 @@
 package org.molgenis.data.mysql;
 
+import static com.google.common.collect.Lists.newArrayList;
+import static org.elasticsearch.common.collect.Maps.uniqueIndex;
 import static org.molgenis.data.RepositoryCapability.QUERYABLE;
 import static org.molgenis.data.RepositoryCapability.UPDATEABLE;
 import static org.molgenis.data.RepositoryCapability.WRITABLE;
@@ -676,7 +678,19 @@ public class MysqlRepository extends AbstractRepository implements Manageable
 	public Iterable<Entity> findAll(Iterable<Object> ids)
 	{
 		if (ids == null) return Collections.emptyList();
-		return findAll(new QueryImpl().in(getEntityMetaData().getIdAttribute().getName(), ids));
+		Query inQuery = new QueryImpl().in(getEntityMetaData().getIdAttribute().getName(), ids);
+		Map<Object, Entity> index = uniqueIndex(newArrayList(findAll(inQuery)), Entity::getIdValue);
+		return Iterables.filter(Iterables.transform(ids, id -> lookup(index, id)), x -> x != null);
+	}
+
+	public Entity lookup(Map<Object, Entity> index, Object id)
+	{
+		Entity result = index.get(id);
+		if (result == null)
+		{
+			LOG.warn("Couldn't find {} for id {}.", getName(), id);
+		}
+		return result;
 	}
 
 	protected String getSelectSql(Query q, List<Object> parameters)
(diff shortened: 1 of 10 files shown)


molgenis-data-mysql/src/main/java/org/molgenis/data/mysql/MysqlRepository.java
```diff
@@ -678,7 +678,19 @@ public Entity findOne(Object id)
 	public Iterable<Entity> findAll(Iterable<Object> ids)
 	{
 		if (ids == null) return Collections.emptyList();
-		return findAll(new QueryImpl().in(getEntityMetaData().getIdAttribute().getName(), ids));
+		Query inQuery = new QueryImpl().in(getEntityMetaData().getIdAttribute().getName(), ids);
+		Map<Object, Entity> index = uniqueIndex(newArrayList(findAll(inQuery)), Entity::getIdValue);
```
fdlk: This means the complete set of results gets cached in memory. On the other hand I'd expect that a query for specific ids is limited to manageable batches of rows. Could add batching here to make sure. What do you guys think?

molgenis-data-mysql/src/main/java/org/molgenis/data/mysql/MysqlRepository.java
```diff
@@ -678,7 +678,19 @@ public Entity findOne(Object id)
...
+		return Iterables.filter(Iterables.transform(ids, id -> lookup(index, id)), x -> x != null);
+	}
+
+	public Entity lookup(Map<Object, Entity> index, Object id)
+	{
+		Entity result = index.get(id);
+		if (result == null)
+		{
+			LOG.warn("Couldn't find {} for id {}.", getName(), id);
```
fdlk: I ran into an issue here where the type of the ID that gets sent to the query method is different from the java type returned by the getIdValue of the entity. I think that it was a bug that ElasticSearch sent String representations of int IDs to the findAll method. But perhaps it happens elsewhere as well.

molgenis-data-mysql/src/main/java/org/molgenis/data/mysql/MysqlRepository.java
```diff
@@ -956,32 +968,7 @@ protected String getSortSql(Query q, List<Object> parameters)
...
-					// define order of returned rows
-					sortSql = new StringBuilder("ORDER BY FIELD(").append("this.").append('`')
-							.append(queryRule.getField()).append('`').append(",").append(fieldIds).append(")");
-
-					// use prepared statements for the values
-					parameters.addAll(StreamSupport.stream(values.spliterator(), false).map(value -> value.toString())
-							.collect(Collectors.toList()));
-				}
-			}
-		}
```
fdlk: This is the old solution, it specifies the order of the result rows in an ORDER BY FIELD clause. In MySQL 5.1 it runs into a bug where the collected MREF IDs get truncated if you add an ORDER BY.

molgenis-data/src/main/java/org/molgenis/data/Repository.java
```diff
@@ -45,9 +45,9 @@
 	Entity findOne(Object id);
 
 	/**
-	 * find entities based on a stream of ids.
+	 * find entities based on a stream of ids
 	 * 
-	 * @return (empty) Iterable where the order of entities matches the order of ids, never null
+	 * @return (empty) Iterable, never null
```
fdlk: How so never null? What's supposed to happen if one of the rows isn't found in the repository?

molgenis-data/src/main/java/org/molgenis/data/support/AbstractRepository.java
```diff
@@ -1,18 +1,35 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.transaction.annotation.Transactional;
 
 /**
  * Base class for repositories. Subclasses can override supported methods
  */
 public abstract class AbstractRepository implements Repository
 {
+	private static final int FIND_ALL_BATCH_SIZE = 1000;
```
fdlk: Picked the same batch size as the elastic search iterator.

fdlk:
Must still go to master

PR #4046, opened 2015-10-27
[promise] IBD Mapping + merge master

tommydeboer:


molgenis-app/src/main/java/org/molgenis/app/promise/mapper/ParelMapper.java
```diff
@@ -0,0 +1,309 @@
...
+	private PromiseDataParser promiseDataParser;
+	private DataService dataService;
+
+	private static final Logger LOG = LoggerFactory.getLogger(ParelMapper.class);
+
+	private static final HashMap<String, List<String>> materialTypesMap;
+
+	static
+	{
+		materialTypesMap = new HashMap<>();
```
fdlk: okay for now but longer term this is way better off somewhere in an emx file, imported and maintained outside this mapper

molgenis-app/src/main/java/org/molgenis/app/promise/model/PromiseMappingProjectMetaData.java
```diff
@@ -0,0 +1,232 @@
...
+				.setDescription("The name of this mapping");
+		addAttribute(BIOBANK_ID).setNillable(false).setLabel(LBL_BIOBANK_ID)
+				.setDescription("The ID of the biobank in the BBMRI-NL Sample Collections entity").setUnique(true);
+		addAttribute(CREDENTIALS).setDataType(MolgenisFieldTypes.XREF).setRefEntity(PromiseCredentialsMetaData.INSTANCE)
+				.setNillable(false).setLabel(LBL_CREDENTIALS)
+				.setDescription("The ProMISe credentials for this biobank");
+		addAttribute(MAPPER).setNillable(false).setLabel(LBL_MAPPER)
+				.setDescription("The mapper to use for this biobank");
+
+		setDecorator(new RepositoryDecorator()
```
fdlk: We have so many decorators, we really need an DefaultRepositoryDecorator that simply delegates everything. Then other decorators can extend this and only change the behavior that needs changing.

molgenis-data/src/main/java/org/molgenis/data/meta/MetaDataServiceImpl.java
```diff
@@ -257,7 +260,24 @@ public Repository add(EntityMetaData emd, RepositoryDecoratorFactory decoratorFa
 	@Override
 	public synchronized Repository addEntityMeta(EntityMetaData emd)
 	{
-		return add(emd, new NonDecoratingRepositoryDecoratorFactory());
+		if (emd.getDecorator() == null)
```
fdlk: This is not correct yet, right?

PR #4056, opened 2015-10-27
[1.12] Fix #4013 No COUNT permission on entity during import

### LINKED ISSUE #4013: No COUNT permission on entity during import
Reproduce:
- Login as non-superuser with permission on importwizard
- Import [prevend-emx-workshop-20-rows-with-package.xlsx](https://github.com/molgenis/molgenis/files/20901/prevend-emx-workshop-20-rows-with-package.xlsx) with default settings

Expected:
- Import successful

Actual:
```No COUNT permission on entity DRINK_1_Ref```

erwinwinder:


bartcharbon:
Trying to reproduce the issue now results in a login dialog instead of an error message

PR #4081, opened 2015-10-28
Fix #4079 Searching for single letters in the attribute screen will throw query shard failures for certain letters

### LINKED ISSUE #4079: Searching for single letters in the attribute screen will throw query shard failures for certain letters
## Reproduce
- Import the uo.owl.zip (unit ontology)
- Create a mapping project
- Add a target
- Edit a single attribute mapping
- in the attribute search, fill in 'e' or 'k'

## Expected
It will return attributes containing those letters

## Observed
Query shard failure
screen shot 2015-10-28 at 16 43 27

ChaoPang:


molgenis-data-semanticsearch/src/main/java/org/molgenis/data/semanticsearch/service/impl/SemanticSearchServiceHelper.java
```diff
@@ -267,27 +274,40 @@ private void recursivelyCollectAttributeIdentifiers(Iterable<Entity> attributeEn
...
-		String regex = "[^\\p{L}'a-zA-Z0-9\\.~^]+";
-		Set<String> searchTerms = stream(description.split(regex)).map(String::toLowerCase)
-				.filter(w -> !STOPWORDSLIST.contains(w) && StringUtils.isNotEmpty(w)).collect(Collectors.toSet());
+		Set<String> searchTerms = stream(description.split(ILLEGAL_CHARS_REGEX)).map(String::toLowerCase)
+				.filter(w -> !NGramDistanceAlgorithm.STOPWORDSLIST.contains(w) && StringUtils.isNotEmpty(w))
+				.collect(Collectors.toSet());
 		return searchTerms;
 	}
+
+	Double getBestInverseDocumentFrequency(List<String> terms)
```
Mark-de-Haan: private ? public?

molgenis-data-semanticsearch/src/test/java/org/molgenis/data/semanticsearch/service/impl/SemanticSearchServiceHelperTest.java
```diff
@@ -216,9 +211,10 @@ public void testFindTagsSync()
 	public void testSearchCircumflex() throws InterruptedException, ExecutionException
 	{
 		String description = "body^0.5 length^0.5";
-		Set<String> expected = Sets.newHashSet("length^0.5", "body^0.5");
+		Set<String> expected = Sets.newHashSet("length", "body", "0.5");
```
fdlk: Is this change desirable behavior?

ChaoPang: Yes it is. 
We do use the caret character internally for boosting specific words in the query after cleaning up the strings by removing stop words and illegal characters. However we don`t allow strings to contain the caret character, that means it`d be impossible to boost a  query outside the semantic search class.

Issue #4082, opened 2015-10-28
No spinner when using mapping-service


jjettenn:
To reproduce:
- Upload the emx_all_datatypes.xlsx file
- Create project with target TypeTest
- Add source TypeTest 
- Click multiple time on the "add source" button.

Result:
The mapping service will hang

Error message:
{
  "errors": [
    {
      "message": "An error occurred. Please contact the administrator.<br />Message:PreparedStatementCallback; SQL [UPDATE `AttributeMapping` SET `identifier` = ?, `targetAttributeMetaData` = ?, `sourceAttributeMetaDatas` = ?, `algorithm` = ?, `algorithmState` = ? WHERE `identifier`= ?]; Lock wait timeout exceeded; try restarting transaction; nested exception is java.sql.BatchUpdateException: Lock wait timeout exceeded; try restarting transaction"
    }
  ]
}

Expected to see a spinner. That I will be not able to click again on the button if it is executed.

jjettenn:
Fixed in #3230

jjettenn:
PR is not accepted

mswertz:
what is status of PR?

dennishendriksen:
Support for MySQL was dropped in MOLGENIS 2.0 in favor of PostgreSQL, hence this exception won't occur anymore.

PR #4083, opened 2015-10-28
[Master] Fix #3230: Empty page with error message in JSON format

### LINKED ISSUE #3230: Empty page with error message in JSON format
Difficult to reproduce (Ask JJ).
You can try: Adding a source entity, Creating a new project.
Clicking around in the mapping service plugin pages can result in the next error:

Example 1
{
  "errors": [
    {
      "message": "An error occurred. Please contact the administrator.<br />Message:Mapping already present for source MappingTarget"
    }
  ]
}

Example 2
screen shot 2015-06-25 at 3 04 43 pm

jjettenn:
Fix #3230 Empty page with error message in JSON format
Related to: No spinner when using mapping-service #4082

Both are fixed

The spinner was missing, causing all the problems

### DIFF
--- a/molgenis-data-mapper/src/main/resources/js/single-mapping-project.js
+++ b/molgenis-data-mapper/src/main/resources/js/single-mapping-project.js
@@ -52,17 +52,20 @@
 	        var currentForm = this;
 	        bootbox.confirm("Are you sure?", function(result) {
 	            if (result) {
+	            	showSpinner();
 	                currentForm.submit();
 	            }
 	        });
 	        return false;
 	    });
 		
-		$('#submit-new-source-column-btn').click(function() {
-			$('#create-new-source-form').submit();
+		$('#submit-new-source-column-btn').on('click', function(e) {
+			showSpinner();
+	        $('#create-new-source-form').submit();
 		});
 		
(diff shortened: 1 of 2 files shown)


molgenis-data-mapper/src/main/resources/js/single-mapping-project.js
```diff
@@ -52,17 +52,20 @@
...
 	            }
 	        });
 	        return false;
 	    });
 		
-		$('#submit-new-source-column-btn').click(function() {
-			$('#create-new-source-form').submit();
+		$('#submit-new-source-column-btn').on('click', function(e) {
+			showSpinner();
+	        $('#create-new-source-form').submit();
```
fdlk: On Safari, I can still click the submit button twice before submission if I'm fast enough. I still get the json error message.

But now if I then want to get back to the menu so I push the back button on the empty json page, I see the previous page with a spinner that I cannot click away.

fdlk:
I personally think that this sort of bug (if it is a bug at all, users with some experience know not to spam form submission buttons in HTML forms) is best solved by switching to React.

jjettenn:
We can beter reactify the Mapping-Service. The spinner is not working correctly with HTML forms

PR #4106, opened 2015-10-30
M3778 As user I want table edit controls based on repo capabilities

erwinwinder:


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
```diff
@@ -135,7 +136,7 @@ public MappingServiceController()
 	public String viewMappingProjects(Model model)
 	{
 		model.addAttribute("mappingProjects", mappingService.getAllMappingProjects());
-		model.addAttribute("entityMetaDatas", getEntityMetaDatas());
+		model.addAttribute("entityMetaDatas", getWritableEntityMetaDatas());
```
fdlk: Why only the writeable ones?

PR #4108, opened 2015-10-30
Fix #4107 MySqlRepository.findAll unnecessary queries

### LINKED ISSUE #4107: MySqlRepository.findAll performs unnecessary queries
Reproduce:
- Enable DEBUG logging on org.molgenis in 'Log manager' plugin in the 'Admin' menu
- Request the following resource ```GET http://localhost:8080/api/v1/settings_app```

The logs show that a MySQL query was performed with ```pageSize=1000, offset=0``` and immediately afterwards a query was performed with ```pageSize=1000, offset=1000```. The last query is not necessary since the number of entities is smaller then the batch size.

dennishendriksen:


dennishendriksen:
Related to #3959

tommydeboer:
Works fine. However the logging only logs the first two queries if you go above 2000 items. Try with
[test_12000rows.xlsx](https://github.com/molgenis/molgenis/files/25971/test_12000rows.xlsx):

```
http://localhost:8080/api/v1/test_manyRows?_method=GET
{'num': 10000}
```

PR #4110, opened 2015-10-30
[google friday] getVersion for REST V2 controller

bartcharbon:
I was trying to create a overview of which serverversion is used where.
This function might help in the future.

PR adds one function, github has some trouble understanding because of the formatting (class was not formatted before with the "eclipse mars" version of the formatter?)


fdlk:
Sorry, I think I just accepted a pull request that needs to be merged with this one.

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -113,6 +116,24 @@ public RestControllerV2(DataService dataService, MolgenisPermissionService permi
 		this.restService = requireNonNull(restService);
 	}
 
+	@Autowired
```
fdlk: I need authorization to access, I don't think the version endpoint should be protected

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -113,6 +116,24 @@ public RestControllerV2(DataService dataService, MolgenisPermissionService permi
...
+	public Map<String, String> getVersion(@Value("${molgenis.version:@null}") String molgenisVersion,
+			@Value("${molgenis.build.date:@null}") String molgenisBuildDate)
+	{
+		if (molgenisVersion == null) throw new IllegalArgumentException("molgenisVersion is null");
+		if (molgenisBuildDate == null) throw new IllegalArgumentException("molgenisBuildDate is null");
+		molgenisBuildDate = molgenisBuildDate.equals("${maven.build.timestamp}")
+				? new SimpleDateFormat("yyyy-MM-dd HH:mm").format(new java.util.Date()) + " by Eclipse"
+				: molgenisBuildDate;
+
+		Map result = new HashMap();
```
fdlk: I get untyped generic warnings here

PR #4116, opened 2015-11-02
[google friday] Create link from genomebrowser popup to entityreport.

bartcharbon:


molgenis-das/src/main/java/org/molgenis/das/impl/RepositoryRangeHandlingDataSource.java
```diff
@@ -181,7 +172,7 @@ else if (patients.containsKey(valuePatient))
 					++score;
 				}
 
-				feature = createDasFeature(valueStart, valueStop, valueIdentifier, valueName, valueDescription,
+				feature = createDasFeature(valueStart, valueStop, entity.getIdValue().toString(), entity.getLabelValue(), valueDescription,
```
fdlk: So the idValue is no longer null, nor empty. Correct?

bartcharbon: we used to use "GenomicDataSettings.Meta.ATTRS_IDENTIFIER" as the id, instead of taking the entity's idValue. In that situation it could be null, and that is indeed not the case for the idAttribute of an entity.

fdlk: Understood!

fdlk:
I tried with a VCF, then if I click "Filter on mutation" it adds a filter for the ID column with the value of the VCF row's generated ID. Unfortunately the filter then doesn't work since the ID is invisible.
No rows shown.

The external ID has the label ID which makes things even more confusing for the user.

molgenis-dataexplorer/src/main/resources/js/dataexplorer-data.js
```diff
@@ -276,14 +276,22 @@
                         createFilter(attr, undefined, undefined, f.id);
                     });
                     if (f.id !== "-" && attr !== undefined) {
-                        info.setTitle(f.id);
-                        info.add('Filter on mutation:', a[0]);
-                        //cache the information
+						if(attr.visible) {
```
fdlk: Does mean that for VCF's the filter no longer works. But looks like this is intended.

PR #4121, opened 2015-11-04
[master] Fix #4119 (addendum)

### LINKED ISSUE #4119: UnknownEntityException when using getEntity() on an mref of a DefaultEntity
When using ```getEntity()``` on an *mref* or *categorical mref* of a ```DefaultEntity``` you get an ```UnknownEntityException``` with the ID of the entity it couldn't find. This is confusing, because the entity does exist.

tommydeboer:
- added unit tests
- also added type check to getEntities()


molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntity.java
```diff
@@ -248,11 +247,9 @@ public Entity getEntity(String attributeName)
 			return new DefaultEntity(attribute.getRefEntity(), dataService, (Map<String, Object>) value);
 
 		FieldType dataType = attribute.getDataType();
-		if (attribute.getDataType().equals(MolgenisFieldTypes.MREF)
-				|| attribute.getDataType().equals(MolgenisFieldTypes.CATEGORICAL_MREF))
+		if (!(dataType instanceof XrefField))
 		{
-			throw new MolgenisDataException(
-					"can't use getEntity() on an mref/categorical_mref, use getEntities() instead");
+			throw new MolgenisDataException("can't use getEntity() on something that's not an xref or categorical");
```
dennishendriksen: Or File

molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntity.java
```diff
@@ -275,6 +272,16 @@ public Entity getEntity(String attributeName)
 	@Override
 	public Iterable<Entity> getEntities(String attributeName)
 	{
+		AttributeMetaData attribute = entityMetaData.getAttribute(attributeName);
+		if (attribute == null) throw new UnknownAttributeException(attributeName);
+
+		FieldType dataType = attribute.getDataType();
+		if (!(dataType instanceof MrefField) && !(dataType instanceof XrefField))
```
dennishendriksen: getEntities cannot be called when dataType instanceof XrefField

tommydeboer: Yes it can and wil return an iterable with size 1

dennishendriksen: This will fail for other implementations of Entity. Do you know a use case where DefaultEntity contains a Iterable<Entity> for a CATEGORICAL_XREF/XREF/FILE attribute?

dennishendriksen:
Build failure

dennishendriksen:
Possible use case /molgenis-data-import/src/test/java/org/molgenis/data/importer/EmxImportServiceTest.java
The xref attribute is requested using getEntities

tommydeboer:
Possible use case or improper use of ```getEntities()``` where ```getEntity()``` should've been used? As you said, returning an ```Iterable``` of ```Entity``` for an xref is the wrong behaviour and this will not work for other implementations of ```Entity```.

dennishendriksen:
"Possible use case or improper use of getEntities() where getEntity() should've been used?"
I don't know, we need to be sure about this, since any problems will only occur at runtime.

tommydeboer:
Agreed, but that would mean an extensive look at the code base. How about we allow ```getEntities()``` on xref for now so we can merge this PR? When we are sure about the situation we can add back the more stringent check.

dennishendriksen:
Sounds like a plan. Could you add a note for this in the source?

molgenis-data/src/test/java/org/molgenis/data/support/DefaultEntityTest.java
```diff
@@ -21,14 +30,72 @@
 	DataService dataService;
 
 	@BeforeMethod
-	public void beforeMethod()
+	public void beforeClass()
```
dennishendriksen: Use BeforeClass annotation and make method static if you want to run this code only once, else rename back to beforeMethod.

dennishendriksen:
Please merge with master.

tommydeboer:
Rebased to 1.13 . Please merge the new PR (#4176)

PR #4125, opened 2015-11-06
fix MySQLRepositoryCount unit test

bartcharbon:


molgenis-data-mysql/src/test/java/org/molgenis/data/mysql/MysqlRepositoryCountTest.java
```diff
@@ -27,8 +30,10 @@
 	DataService dataService;
 
 	@Test
-	public void test()
+	public void test() throws ParseException
 	{
+		SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
```
dennishendriksen: = MolgenisDateFormat.getDateFormat();

bartcharbon: that will destroy the test again :( might be a bug in the embedden mysql?
mysql-connector-mxj 5.0.12 seems to be the up-to-date version...

PR #4144, opened 2015-11-12
ID-Card repository and indexer plugin

dennishendriksen:
- ID-Card repository that retrieves biobanks/registries from ID-Card
- ID-Card indexer that indexes (not stores) biobanks/registries from ID-Card
- REST API aggregate endpoint with RSQL query syntax
- Aggregate table React component
- Updated data service aggregates method to return axis ids instead of labels
- Common HTTP client config


molgenis-data-annotators/pom.xml
```diff
@@ -43,11 +43,6 @@
 			<artifactId>spring-webmvc</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>org.apache.httpcomponents</groupId>
-			<artifactId>httpclient</artifactId>
-			<version>4.2.1</version>
-		</dependency>
-		<dependency>
```
bartcharbon: verified that the annotators still work, probably an artifact from the late chembl annotator

PR #4148, opened 2015-11-13
[1.12] BBMRI mapper scheduler fixes

tommydeboer:


molgenis-app/src/main/java/org/molgenis/bbmri/controller/BbmriNlToEricMapperController.java
```diff
@@ -133,10 +138,17 @@ public void onApplicationEvent(ContextRefreshedEvent event)
 	{
 		try
 		{
-			doScheduleMappingJob();
+			doScheduleMappingJob(createCronTrigger(), "auto");
 		}
 		catch (Throwable t)
 		{
 		}
```
erwinwinder: Do not silently catch, at least log the error

molgenis-data-rest/src/test/java/org/molgenis/data/rest/v2/RestControllerV2Test.java
```diff
@@ -396,6 +396,12 @@ public void retrieveResourceCollection() throws Exception
 	}
 
 	@Test
+	public void retrieveResourceCollectionAggregation() throws Exception
+	{
+
+	}
+
+	@Test
```
erwinwinder: Implement or remove

PR #4152, opened 2015-11-16
Merge with 1.13, using strategy "ours"

fdlk:


fdlk:
See http://stackoverflow.com/questions/29301818/how-should-i-update-the-version-inside-my-pom-xml-when-releasing-using-git-flow

PR #4179, opened 2015-11-19
M3901 molgenis.py

fdlk:
Used in the Python API demo on the molgenis.org site.


molgenis-python/src/main/resources/templates/molgenis.python.ftl
```diff
@@ -1,282 +1,158 @@
...
+
+        You can have multiple file type attributes.
+
+        >>> session.add('Plot', files={'image': ('expression.jpg', open('~/first-plot.jpg','rb')), 
+			'image2': ('expression-large.jpg', open('/Users/me/second-plot.jpg', 'rb'))}, 
+			data={'name':'IBD-plot'})
+		''' 
+		response = self.session.post(self.url + "v1/" + quote_plus(entity),
+			headers = self._get_token_header(),
+			data = self._merge_two_dicts(data, kwargs),
```
npklein: Use copy.copy(data,kwargs)

npklein:
@fdlk Usual spaces-per-tab in Pyhon is 4

fdlk:
@npklein I think the spaces-per-tab is simply wrong in github, the source file contains tabs rather than spaces. Or do you mean it would be better to convert the tabs to spaces in the source file?

npklein:
Ah I only saw the github file where it looked like 8. PEP8 does say spaces preferred over tabs (http://legacy.python.org/dev/peps/pep-0008/#tabs-or-spaces)

fdlk:
Thanks, I'll switch to spaces!

molgenis-python/src/main/resources/templates/molgenis.python.ftl
fdlk: I just updated but it doesn't work for me:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "molgenis.py", line 108, in add
    data = copy.copy(data, kwargs),
TypeError: copy() takes exactly 1 argument (2 given)
```

npklein: Woops I misread what you were doing, it should stay as it is, sorry

PR #4182, opened 2015-11-23
Fix #4122 REST API authentication with invalid JSESSIONID and valid x…

### LINKED ISSUE #4122: REST API authentication with invalid JSESSIONID and valid x-molgenis-token gives 401
If you send an invalid JSESSIONID together with a valid x-molgenis-token, you'll get a 401.
I'd expect the valid token to win.

This happened in the RestClient interactions with molgenis initiated by the selenium tests.
Fixed it there by disabling cookie management

erwinwinder:
…-molgenis-token gives 401


fdlk:
Mhm.. Looks like it now gives the error message in html rather than json

PR #4206, opened 2015-11-25
Add update() function,

npklein:
- Extra info in error message when raise_for_status is called
- Fix error where .raise_for_status was called on a dictionary (when server response does not give an error)


molgenis-python/src/main/resources/templates/molgenis.py.ftl
```diff
@@ -158,3 +175,18 @@ class Session():
...
+        try:
+            if 'errors' in response.json() and 'message' in response.json()['errors'][0]:
+                error = (response.json()['errors'][0]['message'])
+        except json.decoder.JSONDecodeError:
+            pass
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            if error:
+                raise requests.exceptions.HTTPError(str(e)+'\n'+error)
```
fdlk: But won't that mean that you lose the complete response including code, message and details in the error?

molgenis-python/src/main/resources/templates/molgenis.py.ftl
```diff
@@ -118,25 +118,42 @@ class Session():
...
+
+        Args:
+        entity -- fully qualified name of the entity
+        row_id -- id of the row of the entity to update
+        data -- dictionary mapping attribute name to non-file attribute value for the entity row, gets merged with the kwargs argument
+        **kwargs -- keyword arguments get merged with the data argument
+        '''
+        server_response_list = []
+        for key in data:
+            response = self.session.put(self.url+'v1/'+quote_plus(entity_name)+'/'+str(row_id)+'/'+key, data='"'+data[key]+'"', headers = self._get_token_header_with_content_type())
```
fdlk: So if one of the updates fails, the ones that got succesfully updated remain updated and the others are not attempted?

molgenis-python/src/main/resources/templates/molgenis.py.ftl
npklein: The message and details are in str(e). We could also use import traceback traceback.print_exc(), or some other solution. Because with response.raise_for_status() you will only get "<Response 404> Bad request", with no information about why it is a bad request.

fdlk: The error raised by raise_for_status contains the full response, so the caller has the complete information

npklein: When I add a duplicate value this is the traceback I get from raise_for_status:

Traceback (most recent call last):
  File "/Users/NPK/UMCG/git_projects/parse_rnaseq_output/run_parser.py", line 188, in <module>
    parse_output.parse_hisat(rundir_QC,connection,package=package)
  File "/Users/NPK/UMCG/git_projects/parse_rnaseq_output/PublicRNAseqParser/parse_output.py", line 291, in parse_hisat
    added_id = connection.add(package+'Hisat', data)[0]
  File "/Users/NPK/UMCG/git_projects/parse_rnaseq_output/PublicRNAseqParser/molgenis_wrapper.py", line 176, in add
    added_ids = self.session.add_all(entity_name, sanitized_data_list)
  File "/Users/NPK/UMCG/git_projects/parse_rnaseq_output/PublicRNAseqParser/molgenis.py", line 123, in add_all
    response.raise_for_status();
  File "/usr/local/lib/python3.5/site-packages/requests/models.py", line 837, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http://localhost:8080/api/v2/public_rnaseq_4_Hisat

This does not include why there was a bad request. response.json() has this:

{'errors': [{'message': "Duplicate value '1_internal_id_ATACseq-D_S3_L001' for unique attribute 'internalId_sampleid' from entity 'public_rnaseq_4_Hisat' (entity 1)"}]}

fdlk: If you catch the HttpError e then the response will be in e.response and the caller gets complete control over what to do with it. So for instance:
```
>>> try:
...   c.login('a','b')
... except requests.exceptions.HTTPError as e:
...   print(e.response.json())

{u'errors': [{u'message': u"org.springframework.security.core.userdetails.UsernameNotFoundException: unknown user 'a'"}]}
```

npklein: Ah cool! Sorry I missed that

molgenis-python/src/main/resources/templates/molgenis.py.ftl
fdlk: Given this some more thought and I think you need this functionality present in the API, so that it becomes transactional.

PR #4207, opened 2015-11-25
M3389 As Parelsnoer I want to have the correct tissue types mapped in the Promise link

tommydeboer:
- Parels having a ```MATERIAL_TYPES_SUB``` field can distinguish between tissue types (```TISSUE_PARAFFIN_EMBEDDED``` and ```TISSUE_FROZEN```)
- Parels without ```MATERIAL_TYPES_SUB``` that have tissue stored still map to both the paraffin embedded and frozen tissue types

### DIFF
--- a/molgenis-app/src/main/java/org/molgenis/app/promise/mapper/ParelMapper.java
+++ b/molgenis-app/src/main/java/org/molgenis/app/promise/mapper/ParelMapper.java
@@ -293,9 +293,9 @@ public class ParelMapper implements PromiseMapper, ApplicationListener<ContextRe
 		List<String> unknown = Lists.newArrayList();
 		for (Entity sample : promiseSampleEntities)
 		{
-			// when MATERIAL_TYPES_SUB = -1, there is no tissue stored
+			String type = sample.getString("MATERIAL_TYPES");
 			String tissue = sample.getString("MATERIAL_TYPES_SUB");
-			if (tissue != null && !tissue.equals("-1"))
+			if (type.equals("weefsel") && tissue != null)
 			{
 				if (tissueTypesMap.containsKey(tissue))
 				{
@@ -308,7 +308,6 @@ public class ParelMapper implements PromiseMapper, ApplicationListener<ContextRe
 			}
 			else
 			{
-				String type = sample.getString("MATERIAL_TYPES");
 				if (materialTypesMap.containsKey(type))
 				{
 					materialTypesMap.get(type).forEach(t -> ids.add(t));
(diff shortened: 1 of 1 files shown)


molgenis-app/src/main/java/org/molgenis/app/promise/mapper/ParelMapper.java
```diff
@@ -280,14 +290,29 @@ private Entity toAgeType(String ageType)
 		List<String> unknown = Lists.newArrayList();
 		for (Entity sample : promiseSampleEntities)
 		{
-			String type = sample.getString("MATERIAL_TYPES");
-			if (materialTypesMap.containsKey(type))
+			String tissue = sample.getString("MATERIAL_TYPES_SUB");
```
fdlk: I (obviously! :) ) don't know how the mapping works but would expect that the MATERIAL_TYPES_SUB only contains tissue types if the MATERIAL_TYPES attribute contains "Tissue" or "weefsel" or whatever.

tommydeboer: ```MATERIAL_TYPES_SUB``` will be "-1"  when there is no tissue stored. We only look up the tissue type when it is not "-1".

fdlk: I mean the other way around, does the code list used for MATERIAL_TYPES_SUB depend on the value for MATERIAL_TYPES?

So, right now, if MATERIAL_TYPES='weefsel', you get MATERIAL_TYPES_SUB=['TISSUE_PARAFFIN_EMBEDDED'|'TISSUE_FROZEN']. And this is the only type that uses MATERIAL_TYPES_SUB.

But if at some point they decide they want to distinguish between blood types as well as tissue types and suddenly if MATERIAL_TYPES='blood', you'll get MATERIAL_TYPES_SUB = ['red'|'white']

Then, without changes, you'll translate 'red' to 'unknown tissue subtype'. But you can already check the MATERIAL_TYPE = 'weefsel' before parsing MATERIAL_TYPES_SUB.

Feel free to laugh if this is not making any sense, I really don't know anything about the mapping, am just triggered by seeing the code parse X_SUB without first checking the value for X.

tommydeboer: Ah, I understand. The field ```MATERIAL_TYPES_SUB``` was specifically added for this case, since the other cases already are distinguishable (like blood), and are just named in ```MATERIAL_TYPES```. Maybe the naming of the SUB field is a bit misleading, indeed. 

I'll flip the check around for readability and future proofing.

Issue #4084, opened 2015-10-28
Importer spends silly amounts of time copying the imported entities attribute by attribute


fdlk:
Context: 
Marije wants to import a sheet with a single entity with a couple hundred attributes and a couple thousand instance rows.
We aborted the import after half an hour when not a single entity had reached the point yet where it got indexed in elasticsearch.

Debugging showed that the time is spent in the ImportWriter which copies each entity instance to a DefaultEntityImporter instance. The `DefaultEntity#set(Entity)` method that gets called by the DefaultEntityImporter constructor is taking up a lot of time. Most of the time is probably spent figuring out what the atomic attributes of the DefaultEntity are.

Which is extra silly since the answer is of course the same for each entity instance, even if this entity has a great lot of them.
Also, the entity doesn't even have self-references but resolving self-references are the only reason that the DefaultEntityImporter gets created.

mswertz:
propose to select http://www.molgenis.org/ticket/3365 to address this

PR #4214, opened 2015-12-01
Fix #4212 If one attribute is selected in a compound the table shows all columns

### LINKED ISSUE #4212: If one attribute is selected in a compound the table shows all columns
- select any entity with a compound with multiple attributes. (VCF is an easy example)
- select one attribute within the compound, and leave the rest unselected.

the dataexplorer table now shows all attributes instead of just the selected one.

tommydeboer:
- REST API returns only meta data of compound and atomic attributes that are actually requested


jjettenn:
please test again

tommydeboer:
test this please

tommydeboer:
Fixes #4084

Issue #4185, opened 2015-11-23
EMX with self-references import failure


dennishendriksen:
Reproduce:
- Import [self-references.xlsx](https://github.com/molgenis/molgenis/files/41667/self-references.xlsx)

Expected:
- Import successful

Actual:
- ```Trying to update not exsisting PersonTest entities:, grandpa_mom, sis1, grandma_dad, dad, sis0 and more.```

PR #4215, opened 2015-12-02
M3365 Importer performance enhancements

### LINKED ISSUE #4084: Importer spends silly amounts of time copying the imported entities attribute by attribute

dennishendriksen:
See [background information](https://docs.google.com/document/d/1yy2z0m0j3uQpb71ybRhX3Ie8ryz9aUQV_HGIuNTicYw/edit?usp=sharing).

Fixes #4084


tommydeboer:
test this please

tommydeboer:
test this please

jjettenn:
1. Self-referencing entities is not working
Error Message:  Trying to update not exsisting PersonTest entities:, grandpa_mom, sis1, grandma_dad, dad, sis0 and more.
Uploaded data set from the background information document.

Typo row 535: msg.append("Trying to update not exsisting ").append(repo.getName()).append(" entities:"); exsisting -> existing

jjettenn:
Can you edit the background information to show the points that are already done?

molgenis-data-excel/src/main/java/org/molgenis/data/excel/ExcelEntity.java
```diff
@@ -23,6 +24,8 @@
 	private final Map<String, Integer> colNamesMap;
 	private final List<CellProcessor> cellProcessors;
 
+	private transient Map<String, Object> cachedValueMap;
```
jjettenn: Why are not all class parameters are final transient?
cachedValueMap can this not be already instantiated?

molgenis-data-excel/src/main/java/org/molgenis/data/excel/ExcelEntity.java
```diff
@@ -44,17 +47,39 @@ public ExcelEntity(Row row, Map<String, Integer> colNamesMap, List<CellProcessor
...
 	public Object get(String attributeName)
 	{
-		Integer col = colNamesMap.get(attributeName);
-		if (col != null)
+		if (cachedValueMap == null)
+		{
+			cachedValueMap = new CaseInsensitiveLinkedHashMap<>();
+		}
+
+		Object value;
```
jjettenn: You can beter initiate value with null;

molgenis-data/src/main/java/org/molgenis/data/PackageChangeListener.java
```diff
@@ -0,0 +1,22 @@
...
+	 */
+	String getId();
+
+	/**
+	 * Callback that fires when a package property changed
+	 * 
+	 * @param package_
+	 *            updated package
+	 */
+	void onChange(Package package_);
```
jjettenn: Why package_ and not package

molgenis-data/src/main/java/org/molgenis/data/support/EntityDecorator.java
```diff
@@ -0,0 +1,340 @@
...
+import org.molgenis.data.UnknownAttributeException;
+import org.molgenis.fieldtypes.FieldType;
+import org.molgenis.fieldtypes.XrefField;
+import org.molgenis.util.MolgenisDateFormat;
+
+/**
+ * {@link Entity} decorator that uses an entity with different {@link EntityMetaData}. Values are converted on demand to
+ * the data types defined in the new EntityMetaData.
+ */
+public class EntityDecorator implements Entity
```
jjettenn: Where is this class being used?

molgenis-data/src/main/java/org/molgenis/data/meta/AttributeMetaDataRepository.java
```diff
@@ -80,57 +86,78 @@ Repository getRepository()
...
-			attributeMetaDataEntity.set(RANGE_MIN, att.getRange().getMin());
-			attributeMetaDataEntity.set(RANGE_MAX, att.getRange().getMax());
-		}
+		return add(Arrays.asList(att)).iterator().next();
+	}
 
-		if (att.getRefEntity() != null)
+	public Iterable<Entity> add(Iterable<AttributeMetaData> attrs)
+	{
+		Iterable<List<AttributeMetaData>> batches = Iterables.partition(attrs, 1000);
```
jjettenn: Why 1000?
Can it not be a constant?

molgenis-data/src/main/java/org/molgenis/data/support/DefaultAttributeMetaData.java
```diff
@@ -23,8 +47,10 @@
...
 {
-	private String name;
-	private FieldType fieldType = MolgenisFieldTypes.STRING;
+	private Map<String, AttributeChangeListener> changeListeners;
+
+	private final String name;
+	private FieldType fieldType;
 	private String description;
 	private boolean nillable = true;
 	private boolean readOnly = false;
```
jjettenn: Can you check the 
private boolean labelAttribute = false; // remove?
private boolean lookupAttribute = false; // remove?

If we can remove those params than, you can remove some of your code.

molgenis-data/src/main/java/org/molgenis/data/meta/PackageImpl.java
```diff
@@ -18,21 +23,21 @@
  */
 public class PackageImpl implements Package
 {
+	private Map<String, PackageChangeListener> changeListeners;
```
jjettenn: Initiate changeListeners and remove null checks in other places in the code

molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntityMetaData.java
```diff
@@ -69,132 +81,211 @@ public DefaultEntityMetaData(String newName, EntityMetaData entityMetaData)
...
 		this.description = entityMetaData.getDescription();
 		EntityMetaData extends_ = entityMetaData.getExtends();
 		this.extends_ = extends_ != null ? new DefaultEntityMetaData(extends_) : null;
 		this.backend = entityMetaData.getBackend();
+		AttributeMetaData idAttr = entityMetaData.getIdAttribute();
+		this.idAttrName = idAttr != null ? idAttr.getName() : null;
+		AttributeMetaData labelAttr = entityMetaData.getLabelAttribute();
+		this.labelAttrName = labelAttr != null ? labelAttr.getName() : null;
+		this.attributes = new CaseInsensitiveLinkedHashMap<>();
+
```
jjettenn: remove space

molgenis-data-excel/src/main/java/org/molgenis/data/excel/ExcelEntity.java
dennishendriksen: > Why are not all class parameters final transient?
Entity is serializable: Writing it to disk and reading it back into memory requires that the other fields still have values in order for the entity to work as before serialization. Row should not be transient by the way, but that code is already in place.

> cachedValueMap can this not be already instantiated?
Lazy initialization pattern to improve performance of ExcelEntity in case you create the entity but e.g. skip over it when iterating entities.

dennishendriksen:
> Self-referencing entities is not working

Not working in v1.13.0, see #4185 

> Can you edit the background information to show the points that are already done?

See Performance improvements & suggestions section

molgenis-data-excel/src/main/java/org/molgenis/data/excel/ExcelEntity.java
dennishendriksen: See note about lazy initialization pattern above.

molgenis-data/src/main/java/org/molgenis/data/support/EntityDecorator.java
dennishendriksen: Oops, originally this was intended to replace DefaultEntityImporter. But I got stuck in some weird self-reference code constructions. Class is removed.

molgenis-data/src/main/java/org/molgenis/data/support/DefaultAttributeMetaData.java
dennishendriksen: This is a non-trivial task. I created [this story](http://www.molgenis.org/ticket/3933) to remove these fields.

molgenis-data/src/main/java/org/molgenis/data/meta/PackageImpl.java
dennishendriksen: Lazy initialization pattern, do not allocate memory until a listener is added.

molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntityMetaData.java
dennishendriksen: Do you mean newline? I put it there on purpose to have a visual separation between entity meta properties and other fields.

molgenis-data-excel/src/main/java/org/molgenis/data/excel/ExcelEntity.java
jjettenn: Ok, for not forgetting you can use final like this:
final Object value;
Then the code do not compile if it is not instantiated.

molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntityMetaData.java
jjettenn: Okay, yes newline.

PR #4255, opened 2015-12-09
Added an EMX with advanced nested data type examples

Mark-de-Haan:
- Compound -> Xref -> Xref -> Mref -> Categorical
- Xref ID attributes both string and int
- Entity1 extends Entity0
- Entity2 extends Entity1
- Extending from different packages


dennishendriksen:
emx_advanced_import.xlsx is a vague name, how about emx_deep_nesting.xlsx?

PR #4259, opened 2015-12-09
[1.14] Fix #4254 EMX entity import with auto id NPE

### LINKED ISSUE #4254: EMX entity import with auto id NPE
Reproduce:
- Import EMX excel with a data row for a entity with auto ID attribute

Expected:
- Import successful

Actual:
- Import fails with message:
```java.lang.UnsupportedOperationException: null
	at org.molgenis.data.excel.ExcelEntity.set(ExcelEntity.java:91) ~[classes/:na]```

Entity is mutable, so ExcelEntity should implement this operation.

tommydeboer:
- Backport for 1.14.1


dennishendriksen:
Can you do the PR to molgenis:1.14 instead of molgenis:master?

PR #4267, opened 2015-12-10
M2722 Google sign-in

dennishendriksen:


jjettenn:
Known bug: Google sign in creates a redirectloop for inactive user see issue #4291

PR #4285, opened 2015-12-22
Feature/i18n data metadata

Mark-de-Haan:


Mark-de-Haan:
Test plan is everything green!

Issue #4289, opened 2015-12-28
ResourceUtils - NullPointerException is thrown


jjettenn:
14:28:32.088 [ajp-bio-8009-exec-4] ERROR org.molgenis.util.ResourceUtils - NullPointerException is thrown. Resource url: jar:file:/srv/molgenis/apache-tomcat-7.0.65/webapps/ROOT/WEB-INF/lib/molgenis-core-ui-1.15.0-SNAPSHOT.jar!/i18n.xlsx
14:28:32.101 [ajp-bio-8009-exec-4] ERROR o.s.web.context.ContextLoader - Context initialization failed
java.lang.RuntimeException: java.lang.NullPointerException
	at org.molgenis.util.ResourceUtils.getFile(ResourceUtils.java:85) ~[molgenis-core-1.15.0-SNAPSHOT.jar:1.15.0-SNAPSHOT]
	at org.molgenis.util.ResourceUtils.getFile(ResourceUtils.java:27) ~[molgenis-core-1.15.0-SNAPSHOT.jar:1.15.0-SNAPSHOT]
	at org.molgenis.ui.I18nStringsPopulator.onApplicationEvent(I18nStringsPopulator.java:49) ~[molgenis-core-ui-1.15.0-SNAPSHOT.jar:1.15.0-SNAPSHOT]
	at org.molgenis.ui.I18nStringsPopulator.onApplicationEvent(I18nStringsPopulator.java:28) ~[molgenis-core-ui-1.15.0-SNAPSHOT.jar:1.15.0-SNAPSHOT]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:151) ~[spring-context-4.1.8.RELEASE.jar:4.1.8.RELEASE]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:128) ~[spring-context-4.1.8.RELEASE.jar:4.1.8.RELEASE]
	at org.springframework.context.support.AbstractApplicationContext.publishEvent(AbstractApplicationContext.java:331) ~[spring-context-4.1.8.RELEASE.jar:4.1.8.RELEASE]
	at org.springframework.context.support.AbstractApplicationContext.finishRefresh(AbstractApplicationContext.java:775) ~[spring-context-4.1.8.RELEASE.jar:4.1.8.RELEASE]
[...]

PR #4290, opened 2015-12-29
Fix2 #4289 ResourceUtils - NullPointerException is thrown

jjettenn:
Second try

### DIFF
--- a/molgenis-core-ui/src/main/java/org/molgenis/ui/I18nStringsPopulator.java
+++ b/molgenis-core-ui/src/main/java/org/molgenis/ui/I18nStringsPopulator.java
@@ -53,9 +53,8 @@ public class I18nStringsPopulator implements ApplicationListener<ContextRefreshe
 
 		// "i18n is saved as a Application/Library resource.
 		// "It is not possible to use it as a file but should streamed as a resource"
-		ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
-		InputStream is = classLoader.getResourceAsStream(i18nFileName);
-		
+		InputStream is = I18nStringsPopulator.class.getClassLoader().getResourceAsStream(i18nFileName);
+
 		try
 		{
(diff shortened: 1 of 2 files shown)


molgenis-core-ui/src/main/java/org/molgenis/ui/I18nStringsPopulator.java
```diff
@@ -33,26 +34,50 @@
...
 	@Override
 	public void onApplicationEvent(ContextRefreshedEvent event)
 	{
 		LOG.info("Importing i18n strings...");
+		final String i18nFileName = "i18n.xlsx";
+
+		// "i18n is saved as a Application/Library resource.
+		// "It is not possible to use it as a file but should streamed as a resource"
+		ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+		InputStream is = classLoader.getResourceAsStream(i18nFileName);
```
fdlk: getClass().getResourceAsStream is more common I think?

jjettenn: It is more common but it will result in a java.lang.NullPointerException. Resource is not directly in the classloader.

fdlk: Not entirely sure if I dare to ask, but how is that possible?

jjettenn: Look at (Warning for users of Tomcat 7 and below) in: http://stackoverflow.com/questions/676250/different-ways-of-loading-a-file-as-an-inputstream 

It should be the same but it is not.

jjettenn: I18nStringsPopulator.class.getClassLoader().getResourceAsStream(i18nFileName) should probably also work. Resource is in the same package.

PR #4293, opened 2016-01-04
Permission name must be uppercase [master]

erwinwinder:


Mark-de-Haan:
Master will be updated automatically

PR #4304, opened 2016-01-05
Stabilize POMs

Mark-de-Haan:
Accept #4303 first!


fdlk:
Heh. Should've accepted #4303 first?

PR #4305, opened 2016-01-05
Fix #4296: Aggregates with threshold smaller than rendered wrong

### LINKED ISSUE #3252: Javascript files (and other static resources) get encoded wrong
The select 2 box displays an encoding error:
searching

The javascript file contains an UTF-8 character: `return "Searching…";`.
In the browser (Safari and Chrome tested) you see it interpreted as iso-latin, `Searchingâ€¦`

### LINKED ISSUE #4296: Aggregates with threshold smaller than rendered wrong
Reproduce:
- Set 'Aggregate threshold' to 10 in App Settings, Save changes
- Go to Data explorer plugin, Select MolgenisUser entity, Select Aggregates tab

Expected:
- Smaller than symbol is rendered

Actual:
- Smaller than symbol is rendered incorrectly
bug_aggregates

fdlk:
This works for me locally, still needs to be tested on a VM with apache httpd and tomcat instead of jetty.
Make sure to clear your cache when testing.

Also fixes #3252 which I've closed as a duplicate.

The fix changes the encoding used to serve all FreeMarker views from iso-latin-1 to UTF-8, so the potential impact is across the board, wherever non-ascii characters are served. Still, I'd argue that this is how we want and expect the views to be served.

### DIFF
--- a/molgenis-core-ui/src/main/java/org/molgenis/ui/MolgenisWebAppConfig.java
+++ b/molgenis-core-ui/src/main/java/org/molgenis/ui/MolgenisWebAppConfig.java
@@ -331,7 +331,7 @@ public abstract class MolgenisWebAppConfig extends WebMvcConfigurerAdapter
 		FreeMarkerViewResolver resolver = new FreeMarkerViewResolver();
 		resolver.setCache(true);
 		resolver.setSuffix(".ftl");
-
+		resolver.setContentType("text/html;charset=UTF-8");
 		return resolver;
 	}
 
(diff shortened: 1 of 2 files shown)


Mark-de-Haan:
This will have to be tested tomorrow on molgenis01 because locally I can not reproduce #4296

fdlk:
You can reproduce locally by setting environment=production in the molgenis-server.properties.
But the problem might be influenced by the way the jarfile is hosted, i.e. locally we host with jetty and perhaps using tomcat / ajp / httpd it's different

PR #4316, opened 2016-01-11
Streaming data API M3913

dennishendriksen:


molgenis-core-ui/src/main/java/org/molgenis/ui/EntityListenerRepositoryDecorator.java
```diff
@@ -151,6 +170,22 @@ public void update(Iterable<? extends Entity> records)
 	}
 
 	@Override
+	public void update(Stream<? extends Entity> entities)
+	{
+		if (entityListeners != null)
+		{
+			entities = entities.filter(entity -> {
```
jjettenn: Why using filter is all results in true? Why not forEach?

dennishendriksen: forEach is a terminal operation, the decorated repository requires a stream that can be consumed.

molgenis-security/src/test/java/org/molgenis/security/user/MolgenisUserDecoratorTest.java
```diff
@@ -57,6 +68,57 @@ public void addEntity()
...
+	{
+		Entity entity0 = mock(Entity.class);
+		entity0.set(MolgenisUser.PASSWORD_, "password");
+		Stream<Entity> entities = Stream.of(entity0);
+		ArgumentCaptor<Stream<Entity>> captor = ArgumentCaptor.forClass((Class) Stream.class);
+		doNothing().when(decoratedRepository).update(captor.capture());
+		decoratedRepository.update(entities);
+		assertEquals(captor.getValue().collect(Collectors.toList()), Arrays.asList(entity0));
+		verify(entity0).set(eq(MolgenisUser.PASSWORD_), anyString());
+		// TODO add authority tests
```
jjettenn: Did you forgot this? // TODO add authority tests

molgenis-data/src/main/java/org/molgenis/data/AutoValueRepositoryDecorator.java
```diff
@@ -289,4 +336,34 @@ public void removeEntityListener(EntityListener entityListener)
...
 		decoratedRepository.removeEntityListener(entityListener);
 	}
+
+	private List<AttributeMetaData> getAutoAttrs()
+	{
+		return stream(getEntityMetaData().getAtomicAttributes().spliterator(), false).filter(AttributeMetaData::isAuto)
+				.collect(Collectors.toList());
+	}
+
+	private Entity initAutoAttrs(Entity entity, List<AttributeMetaData> autoAttrs)
```
jjettenn: Here

PR #4317, opened 2016-01-12
Improve VCF commandline annotation performance M3990

dennishendriksen:


molgenis-data-annotators/src/main/java/org/molgenis/data/annotator/tabix/TabixRepository.java
```diff
@@ -122,19 +122,26 @@ public EntityMetaData getEntityMetaData()
 		try
 		{
 			org.molgenis.data.annotator.tabix.TabixReader.Iterator iterator = reader.query(queryString);
-			String line = iterator.next();
-			while (line != null)
+			if (iterator != null)
 			{
-				Entity entity = toEntity(line);
-				if (entity.getLong(positionAttributeName) == pos)
```
bartcharbon: Tabix can sometimes return a row that does not exactly match the query, this check should remain!

molgenis-data-annotators/src/main/java/org/molgenis/data/annotator/tabix/TabixRepository.java
```diff
@@ -122,19 +122,26 @@ public EntityMetaData getEntityMetaData()
...
+				String line = iterator.next();
```
jjettenn: why not using hasNext() instead of while (line != null)

dennishendriksen: I did not touch this code, the changes I made were adding the if (iterator != null) { ... } block around existing code.

PR #4326, opened 2016-01-18
GoogleFriday: use vcf filename (processed to comply with entityname rules) as default

bartcharbon:


molgenis-data-import/src/main/resources/templates/OptionsWizardPage.ftl
```diff
@@ -4,7 +4,13 @@
 		<div class="col-md-4">
 			<div class="form-group">
             	<label class="control-label" for="name">Entity name *</label>
-            	<input type="text" class="form-control" name="name" required placeholder="Enter entity name">
+            	<input type="text" class="form-control" name="name" required placeholder="Enter entity name"
+					   value="${wizard.file.name
+					   ?keep_before_last(".")
```
dennishendriksen: Won't fly with .vcf.gz

PR #4338, opened 2016-01-21
M3777 As Mapping Service user I can update an existing entity

tommydeboer:
- It is now possible to map to a target repository more than once
- New source entities are added to the target
- Old source entities are updated in the target
- Removed source entities are removed from the target


tommydeboer:
test this please

tommydeboer:
test this please

Issue #4337, opened 2016-01-21
Using other themes than the molgenis theme, will cause a glyphicon not found error


Mark-de-Haan:
Choose a theme and open the dataexplorer with console open
You will see three 404 errors.

Cause: themes will try to look for these icons in ../fonts. Our bootstrap themes are nested one deeper, resulting in css/fonts which does not exist.

PR #4342, opened 2016-01-25
M3494 Implemented modular javascript to get proper frontend dependency management

Mark-de-Haan:


dennishendriksen:
Eclipse: Building workspace Invoking 'Maven Project Builder' on '/molgenis-core-ui' stuck on 51%

dennishendriksen:
```[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:0.0.27:install-node-and-npm (install node and npm) on project molgenis-core-ui: The plugin com.github.eirslett:frontend-maven-plugin:0.0.27 requires Maven version 3.1.0 -> [Help 1]```
Update/mention the required Maven version in the MOLGENIS documentation

**Comment by dennishendriksen** [comment]
Error performing ```mvn clean install```:
```[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:0.0.27:npm (npm test) on project molgenis-core-ui: Failed to run task: 'npm test --color=false' failed. (error code 1) -> [Help
 1]```

**Comment by dennishendriksen** [comment]
```mvn clean install``` on molgenis-core-ui takes 240 seconds on my laptop

**Comment by dennishendriksen** [comment]
After building and switching to master branch
```
PS D:\git\molgenis> mvn clean install -DskipTests
...
PS D:\git\molgenis> git checkout master
error: Your local changes to the following files would be overwritten by checkout:
        molgenis-core-ui/package.json
Please, commit your changes or stash them before you can switch branches.
Aborting
PS D:\git\molgenis> git diff
warning: LF will be replaced by CRLF in molgenis-core-ui/package.json.
The file will have its original line endings in your working directory.
```

**Comment by dennishendriksen** [comment]
Transferred file size visiting MOLGENIS main page:
```
development			2.5MB
production			1.9MB
webpack development 9.4MB
webpack production	5.7MB
```
Note that these sizes are for Jetty on localhost, Tomcat might compress resources

**Comment by fdlk** [comment]
We can reduce the amount of used ace themes and modules, that would probably help. Or package the non-standard ones separately.

**Comment by fdlk** [comment]
The first build on a machine takes longer than the next ones because the node_modules directory needs to be filled.

**Comment by fdlk** [comment]
The message about updated package.json comes from the fact that the pom version gets written into it.

I tried first with a package.json resource that gets filtered into the proper package.json by filling in the pom version but it is very confusing to have two versions of package.json lying around.
npm install --save updates the filtered copy and you have to manually move changes to the to-be-filtered resource. Also you cannot/shouldn't run npm in the directory containing the package.json resource then even though it looks like an npmmable directory since it contains package.json.

Two ways to go about this that I can think of:
* Commit the package.json change when the pom version changes and package.json gets updated. (current state of affairs).
* Forget about having a semantically correct package.json version. Set it to 1.0.0 or so, and remove the gulp task that updates it.

What do you think @dennishendriksen ?

**Comment by dennishendriksen** [comment]
The differences in package.json seems to be in the whitespace, no actual content changes.

**Comment by dennishendriksen** [comment]
I understood from @Mark-de-Haan that there are now two seperate places to define whether you are dealing with a development or production environment. We should only have one place to define it.

**Comment by dennishendriksen** [comment]
How did you solve the one-node install per module? Are JavaScript bundles produced in the maven module that contains the JavaScript?

**Comment by Mark-de-Haan** [comment]
Production development profile is specified in the parent pom, and activated by running -P production or development

the molgenis-server.properties also still contains the environment property, although we were not sure wether this was still used.

**Comment by Mark-de-Haan** [comment]
Node is installed via the core-ui pom. The webpack config in the core-ui defines the bundles for every module. I need to check this but I think it is possible to specify were a bundle is placed. 

But then we get to the problem we discussed yesterday: What if I use a stable version of the core and develop my own module, I am not able to edit the webpack bundle from there.

It would be possible if a new module also used its own webpack config.

**Comment by dennishendriksen** [comment]
@Mark-de-Haan Verify if used:
- If yes, refactor so it uses the 'other' method and write a migration step to remove it from molgenis-server.properties.
- If no, write a migration step to remove it from molgenis-server.properties.

**Comment by dennishendriksen** [comment]
@Mark-de-Haan "What if I use a stable version of the core and develop my own module, I am not able to edit the webpack bundle from there." We need to solve this first. core-ui should of course never have dependencies on other plugin modules (if that is the case right now).

**Comment by dennishendriksen** [comment]
@Mark-de-Haan We need to update the developer guidelines regarding the production development profile as well as add additional information required for developers to know how to develop new modules

**Comment by dennishendriksen** [comment]
@fdlk reduce the amount of data transferred: Good idea, lets do it.

**Comment by dennishendriksen** [comment]
@fdlk Subsequent clean installs on molgenis-core-ui take ~3 minutes, the first build takes longer than that

File: molgenis-core-ui/package.json [hunk 1]
```diff
@@ -0,0 +1,86 @@
+{
+  "name": "MOLGENIS",
+  "version": "1.16.0-SNAPSHOT",
```
dennishendriksen: Can we use Maven process-resources to replace a version token with the current version?

molgenis-core-ui/src/main/javascript/modules/react-components/CheckboxWithLabel.js
```diff
@@ -0,0 +1,31 @@
+// CheckboxWithLabel.js
+import React from 'react';
+
+class CheckboxWithLabel extends React.Component {
```
dennishendriksen: What is the motivation to use a different style of writing React components? If we want to switch for a different style we should refactor current components.

molgenis-core-ui/src/main/javascript/modules/react-components/Form.js
```diff
@@ -272,9 +285,10 @@
 					)
 				)
 			) : null;
-					
+
+			//TODO: Rename; this shadows the Form class!
```
dennishendriksen: Do it before merging this PR

molgenis-core-ui/src/main/javascript/modules/react-components/FormControlGroup.js
```diff
@@ -102,10 +106,6 @@
...
-    // export component
-    molgenis.ui = molgenis.ui || {};
-    _.extend(molgenis.ui, {
-        FormControlGroup: React.createFactory(FormControlGroup)
-    });
-}(_, React, molgenis));
\ No newline at end of file
+
+const FormControlGroupFactory = React.createFactory(FormControlGroup);
+export default FormControlGroupFactory;
```
dennishendriksen: To be consistent with other module syntax merge line 110 and 111.

molgenis-core-ui/src/main/javascript/modules/react-components/Table.js
```diff
@@ -1009,9 +1029,5 @@
...
 
-	// export component
-	molgenis.ui = molgenis.ui || {};
-	_.extend(molgenis.ui, {
-		Table: React.createFactory(Table)
-	});
-}(_, React, molgenis));
\ No newline at end of file
+const TableFactory = React.createFactory(Table);
+export default TableFactory;
```
dennishendriksen: Merge line 1032 and 1033 to be consistent with syntax in other modules

molgenis-core-ui/src/main/javascript/modules/react-components/index.js
```diff
@@ -0,0 +1,64 @@
+/**
```
dennishendriksen: What does this file do? Can you put it in de docs on the top of this file?

molgenis-core-ui/src/main/javascript/modules/react-components/mixin/index.js
```diff
@@ -0,0 +1,23 @@
+/**
```
dennishendriksen: What does this file do? Can you put it in de docs on the top of this file?

molgenis-core-ui/src/main/javascript/modules/react-components/wrapper/Ace.js
```diff
@@ -34,7 +34,7 @@
 			};
 		},
 		getInitialState: function() {
-			return {value: this.props.value};
+			return {value: this.props.value || ''};
```
dennishendriksen: Is this a bug fix? If yes, create a issue with seperate PR so that it will end up in the release notes and everyone is aware of this fix. If no, why this change?

molgenis-core-ui/src/main/javascript/modules/react-components/wrapper/Ace.js
```diff
@@ -42,9 +42,11 @@
 		componentDidMount: function() {
 			var container = this.refs.editor.getDOMNode();
 			var editor = ace.edit(container);
+			require('brace/theme/'+this.props.theme);
```
dennishendriksen: What is happening here, shouldn't this be dealt with using import?

molgenis-core-ui/src/main/javascript/modules/react-components/wrapper/index.js
```diff
@@ -0,0 +1,19 @@
+/**
```
dennishendriksen: What does this file do? Can you put it in de docs on the top of this file?

molgenis-core-ui/src/main/javascript/modules/rest-client/index.js
```diff
@@ -0,0 +1,7 @@
+import RestClientV1 from "./RestClientV1";
```
dennishendriksen: What does this file do? Can you put it in de docs on the top of this file?

molgenis-core-ui/src/main/javascript/molgenis-global-ui-webpack.js
```diff
@@ -0,0 +1,5 @@
+import ui from 'react-components';
```
dennishendriksen: What does this file do? Can you put it in de docs on the top of this file?

molgenis-core-ui/src/main/javascript/molgenis-global-webpack.js
```diff
@@ -0,0 +1,31 @@
+window.top.$ = require('jquery');
```
dennishendriksen: What does this file do? Can you put it in de docs on the top of this file?

molgenis-core-ui/src/main/javascript/molgenis-vendor-webpack.js
```diff
@@ -0,0 +1,24 @@
+import "jquery";
```
dennishendriksen: What does this file do? Can you put it in de docs on the top of this file?

dennishendriksen:
Rename PR so that it is clear to users what changed when reading the release notes

molgenis-core-ui/src/main/javascript/modules/react-components/CheckboxWithLabel.js
Mark-de-Haan: Ah, this is the Jest example component, we can remove this.

molgenis-core-ui/src/main/javascript/modules/react-components/index.js
Mark-de-Haan: an index.js is basically like an index.html. It is automatically executed. It is used to produce legacy javascript bundles for the wrappers, react components, and mixins.

So if we want the entire mixin bundle, you can do require('mixins').

@fdlk can probably explain it better

molgenis-core-ui/src/main/javascript/modules/react-components/wrapper/Ace.js
Mark-de-Haan: Webpack allows for asynchronous imports. This is an example of such an import.
We only retrieve the theme when the component is mounted, to reduce the initial load to the server. Ace brings a lot of kb's to the client, we try to lessen it by asynch requires

molgenis-core-ui/src/main/javascript/modules/react-components/wrapper/Ace.js
fdlk: This was a bug fix for a bug that we triggered by importing a higher version of ace than the one the master used. We went back to the lower version so this can be removed. Well spotted.

fdlk:
Regarding file sizes:
I've split off the ace modes and that results in nicer build sizes.
Production build:

```
-rw-r--r--   1 fkelpin  staff  4007377 Jan 28 13:42 molgenis-ace.js
-rw-r--r--   1 fkelpin  staff   135835 Jan 28 13:42 molgenis-global-ui.js
-rw-r--r--   1 fkelpin  staff     8591 Jan 28 13:42 molgenis-global.js
-rw-r--r--   1 fkelpin  staff  1174873 Jan 28 13:42 molgenis-vendor-bundle.js
```

So question is, I think: Do we have a list of ace modes and themes that we actually use?
Then we can include those explicitly in the vendor bundle.
And do we want to provide the complete rest to existing users?

molgenis-core-ui/src/main/javascript/modules/react-components/Form.js
fdlk: But it's an existing issue, unrelated to the PR.

mswertz: But if you notice the todo why not fix it; is it a huge amount of work?

molgenis-core-ui/README.md
```diff
@@ -0,0 +1,83 @@
...
+We use the [frontend maven plugin](https://github.com/eirslett/frontend-maven-plugin)
+to allow maven to install node and run npm and other javascript tools.
+
+The javascript files in and other static resources src/main/javascript are packed into bundles using webpack.
+
+### Directory structure
+```
+|-- molgenis-core-ui
+    |-- README.md // this file
+    |-- gulpfile.js // javascript build actions, used to write version from pom.xml to package.json
```
mswertz: Please explain better what this does. As I understand this is used to read the POM and generate javascript with proper version number

molgenis-core-ui/README.md
```diff
@@ -0,0 +1,83 @@
...
+    |   |-- npm.cmd
+    |-- node_modules // npm modules listed in package.json get downloaded to here by npm and read by webpack and node
+    |   |-- [...]
+    |-- src
+    |   |-- main
+    |   |   |-- [...]
+    |   |   |-- javascript
+    |   |   |   |-- molgenis-global-ui-webpack.js // bundle files specify the contents of the bundles
+    |   |   |   |-- molgenis-global-webpack.js
+    |   |   |   |-- molgenis-vendor-webpack.js
```
mswertz: ouch, what a huge number of webpack configs.
Is this only once or would each module need to this as well?

molgenis-core-ui/README.md
```diff
@@ -0,0 +1,83 @@
...
+### Directory structure
+```
+|-- molgenis-core-ui
+    |-- README.md // this file
+    |-- gulpfile.js // javascript build actions, used to write version from pom.xml to package.json
+    |-- package.json // describes what npm packages are used
+    |-- pom.xml // the maven buildfile
+    |-- webpack.config.js // webpack configuration, lists the entry points for the bundles and the loaders
+    |-- webpack.test.config.js // webpack configuration for tape tests
+    |-- node // node gets installed here by frontend maven plugin
```
mswertz: Is this the output of a maven step? Then move this to /target/...

molgenis-core-ui/README.md
```diff
@@ -0,0 +1,83 @@
...
+    |   |   |   |-- molgenis-vendor-webpack.js
+    |   |   |   |-- modules // contain the source files for the molgenis modules, interpreted by the babel loader, included in the bundles
+    |   |   |   |   |-- i18n
+    |   |   |   |   |   |-- I18nStrings.js
+    |   |   |   |   |-- react-components
+    |   |   |   |   |   |-- AggregateTable.js
+    |   |   |   |   |   |-- AlertMessage.js
+    |   |   |   |   |   |-- [...]
+    |   |   |   |   |   |-- index.js
+    |   |   |   |   |   |-- css // if css files are required by the javascript files in the module, webpack bundles them
```
mswertz: Is there a way to get css (which is evil) can be closer to the .js/.jsx it belongs to?
And do we have ground rules for 'good' css?
E.g. like standards refered to in https://www.youtube.com/watch?v=VkTCL6Nqm6Y (minute ~20 and the Q&A)

molgenis-core-ui/package.json
```diff
@@ -0,0 +1,80 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/molgenis/molgenis.git"
+  },
+  "author": "MOLGENIS",
+  "license": "LGPL-3.0",
+  "bugs": {
+    "url": "https://github.com/molgenis/molgenis/issues"
+  },
+  "homepage": "https://github.com/molgenis/molgenis#readme",
```
mswertz: homepage is http://molgenis.github.io or http://molgenis.org

molgenis-core-ui/package.json
```diff
@@ -0,0 +1,80 @@
...
+  "dependencies": {
```
mswertz: I suppose we need a story to assess if we really need al this stuff!

molgenis-core-ui/src/main/java/org/molgenis/ui/settings/AppDbSettings.java
```diff
@@ -248,10 +248,10 @@ public void setBootstrapTheme(String bootstrapTheme)
 	{
 		// verify that css file exists
 		PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
-		Resource resource = resolver.getResource("/css/themes/" + bootstrapTheme);
+		Resource resource = resolver.getResource("/css/" + bootstrapTheme);
```
mswertz: why this change? is there a common practice for bootstrap themes?

molgenis-core-ui/src/main/javascript/modules/react-components/AttributeControl.js
```diff
@@ -61,22 +74,22 @@
 				
 				switch(attr.fieldType) {
```
mswertz: Why is there a big switch here?
I would expect that
- Controls have all the logic they need to render themselves from a attributemetdata
- New controls can register themselves
(so then this becomes extensible by new data/control types)

molgenis-core-ui/src/main/javascript/modules/react-components/AggregateTable.js
```diff
@@ -12,7 +14,7 @@
 	 * @memberOf component
 	 */
 	var AggregateTable = React.createClass({
-		mixins: [molgenis.ui.mixin.DeepPureRenderMixin],
+		mixins: [DeepPureRenderMixin],
```
mswertz: Can these mixins not be pulled up to superclass (it seems we need them almost everywhere)

molgenis-core-ui/src/main/javascript/modules/react-components/AlertMessage.js
```diff
@@ -1,14 +1,14 @@
-/* global _: false, React: false, molgenis: true */
-(function(_, React, molgenis) {
-	"use strict";
-	
+import React from "react";
+import DeepPureRenderMixin from "./mixin/DeepPureRenderMixin";
+import Icon from "./Icon";
+
 	var div = React.DOM.div, span = React.DOM.span, button = React.DOM.button;
```
mswertz: What is the style guide on indents?
I would expect this not indented (many more places)

molgenis-core-ui/src/main/javascript/modules/react-components/EntitySelectBox.js
```diff
@@ -1,16 +1,22 @@
-/* global _: false, React: false, molgenis: true */
-(function(_, React, molgenis) {
-	"use strict";
-	
+import React from "react";
+import DeepPureRenderMixin from "./mixin/DeepPureRenderMixin";
+import EntityLoaderMixin from "./mixin/EntityLoaderMixin";
+import ReactLayeredComponentMixin from "./mixin/ReactLayeredComponentMixin";
+import RestClient from "rest-client/RestClientV1";
```
mswertz: Why here not ./rest-client/... ?

molgenis-core-ui/src/main/javascript/modules/react-components/Form.js
```diff
@@ -1,15 +1,28 @@
...
+import RestClient from "rest-client/RestClientV1";
+import DeepPureRenderMixin from "./mixin/DeepPureRenderMixin";
+import EntityLoaderMixin from "./mixin/EntityLoaderMixin";
+import EntityInstanceLoaderMixin from "./mixin/EntityInstanceLoaderMixin";
+import ReactLayeredComponentMixin from "./mixin/ReactLayeredComponentMixin";
+import Modal from "./Modal";
+import Spinner from "./Spinner";
+import Button from "./Button";
+import _ from "underscore";
+import $ from "jquery";
```
mswertz: should we try to get rid of $?

molgenis-core-ui/src/main/javascript/modules/react-components/Table.js
```diff
@@ -18,7 +38,7 @@
 		},
 		_canExpandAttr: function(attr, path) {
 			// expanding mrefs in expanded attr not supported
-			return molgenis.isRefAttr(attr) && !(molgenis.isMrefAttr(attr) && _.size(path) > 0);
+			return isRefAttr(attr) && !(isMrefAttr(attr) && _.size(path) > 0);
```
mswertz: Hmm, is this better then importing the container of these functions?

molgenis-core-ui/src/main/javascript/modules/react-components/AggregateTable.js
```diff
@@ -39,7 +41,7 @@
 		},
 		render: function() {
 			if(this.state.data === null) {
-				return molgenis.ui.Spinner(); // entity not available yet
+				return Spinner(); // entity not available yet
```
mswertz: We return spinners like crazy. Is there not a more elegant way to return the component that you actually want to show and that that will render a spinner if not ready (I am looking for a way to reduce this code).
Moreover, we now throw many seperate spinners while ideally you would want only one, i.e., only the container throws a spinner (sorry, I am not being very clear am I?)

molgenis-core-ui/src/main/javascript/modules/react-components/wrapper/Ace.js
mswertz: So this is a very cool feature webpack now gives us, cool

molgenis-core-ui/src/main/javascript/modules/react-components/EntitySelectBox.js
fdlk: We made rest-client into a separate module. It's also used on pages without react.

molgenis-core-ui/src/main/javascript/modules/react-components/AlertMessage.js
fdlk: Yes. Except then git no longer recognizes the updates as file moves and changes at the top and bottom. It simply sees them as deletions and creations. Then it's hard to judge the pull request.

molgenis-core-ui/README.md
fdlk: No, it's supposed to document what all the files do

molgenis-core-ui/src/main/javascript/modules/react-components/Form.js
fdlk: I'm worried that rewriting it causes new bugs, we're doing this with our refactor hat on: rewrite as little as possible, leave everything working the same.

molgenis-core-ui/src/main/javascript/modules/react-components/index.js
fdlk: index.js files export the public interface of a module

molgenis-core-ui/README.md
fdlk: Good question. I don't know yet how to bundle multiple modules. The most efficient result is probably obtained if we pack all molgenis entries in one big global sweep. On the other hand we have a modular system where we have build artifacts per module.

There's no technical reason why we must pick one or the other, it's just figuring out what works best for us. But I don't think it's part of this story.

molgenis-core-ui/README.md
fdlk: webpacking css is tricky because the 'cascading' bit means that the order in which you include them matters. Imports specify what you need and only care that the import happens before you get  executed. So these two models are at odds. For this reason, we've kept bootstrap.css and the bootstrap theme separate. Require-ing it causes more trouble than it solves.
For small components it's nice to require their css and the style.

We need to make some rules for css, also for what syntax we use for the components. But again, I'd like to keep that out of the story.

molgenis-core-ui/src/main/javascript/modules/react-components/Form.js
fdlk: Yes! (but not in this story)

molgenis-core-ui/src/main/javascript/modules/react-components/Table.js
fdlk: Unsure. In a far and distant future, you can then pack only those functions that you use, see rollup-js.

molgenis-core-ui/src/main/javascript/modules/react-components/AggregateTable.js
fdlk: Yes! (But not in this story)

fdlk:
I'll try to make an inventory of the non-line-related issues here plus how (if) they're addressed.

### npm run-script test fails on windows
I've used && instead of ; to separate the steps, that should fix it I'm told.

### Required maven version changes
On http://molgenis.github.io/develop/start we recommend eclipse luna, which I believe already comes with maven 3.3.3 embedded. Is there a different place where we mention maven versions?1

### Eclipse stuck on 51%
It's not stuck, you can open the maven console in eclipse to check what it's doing. But production build took very long cause it was packing up all ace modes. I've taken a look and picked the relevant modes, they get packed, the rest gets discarded.

### Long build times
They went down substantially when I removed the non-used ace modes.
Production build without tests takes 1:22 now, of which 1:07 is spent in webpack packing and minifying the javascript.
Development build without tests takes 0:30 now, of which 0:14 is spent in webpack packing the javascript.

### Line endings in package.json on windows machine
I've updated the gulp script so that it doesn't update the package.json file if the version number matches the pom.

### File sizes
Went down with reduction of ace modes bundled.
Production build of the vendor bundle now somehow oscillates between 2Mb and 1Mb(?!), still looking into that.
Dev build of the vendor bundle is 4Mb
Size of our own code is 200 Kb or so, hardly worth mentioning.

### One node install per module
We currently have only one module that needs it, so this is not yet an issue.
We've experimented with a single global build sitting in the molgenis-app folder, that worked too but gives little added value for now. Also, you don't want to rebuild much when you're simply starting up the app.

### Production profile vs property in the molgenis-server.properties
I don't really like this either. I don't quite understand how the old system works.
I don't think that writing a migration script for development settings is the most productive use of our time, though.

molgenis-core-ui/src/main/javascript/modules/react-components/AggregateTable.js
fdlk: React folks are thinking about a solution for this. See https://facebook.github.io/react/docs/reusable-components.html

> Instead, we're working on making it easier to support such use cases without resorting to mixins.

molgenis-core-ui/src/main/java/org/molgenis/ui/settings/AppDbSettings.java
fdlk: This fixes #4337 . We ran into this when we were fiddling with the import order of the theme and the bootstrap css. We fixed the bug cause otherwise we couldn't test if the theme switcher still worked after this PR. (Cause it already wasn't working properly before the PR)

molgenis-core-ui/README.md
fdlk: Done.

molgenis-core-ui/README.md
fdlk: We found a post by sokra, explaining how one config can read the other and then modify it

Issue #4368, opened 2016-01-26
Unable to upload test-edgecases.vcf


Mark-de-Haan:
the molgenis-annotators/src/test/resources/test-edgecases.vcf gives the following error when trying to import:

org.molgenis.data.MolgenisDataException: Validation error in entity [testEdgecases_Sample]: Invalid characters in: [.] Only letters (a-z, A-Z), digits (0-9), underscores (_) and hashes (#) are allowed.

erwinwinder:
Duplicate of https://github.com/molgenis/molgenis/issues/4331 ?

Mark-de-Haan:
No I don't think so. This is probably an old vcf test file that was never migrated to work with the newer validation rules or something

bartcharbon:
We support uploading VCF's it strike me as "odd" if we need to migrate the actual file...

mswertz:
If its wrong format then this is a 'wontfix'?

tommydeboer:
Fixed in #4395

Issue #3965, opened 2015-10-16
VCF files that contain reserved words cannot be imported


joerivandervelde:
VCF files sometimes contain INFO fields which turn out to be MOLGENIS reserved words. For example, in GoNL there is an INFO field named "SET", which causes the imported to fail with an error.

PR #4395, opened 2016-01-28
Fix #4355 #3965 #4368 + small vcf import entity/attribute name improvements/checks

### LINKED ISSUE #4355: When importing a VCF with an existing entity name, the error message is very unclear, and you have no idea what the error actually is
## Reproduce
- Import test_samples.vcf, call it test_samples
- Import test_samples.vcf again, and call it test_samples

## Expected
An error message telling you this entity name is already taken within this package

## Observed
```File did not pass validation see results below. Please resolve the errors and try again.```
Followed by ```Importable: No```

### PR DESCRIPTION
**Pull request by bartcharbon**
improvements:
- replaces illegal column names in Sample and INFO fields, leaving the label the same as the name in the file
- limits the entityname taking into account the postfix "_SAMPLES"
- generates a default name based on both ".vcf" and ".vcf.gz" (was only working for ".vcf")
- adds replacing "-" in the default entity name, this one was missing from the regex.
- checks for exisiting entities with the same name, showing an meaningful message

### DIFF
--- a/molgenis-data-import/src/main/resources/templates/OptionsWizardPage.ftl
+++ b/molgenis-data-import/src/main/resources/templates/OptionsWizardPage.ftl
@@ -4,6 +4,7 @@
 		<div class="col-md-4">
 			<div class="form-group">
             	<label class="control-label" for="name">Entity name *</label>
+                <i>(Only letters (a-z, A-Z), digits (0-9), underscores (_) and hashes (#) are allowed.)</i>
             	<input type="text" class="form-control" name="name" required placeholder="Enter entity name"
 					   value="${wizard.file.name
 					   ?replace("\\.vcf|\\.vcf\\.gz",'','r') <#-- remove extention -->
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: molgenis-data-import/src/main/resources/templates/OptionsWizardPage.ftl [hunk 1]
```diff
@@ -4,13 +4,14 @@
...
             	<label class="control-label" for="name">Entity name *</label>
+                <i>(Only letters (a-z, A-Z), digits (0-9), underscores (_) and hashes (#) are allowed.)</i>
             	<input type="text" class="form-control" name="name" required placeholder="Enter entity name"
 					   value="${wizard.file.name
-					   ?keep_before_last(".")
-					   ?js_string[0..*29]
-					   ?replace("\\.|\\*|\\$|\\&|\\%|\\^|\\(|\\)|\\#|\\!|\\@|\\?",'_','r')
-					   ?replace("^[0-9]",'_','r')
-					   }">
+					   ?replace("\\.vcf|\\.vcf\\.gz",'','r') <#-- remove extention -->
```
tommydeboer: When I upload ```cadd_test.vcf.gz``` the name becomes ```cadd_test_gz```

PR #4398, opened 2016-01-29
Fix #4397 migration of molgenisUser from older molgenis versions

### LINKED ISSUE #4397: migration from 1.8.1 to 1.16 fails because of changes in the JPA entities
Because the description for the password field was changed between 1.8 and 1.15 the migration fails when starting out from this version.

MOLGENIS will complain about the attempt to change metadata for the password field on startup.

bartcharbon:
manual execution of this added line of SQL proved to work on a server.


molgenis-data-migrate/src/main/java/org/molgenis/migrate/version/v1_16/Step26migrateJpaBackend.java
```diff
@@ -53,6 +53,9 @@ public void upgrade()
 		jdbcTemplate.update("INSERT INTO entities_attributes (`order`, `fullName`, `attributes`) VALUES (?, ?, ?)", 22,
 				"MolgenisUser", languageCode);
 
+		jdbcTemplate.update(
```
dennishendriksen: Can you add a comment pointing to the GitHub issue?

dennishendriksen:
test this please

PR #4442, opened 2016-02-05
[poc/view] Disable filtering (tree) and sorting (table) for non-queryable attributes

tommydeboer:


molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/AttributeMetaDataResponseV2.java
```diff
@@ -126,6 +127,8 @@ else if (attr.getDataType() instanceof XrefField || attr.getDataType() instanceo
 			this.attributes = null;
 		}
 
+		this.queryable = queryableAttributes.contains(attr) ? true : false;
```
jjettenn: 1

PR #4455, opened 2016-02-08
Fix #4433 When doing an add / update the importer throws an error

### LINKED ISSUE #4433: When doing an add / update the importer throws an error
## Reproduce
Import the following file: [emx_all_datatypes.xlsx]
(https://github.com/molgenis/molgenis/files/117347/emx_all_datatypes.xlsx)
Or this one: [emx_all_datatypes_csv.zip](https://github.com/molgenis/molgenis/files/117350/emx_all_datatypes_csv.zip)

Then do an add / update with the same file you uploaded

## Expected
Updated entity

## Observed
`null key in entry: null=org.molgenis.data.support.PartialEntity@7aa22ea5`

Mark-de-Haan:


tommydeboer:
Please add a unit test

Mark-de-Haan:
Unit test doesn't add much, should I test if the idAttribute is in the field? Doesn't make much sense

Issue #4336, opened 2016-01-21
InMemoryRepository findAll() includes nulls


tommydeboer:
According to the specs, ```findAll()``` should return an empty stream when no entities were found. 

```java
/**
  * find entities based on a stream of ids.
  * 
  * @return (empty) Stream where the order of entities matches the order of ids, never null
  */
Stream<Entity> findAll(Stream<Object> ids);
```

The ```InMemoryRepository``` adds a null value for each unfound entity:

```java
    @Override
	public Stream<Entity> findAll(Stream<Object> ids)
	{
		return ids.map(id -> entities.get(id));
	}

	@Override
	public Stream<Entity> findAll(Stream<Object> ids, Fetch fetch)
	{
		return ids.map(id -> entities.get(id));
	}
```

These values should be filtered out (for example with ```.filter(Objects::nonNull```).

This might influence tests + code that is tested so beware...

mswertz:
what is the desired behavior?
Because I can imagine you want to know if a particular id resulted in
'null'?

On Thu, Jan 21, 2016 at 3:08 PM, Tommy de boer <notifications@github.com>
wrote:

> According to the specs, findAll() should return an empty stream when no
> entities were found.
>
> /**  * find entities based on a stream of ids.  *   * @return (empty) Stream where the order of entities matches the order of ids, never null  */Stream<Entity> findAll(Stream<Object> ids);
>
> The InMemoryRepository adds a null value for each unfound entity:
>
>     @Override
>     public Stream<Entity> findAll(Stream<Object> ids)
>     {
>         return ids.map(id -> entities.get(id));
>     }
>
>     @Override
>     public Stream<Entity> findAll(Stream<Object> ids, Fetch fetch)
>     {
>         return ids.map(id -> entities.get(id));
>     }
>
> These values should be filtered out (for example with
> .filter(Objects::nonNull).
>
> This might influence tests + code that is tested so beware...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/molgenis/molgenis/issues/4336>.
>



-- 
Dr. Morris A. Swertz
Coordinator BBMRI-NL common services for IT
Coordinator DTL sector health
Coordinator NFU biobank catalogue working group
Head Section Systems Genetics & Genomics Coordination Center
Associate Professor - Big data in biomedicine

Dept. of Genetics, CB50
University Medical Center Groningen
P.O. Box 30001
9700 RB  GRONINGEN
The Netherlands

[...]

PR #4474, opened 2016-02-10
Fix #4473 Repository validator doesn't validate ref attribute value types

### LINKED ISSUE #4473: Repository validator does not validate ref attribute value types
Reproduce:
- As a developer add or update an entity with 1) a ref attribute where the value is a entity of the wrong type or 2) a mref attribute where the value is a iterable containing null values

Expected:
- Validation error is thrown by the validation decorator

Actual:
- Depends on the entity implementation, silently returns null for some implementations

dennishendriksen:


molgenis-data-validation/src/main/java/org/molgenis/data/validation/EntityAttributesValidator.java
```diff
@@ -97,6 +114,58 @@
...
+		if (refEntities == null)
+		{
+			return createConstraintViolation(entity, attr, meta, "Not a valid entity, expected an entity list.");
+		}
+		for (Entity refEntity : refEntities)
+		{
+			if (refEntity == null)
+			{
+				return createConstraintViolation(entity, attr, meta, "Not a valid entity, null is not allowed");
+			}
```
tommydeboer: https://github.com/molgenis/molgenis/issues/4336#issuecomment-178492639 :
> The behavior for requesting a single entity by id is returning null. The behavior of requesting multiple entities by id is thus consistent and I propose to a) keep it that way b) revert the ElasticSearchRepository.findAll() fix c) mention this behavior in the DataService and Repository docs.

Didn't we decide that null *is* allowed/required when requesting multiple entities?

tommydeboer:
test this please

tommydeboer:
test this please

Mark-de-Haan:
test this please

PR #4508, opened 2016-02-16
Fix #3895 SnpEff annotator fails for missing CHROM or POS attribute

### LINKED ISSUE #3895: SnpEff annotator fails for missing CHROM or POS attribute.
log states:
java.lang.ArrayIndexOutOfBoundsException: 1 at org.molgenis.data.annotation.entity.impl.SnpEffAnnotator$SnpEffRepositoryAnnotator.parseOutputLineToEntity

Trying to reproduce locally, will add additional information when it becomes available.

Mark-de-Haan:
When SnpEff is unable to process a line, it will produce an empty line in the output VCF which will cause the annotator to throw a null pointer.

Not so much a fix, but better error handling.


dennishendriksen:
Add a unit test that tests for the NPE

bartcharbon:
this will probably mismatch due to the "SnpEff" assumes order bug.
so not a fix I think.

PR #4509, opened 2016-02-16
[1.17] Fix #4473 Repository validator does not validate ref attribute value + fix integration tests (cherrypicked from master)

### LINKED ISSUE #4473: Repository validator does not validate ref attribute value types

bartcharbon:


bartcharbon:
test this please

PR #4519, opened 2016-02-17
Fix #3855 SnpEff annotator assumes resultfile is same order and size as input

### LINKED ISSUE #3855: SnpEff annotator assumes resultfile is same order and size as input
When using the SnpEff annotator the resultfile is parsed line by line while iterating over the source repository to be annotated, assuming that the order and size of the file SnpEff returns is the same as the repository to be annotated.

This can potentially lead to mismatched data! Result from SnpEff should result in a repository or resource that can than be queried for the correct CHROM/POS/REF/ALT for the entity to be annotated.

Mark-de-Haan:


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotator.java
```diff
@@ -196,29 +199,36 @@ public AnnotatorInfo getInfo()
...
 						if (!next)
 						{
-							IOUtils.closeQuietly(reader);
+							try
+							{
+								repo.close();
+							}
+							catch (IOException e)
+							{
+								e.printStackTrace();
```
bartcharbon: throw runtimeexception oid?

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotator.java
```diff
@@ -294,36 +333,12 @@ public File getInputVcfTempFile(Iterable<Entity> source) throws IOException
...
-			{
-				if (ann_field[2].startsWith("LOF="))
-				{
-					lof = ann_field[2];
-				}
-				else if (ann_field[2].startsWith("NMD="))
-				{
-					nmd = ann_field[2];
-				}
-			}
```
bartcharbon: please put this back...

Mark-de-Haan: Now replaced with entity.getString(LOF) and entity.getString(NMD)

bartcharbon:
parseOutputLineToEntity does is not a proper function  name anymore, it's more like "parseSnpEffEntity to resultEntity" or something like that.
Also: let's think this through with taking the "multiple result" story into account

bartcharbon:
in retrospect we shouldn't have started on this bug, lets park it until we start story 3766

fdlk:
test this please

fdlk:
@bartcharbon so we close this for now?

PR #4520, opened 2016-02-17
Fix #4076 Importing an ontology that already exists gives unclear error

### LINKED ISSUE #4076: Importing an ontology that already exists gives unclear error
### To reproduce:
1. Import [an ontology](https://drive.google.com/drive/folders/0B7qHi41THTs_T3BnaHY2TGJ1Mzg)
2. Import the same ontology again

### Expected:
A message explaining why this ontology can't be imported/updated (for example because it already exists).

### Actual:
image

Mark-de-Haan:


fdlk:
test this please

PR #4533, opened 2016-02-18
M4016 Job plugin + Tape unit tests for Job plugin React components

Mark-de-Haan:


molgenis-core-ui/src/main/javascript/modules/react-components/jobs/JobContainer.js
```diff
@@ -0,0 +1,56 @@
...
+	render: function() {
+		if(this.state.jobs === null) {
+			return <Spinner  />;
+		} else {
+			return <div><Jobs jobs={this.state.jobs}/></div>
+		}
+	},
+	_retrieveJobs: function() {
+		var self = this;
+		setInterval(function() {
```
fdlk: Better to store the result and call clearInterval in componentWillUnmount.
See also https://facebook.github.io/react/docs/reusable-components.html

molgenis-core-ui/src/main/javascript/modules/react-components/jobs/JobTable.js
```diff
@@ -0,0 +1,62 @@
...
+
+var JobTable = React.createClass({
+	mixins: [DeepPureRenderMixin],
+	displayName: 'JobTable',
+	propTypes: {
+		jobs: React.PropTypes.array.isRequired
+	},
+	render: function() {
+		const {jobs} = this.props;
+		return <div>
```
fdlk: Looks like this div doesn't do much anymore.

molgenis-core-ui/src/main/javascript/modules/react-components/jobs/Jobs.js
```diff
@@ -0,0 +1,43 @@
...
+	displayName: 'Jobs',
+	propTypes : {
+		jobs : React.PropTypes.array.isRequired
+	},
+	render: function() {
+		const {jobs} = this.props;	
+		const runningJobs = jobs.filter(job => job.status === 'RUNNING');
+		const finishedJobs = jobs.filter(job => job.status !== 'RUNNING');
+		
+		return <div>
```
fdlk: Same here.

molgenis-core-ui/src/test/javascript/JobComponentsTest.js
```diff
@@ -0,0 +1,124 @@
...
+	startDate: "2100-01-01T00:15:00+0100",
+	endDate: "2100-01-01T00:30:00+0100",
+};
+
+const job_running = {
+	identifier: "test_2",
+	status: "RUNNING",
+	type: "TEST job",
+	entityName: "TapeTest",
+	progessMax: 1,
```
fdlk: Make this a bit higher so the test tests a bit more

molgenis-data/src/main/java/org/molgenis/data/jobs/Job.java
```diff
@@ -0,0 +1,182 @@
+package org.molgenis.data.jobs;
```
bartcharbon: What is this class... looks like JobMetadata rather than a job

molgenis-core-ui/src/main/java/org/molgenis/ui/jobs/JobsController.java
```diff
@@ -0,0 +1,108 @@
...
+
+		if (jobs.size() > MAX_JOBS_TO_RETURN)
+		{
+			return jobs.subList(0, MAX_JOBS_TO_RETURN);
+		}
+
+		return jobs;
+	}
+
+	private Job toJob(Entity entity)
```
bartcharbon: I fail to see the point, Job seems a copy of the JobMetaData. And it is not a job, but once again metadata...

fdlk: I was a bit surprised as well, but I think it's simply a DTO to send the data from the controller to the browser in a json message.

molgenis-data/src/main/java/org/molgenis/data/jobs/Job.java
fdlk: Owh! Now I see where the "old" Job entity went, it got renamed to JobMetadata? And now the DTO is once again called Job?

fdlk:
@dennishendriksen can you take a look at the EntitySerializer? We thought it a bit silly to create a POJO just to return a JobMetaData entity from the controller.

fdlk:
Closing for a new one.

PR #4549, opened 2016-02-25
Fix #4548: Cannot serialize entities using gson

### LINKED ISSUE #4548: Cannot serialize entities using gson.
If a controller wants to serve data from an entity as json, it needs to explicitly create a POJO with fields for the entity names. This is because gson by default only serializes fields and Entities have dynamic metadata.

Often, you can get away with using the REST API, but in those cases where you'd like to return a specific Entity, it'd be real nice if you can simply do so.

fdlk:


molgenis-data/src/main/java/org/molgenis/util/EntitySerializer.java
```diff
@@ -0,0 +1,81 @@
...
+				switch (attr.getDataType().getEnumType())
+				{
+					case BOOL:
+						result.addProperty(attributeName, entity.getBoolean(attributeName));
+						break;
+					case CATEGORICAL:
+					case XREF:
+					case FILE:
+						// TODO: or just refuse it for now?
+						Entity refEntity = entity.getEntity(attributeName);
```
dennishendriksen: What happens for self-referencing entities?

fdlk: *grin* interesting things. Will take a look.

Issue #4424, opened 2016-02-02
Can't set a reference to EntityMDMD or AttributeMDMD


tommydeboer:
### To reproduce
- Make a new ```MetaData``` class with a reference to ```EntityMetaDataMetaData``` or ```AttributeMetaDataMetaData```
```java
public class NewMetaData extends DefaultEntityMetaData
{

        ...

	public NewMetaData()
	{
		super(ENTITY_NAME);

		addAttribute(IDENTIFIER).setIdAttribute(true).setNillable(false).setDataType(STRING);
		addAttribute(REFERENCE).setNillable(false).setDataType(MolgenisFieldTypes.XREF)
				.setRefEntity(EntityMetaDataMetaData.INSTANCE);
	}
}
```
- Start the app

### Expected
The app starts up normally and the references work.

### Actual
```java
org.molgenis.data.MolgenisDataException: Could not resolve dependencies of entities [NewMetaData] are there circular dependencies?
	at org.molgenis.util.DependencyResolver.resolve(DependencyResolver.java:103)
	at org.molgenis.data.meta.MetaDataServiceImpl.onApplicationEvent(MetaDataServiceImpl.java:464)
	at org.molgenis.data.meta.MetaDataServiceImpl.onApplicationEvent(MetaDataServiceImpl.java:1)
```

tommydeboer:
Fixed ```DependencyResolver.resolve()``` by removing dependencies on ```entities``` and ```attributes``` (we know these entities will be available anyhow). However, at second startup, the app crashes. The ```entityMetaDataCache``` in ```EntityMetaDataRepository``` does not include ```entities``` and ```attributes``` so any references to those entities aren't added. This is tricky to fix, so I'll put this one on hold.

dennishendriksen:
Fixed in 2.0.0-SNAPSHOT, see for example FileIngestMetaData:
```
addAttribute(ENTITY_META_DATA).setDataType(XREF).setRefEntity(entityTypeMetadata).setLabel("Target EntityType").setNillable(false);
```

PR #4552, opened 2016-02-25
M4036: As RD-connect and BBMRI-ERIC I want to automatic ingest of Excel/CSV file

erwinwinder:
Also fix needed for this issue: github issue #4424 Can't set a reference to EntityMDMD or AttributeMDMD


tommydeboer:
Deleting an automatic import fails:

```
Error! PreparedStatementCallback; SQL [DELETE FROM `FileIngest` WHERE `id` = ?]; Cannot delete or update a parent row: a foreign key constraint fails (`omx`.`fileingestjobmetadata`, CONSTRAINT `fileingestjobmetadata_ibfk_3` FOREIGN KEY (`fileIngest`) REFERENCES `FileIngest` (`id`)); nested exception is java.sql.BatchUpdateException: Cannot delete or update a parent row: a foreign key constraint fails (`omx`.`fileingestjobmetadata`, CONSTRAINT `fileingestjobmetadata_ibfk_3` FOREIGN KEY (`fileIngest`) REFERENCES `FileIngest` (`id`))
```

tommydeboer:
Is it possible to refresh the import job table after you click 'Execute now'  so you can immediately see the new job?

erwinwinder:
You must delete the FileIngest jobs first. This is standard molgenis behaviour, no cascading deletes.

Import table refresh after 'Execute now' can be done but, but when the jobs actually runs is up to the scheduler, so no guarantee that job is already created.

tommydeboer:
Ok I see. Maybe we should add a 'delete all' button in the future so you don't have to manually remove all the jobs (which could be hundreds).

erwinwinder:
You can delete all jobs in the dataexplorer

tommydeboer:
We have the request that you do a separate pull request for the fix for issue #4424 so we can test it separately.

tommydeboer:
>You can delete all jobs in the dataexplorer
Wouldn't that also delete all jobs of other automatic imports?

erwinwinder:
Implemented refresh after 'execute now'

fdlk:
I merged with master, opening pull request from my repo.

PR #4556, opened 2016-02-25
Fix annotation issue from release 1.17

bartcharbon:


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/AnnotationJob.java
```diff
@@ -124,26 +144,52 @@ private void annotate(String username, AnnotationJobMetaData annotationJobMetaDa
...
-		});
+
+		Runnable task = () -> {
+			RunAsSystemProxy.runAsSystem(() -> {
+				dataService.update(AnnotationJobMetaData.ENTITY_NAME, annotationJobMetaData);
+			});
+		};
+		Thread thread = new Thread(task);
+		thread.start();
+		thread.join();
```
fdlk: and here too?

fdlk:
Opening a new one with only the relevant commits.

PR #4557, opened 2016-02-25
Fix format README enumeration for setting the RuntimeProperty

JanTimmermans:
Changes the enumeration for setting the RuntimeProperty on the README page. Aligns the enumeration vertically rather than on the same line.


JanTimmermans:
The changes are hardly feasible when comparing the code.
Note the additional spaces add the end of the enumeration.

PR #4566, opened 2016-02-26
Fix annotation issue from release 1.17

fdlk:


fdlk:
The moment that the first annotator fails, the job entity goes from RUNNING to FAILED and while the annotation continues, the progress bar goes away and an error message is shown.

fdlk:
Nevermind.

PR #4570, opened 2016-02-26
Fix #4569 Repository validation fails for self-referencing entities

### LINKED ISSUE #4569: Repository validation fails for self-referencing entities
Reproduce:
- Add entities of a entity class with an attribute x with a reference to the entity of the same type:
  - First entity: entity0
  - Second entity: attribute x has value entity0

Expected:
- Validation successful

Actual:
- Validation fails because reference entity0 does not exist

dennishendriksen:


bartcharbon:
test this please

bartcharbon:
test this please

PR #4580, opened 2016-02-29
Fix #4510 - If I copy my entity when annotating and the annotation fails, the created entity is not removed

### LINKED ISSUE #4510: If I copy my entity when annotating and the annotation fails, the created entity is not removed
## Reproduce
- Import the following file:

- Annotation fails

## Expected

## Observed

jjettenn:


molgenis-core-ui/src/main/resources/js/jquery.molgenis.tree.js
```diff
@@ -67,8 +67,8 @@
 		items.push('<div class="row">');
 		items.push('<div class="col-md-12">');
 		items.push('<div class="molgenis-tree-controls">');
-		items.push('<a href="#" class="btn btn-link pull-right tree-deselect-all-btn">Deselect all</a>');
-		items.push('<a href="#" class="btn btn-link pull-right tree-select-all-btn">Select all</a>');
+		items.push('<a href="#" id="tree-deselect-all-btn" class="btn btn-link pull-right">Deselect all</a>');
```
fdlk: But what if more than one tree in same page?

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotator.java
```diff
@@ -296,60 +296,63 @@ public File getInputVcfTempFile(Iterable<Entity> source) throws IOException
...
 		public void parseOutputLineToEntity(String line, Entity entity)
 		{
-			String lof = "";
-			String nmd = "";
-			String[] fields = line.split("\t");
-			String[] ann_field = fields[7].split(";");
-			String[] annotation = ann_field[0].split(Pattern.quote("|"), -1);
-
-			if (ann_field.length > 1)
+			if (line != null)
```
fdlk: Is this needed for this PR in specific? Existing bug?

jjettenn:
Yes this is an existing NullPointerException bug

molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
```diff
@@ -558,6 +567,36 @@ function($, molgenis, settingsXhr) {
 			});
 		});
 		
+		$('#copy-data-btn').on('click', function(){
+			bootbox.prompt({
+				title: "<h4>Duplicate entity [" + selectedEntityMetaData.label + "]<h4/>" +
```
fdlk: Perhaps "Copy entity" is clearer, since "Duplicate entity" could also mean that the entity name you've entered is already taken.

molgenis-data/src/main/java/org/molgenis/data/support/DataServiceImpl.java
```diff
@@ -359,6 +360,13 @@ public Repository copyRepository(Repository repository, String newRepositoryId,
 	}
 
 	@Override
+	@RunAsSystem
+	public Repository copyRepositoryRunAsSystem(Repository repository, String newRepositoryId, String newRepositoryLabel)
```
fdlk: I'd hope we can put this on a bit higher level, say in the rest controller?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -465,6 +534,32 @@ public ErrorMessageResponse handleValidationException(MolgenisValidationExceptio
...
+	@ResponseBody
+	public ErrorMessageResponse handleDuplicateEntityException(DuplicateEntityException e)
+	{
+		LOG.info("Duplicate entity exception occurred.", e);
+		return new ErrorMessageResponse(new ErrorMessage(e.getMessage()));
+	}
+
+	@ResponseStatus(BAD_REQUEST)
+	@ResponseBody
+	public ErrorMessageResponse handleUnknownEntityException(UnknownEntityException e)
```
fdlk: The handlers are very similar, can we reuse the same handler for all MolgenisDataExceptions?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -299,7 +323,52 @@ public EntityCollectionBatchCreateResponseBodyV2 createEntities(@PathVariable("e
...
+		boolean readPermission = permissionService.hasPermissionOnEntity(repositoryToCopy.getName(), Permission.READ);
+		if (!readPermission) throw createNoReadPermissionOnEntityException(entityName);
+
+		// Capabilities
+		boolean writableCapabilities = dataService.getCapabilities(repositoryToCopy.getName()).contains(
+				RepositoryCapability.WRITABLE);
+		if (!writableCapabilities) throw createNoWriteCapabilitiesOnEntityException(entityName);
+
+		// Copy
+		Repository repository = dataService.copyRepositoryRunAsSystem(repositoryToCopy, request.getNewEntityName(),
```
fdlk: If you put the copyRunAsSystem in a separate method here that runs as system, wouldn't that be a bit cleaner?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -91,17 +99,32 @@ static UnknownEntityException createUnknownEntityException(String entityName)
...
+	{
+		return new MolgenisDataAccessException("No read permission on entity " + entityName);
+	}
+
+	static MolgenisDataException createNoWriteCapabilitiesOnEntityException(String entityName)
+	{
+		return new MolgenisRepositoryCapabilitiesException("No write capabilities for entity " + entityName);
+	}
+
+	static DuplicateEntityException createDuplicateEntityException(String entityName)
```
fdlk: Inline these three, I think.

molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
```diff
@@ -182,9 +183,20 @@ else if (moduleId.equals("annotators"))
...
 		return "view-dataexplorer-mod-" + moduleId; // TODO bad request in case of invalid module id
 	}
 
+	@RequestMapping(value = "/copy", method = GET)
+	@ResponseBody
+	public boolean showCopy(@RequestParam("entity") String entityName)
+	{
+		boolean showCopy = false;
+		showCopy = molgenisPermissionService.hasPermissionOnEntity(entityName, Permission.READ)
+				&& dataService.getCapabilities(entityName).contains(RepositoryCapability.WRITABLE);
```
fdlk: As long as the default backend supports the creation of new entities, a copy of a readonly entity should work fine, no?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
jjettenn: I have thought the same, but it will break other functionality.
Sometimes the application throws other MolgenisDataException exceptions. I don't want to manage all of them

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotator.java
fdlk: Can you file that bug separately then and report it fixed in this PR?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
fdlk: How does the annotation work, if you put three annotations with different exception classes on the same method, will it handle those three (and only those three)?
So I mean

```
@ExceptionHandler(DuplicateEntityException.class)
@ExceptionHandler(MolgenisRepositoryCapabilitiesException.class)
@ExceptionHandler(UnknownEntityException.class)
handleException(MolgenisDataException ex)
```

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -299,7 +323,52 @@ public EntityCollectionBatchCreateResponseBodyV2 createEntities(@PathVariable("e
...
+	 * @param entityName
+	 *            name of the entity that will be copied.
+	 * @param request
+	 *            CopyEntityRequestV2
+	 * @param response
+	 *            HttpServletResponse
+	 * @return String name of the new entity
+	 * @throws Exception
+	 */
+	@RequestMapping(value = "copy/{entityName}", method = POST, produces = APPLICATION_JSON_VALUE)
```
fdlk: You're creating a new resource when you copy a repository, so a PUT(or POST?) to the location where that resource is going to reside would be a bit more RESTful perhaps?

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/SnpEffAnnotator.java
jjettenn: Issue: SnpEffAnnotator.java NullPointerException if no results are found. #4590

molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
jjettenn: I don't want to show the button when user has count permission.

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
jjettenn: Can you please explain more about "Inline these three, I think."?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
jjettenn: This is not possible, you can use only one ExceptionHandler per methode. I will find a solution

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
fdlk: I mean to refactor them using "inline method" refactoring. That is, to call the constructor directly where you now call the method that calls the constructor for you.

molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
fdlk: Yes, but I mean when the repository isn't writable and the user has count permission, I think he should be able to copy it to a new repo that's created in the default repositorycollection. No?

fdlk: That said, shouldn't the user be able to specify the repositorycollection he'd like to copy to?

molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
```diff
@@ -182,9 +183,20 @@ else if (moduleId.equals("annotators"))
 			model.addAttribute("annotationRun", annotationRun);
 			model.addAttribute("entityName", entityName);
 		}
+
 		return "view-dataexplorer-mod-" + moduleId; // TODO bad request in case of invalid module id
 	}
 
+	@RequestMapping(value = "/copy", method = GET)
+	@ResponseBody
+	public boolean showCopy(@RequestParam("entity") String entityName)
```
fdlk: This exact information is already available in the REST api under /api/v2/entityName. If you have no right to read items, you get a not-authorized status code and the response shows in meta.writable if the repository is writable.

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
jjettenn: I am working on it

jjettenn:
test this please

PR #4582, opened 2016-03-01
M4016: For long running processes (annotators, importers, downloads, mapping service) we have a progress bar that can run in the background

fdlk:
To a separate branch for now, so as not to hinder the merging of the csv slurper.


fdlk:
Looks like I broke the unit tests when I added the log to the detail view. Brace does not work in node.js. Taking a look.

molgenis-core-ui/package.json
```diff
@@ -33,10 +33,13 @@
     "jquery-validation": "1.14.0",
     "jquery.cookie": "^1.4.1",
     "jquery.fancytree": "2.14.0",
-    "moment": "2.9.0",
+    "moment": "2.11.2",
```
fdlk: This is a bit of an issue, until we merge this to master they'll require different versions of moment and that's awkward cause the pull request builder runs them in the same node_modules install.

fdlk: Fixed in #4584

bartcharbon:
I have a failing unittest in the annotators:
the EdgeCaseExacCmdLineAnnotatorIT says:
Could not autowire field: private org.springframework.transaction.PlatformTransactionManager org.molgenis.data.annotation.AnnotationJobFactory.transactionManager

bartcharbon:
very minor UI remark: "Show details","Go to result" in the job overview screen are so close together it seems like one sentence/link.

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/AnnotationJobFactory.java
```diff
@@ -0,0 +1,79 @@
+package org.molgenis.data.annotation;
```
bartcharbon: this class could use a "optimize imports"

fdlk:
Good question about the command line test. In the command line version the whole job thing isn't relevant, so the JobFactory shouldn't get instantiated. It probably doesn't belong in the annotator package to begin with.

bartcharbon:
In the "stack of stickies" on the scrumboard there is also the task to "move the message to underneath the progress bar" this is not done yet. 
I can fix this in a separate PullRequest.

fdlk:
Closing for a PR to the master.

PR #4585, opened 2016-03-01
M4036: As RD-connect and BBMRI-ERIC I want to automatic ingest of Excel/CSV file

fdlk:


fdlk:
I've taken a look at the entity delete requested by @tommydeboer but that's all rather awkward. We're using the standard table and it's not possible to override the delete button behavior for the table with the job definitions. It has a mind of its own.

fdlk:
Tested the csv ingestion and it works fine.

fdlk:
So that leaves the comments about the xref to entitymetadata. There, too, we cannot have it both ways. If this has to be a standard form, the target entity has to be a standard xref. So either we customize the form and the controller or we standardize the xref and the metadata.

PR #4592, opened 2016-03-03
M4016: For long running processes (annotators, importers, downloads, mapping service) we have a progress bar that can run in the background

fdlk:


fdlk:
test this please

PR #4601, opened 2016-03-07
Fix #4600 Cannot import emx_all_datatypes.xlsx

### LINKED ISSUE #4600: Cannot import emx_all_datatypes.xlsx
* import emx_all_datatypes.xlsx
* Error message:
Failed to convert from type java.lang.String to type java.sql.Date for value '2014-08-01T00:00:00+0200'; nested exception is java.lang.IllegalArgumentException

Mark-de-Haan:
Changed the default value for xdatenillable and xdate from 01/08/2014 and 01/04/2015 to 2014-08-01 and 2015-04-01. Previous format caused exceptions with java.sql.date


bartcharbon:
please explain what has been changed...

PR #4621, opened 2016-03-08
Merge bug fixes from 1.18 to master branch

Mark-de-Haan:


dennishendriksen:
Maybe leave out a few for the case where no label or description was provided and the name should not be displayed?

PR #4627, opened 2016-03-08
Fix #4510 copy bug 

### LINKED ISSUE #4510: If I copy my entity when annotating and the annotation fails, the created entity is not removed

jjettenn:
Remove RunAsSystem annotation and call the proxy RunAsSystemProxy.
Add try to delete the copy if somthing went wrong.
Add test


jjettenn:
test me please

PR #4632, opened 2016-03-08
Fix #4510 If I copy my entity when annotating and the annotation fails, the created entity is not removed

### LINKED ISSUE #4510: If I copy my entity when annotating and the annotation fails, the created entity is not removed

jjettenn:


jjettenn:
test me please

tommydeboer:
It is not possible to copy entities that are in the base package. The error is ```Error! Unknown entity [base_Test]``` but entities in the base package are not prefixed with that package name.

jjettenn:
I have just copied an entity with package base (AnnotationJobExecution). Can you be more specific?

tommydeboer:
I tried multiple entities including ```AnnotationJobExecution``` but I still get the error. I both tried checking out this branch (*fix/4510sprint82*) and by checking out the master and pulling from this branch. I did a clean install with dropped database and index.

PR #4660, opened 2016-03-09
Fix #4624: importByUrl returns content-type application/json but is raw text

### LINKED ISSUE #4624: importByUrl returns content-type application/json but is raw text
* POST /plugin/importwizard/importByUrl?notify=false&entityName=demo&url=https://raw.githubusercontent.com/bartcharbon/molgenis/feature/importService/molgenis-app/src/test/resources/vcf/test.vcf
* Returns /api/v2/ImportRun/AAAACU2V6GVNOATAPVHJR3AAAE", Content-Type application/json
  This is not json, so change response body or change ContentType to text/plain

fdlk:


molgenis-data-import/src/test/java/org/molgenis/data/importer/ImportWizardControllerTest.java
```diff
@@ -26,6 +53,7 @@
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
```
erwinwinder: Please remove unusedimport org.springframework.http.MediaType;

PR #4661, opened 2016-03-09
Fix #4626: when there are failing annotators the job status is SUCCESS

### LINKED ISSUE #4626: when there are failing annotators the job status is SUCCESS
this line in the AnnotationJobExecution should not be possible:
SUCCESS	annotators	2016-03-08T10:41:19+0100	2016-03-08T10:41:19+0100	2016-03-08T10:41:19+0100	8	8	Failed annotators: cadd,dann,clinvar,gonl,exac,fitcon,snpEff,thousand_genomes

fdlk:


bartcharbon:
Also fixed this since you unassigned ;) Your fix is similar but slightly cleaner, and your unit test is way better. So I will merge it anyway :)

PR #4664, opened 2016-03-09
Fix # 4663 Inconsistent check for length password

JanTimmermans:
Replace the hard coded minimal password length with the value of min_password_length
(which is already defined in molgenis-security/src/main/java/org/molgenis/security/user/UserAccountService.java )


tommydeboer:
Jenkins, test this please

fdlk:
add to whitelist

tommydeboer:
Thanks for your help Jan, we appreciate it!

PR #4674, opened 2016-03-11
Fix #4672 SQL error when using mref name that results in the same name as the referenced entity

### LINKED ISSUE #4672: SQL error when using mref name that results in the same name as the referenced entity
#### Reproduce
1. Make an entity _Test_
2. Make an entity _Test_Ref_
3. Give _Test_ an mref called _Ref_
4. Import
#### Expected:

Works.

#### Actual:

```
org.springframework.jdbc.BadSqlGrammarException: StatementCallback; bad SQL grammar [ 
CREATE TABLE `Test_Ref`(`order` INT,`identifier` VARCHAR(255) NOT NULL, `EFFECT`
VARCHAR(255) NOT NULL, FOREIGN KEY (`identifier`) REFERENCES `Test`(`identifier`) ON 
DELETE CASCADE, FOREIGN KEY (`EFFECT`) REFERENCES `Test_Ref`(`id`) ON 
DELETE CASCADE) ENGINE=InnoDB;]; nested exception is
com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: Table 'Ref_Test' already 
exists
```

tommydeboer:


molgenis-data/src/main/java/org/molgenis/data/meta/MetaValidationUtils.java
```diff
@@ -108,6 +110,15 @@ protected static void validateAttribute(AttributeMetaData amd)
...
 		}
+		if (amd.getDataType().equals(MREF))
+		{
+			if ((emd.getName() + "_" + amd.getName()).equals(amd.getRefEntity().getName()))
+			{
+				throw new MolgenisDataException("The name of mref attribute [" + amd.getName()
+						+ "] is invalid because its full name will collide with the referenced entity ["
+						+ amd.getRefEntity().getName() + "] ");
+			}
+		}
```
fdlk: But the problem is Mysql backend specific.

PR #4678, opened 2016-03-13
Fix #4598 add *.vcf.zip support

### LINKED ISSUE #4598: Import of ZIP/TAR.GZ compressed VCF is not possible
It is not possible to import a zipped (.vcf.zip) VCF. The error thrown:
```
Error! Your import failed:
Can not import file. No suitable importer found
```
Importing a gzipped TAR (.vcf.tar.gz) does not work either. The error thrown:
```
Error! Your import failed:
Unknown extension for file 'myfile.tar.gz'
```
NOTE: Just gzipping the file does work! the extension becomes ".vcf.gz" in that case.

QiaoN:
This change support the file format .vcf.zip (one file in zip only). Which means a zipped vcf file will be accepted. Before this fix the .vcf.zip file will cause import failed.

### DIFF
--- a/molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/VcfReaderFactoryImpl.java
+++ b/molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/VcfReaderFactoryImpl.java
@@ -9,7 +9,10 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.Charset;
 import java.util.List;
+import java.util.Enumeration;
 import java.util.zip.GZIPInputStream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
 
 import org.molgenis.data.MolgenisDataException;
 import org.molgenis.vcf.VcfReader;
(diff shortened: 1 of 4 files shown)


fdlk:
Add to whitelist

fdlk:
Hi! Is there an issue with the pull request that caused you to close it? Cause quick code inspection looked good to me.

fdlk:
add to whitelist

QiaoN:
@fdlk Sorry, I accidentally close the pull request. It is fine!

molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/VcfReaderFactoryImpl.java
```diff
@@ -41,6 +44,13 @@ public VcfReader get()
 			if (file.getName().endsWith(".gz"))
 			{
 				inputStream = new GZIPInputStream(inputStream);
+			} 
+			else if (file.getName().endsWith(".zip"))
+			{
+				   ZipFile zipFile = new ZipFile(file.getPath());
+				   Enumeration<? extends ZipEntry> e = zipFile.entries();
+				   ZipEntry entry = (ZipEntry) e.nextElement(); // your only file
```
fdlk: Some OS-es add system-specific entries to a zipfile when you create them from the finder / explorer.
Could you make it so that it imports the first entry with extension .vcf and explains why it fails if no such entry is found?

QiaoN: No problem to fix this part. Will push the change soon if needed.
But one concert is that the original fix was focus on format "*.vcf.zip" which mean directly zip original vcf file. For the folder contain vcf file then zip, the extension will be "*.zip". That will be conflicted with csv reader implementation as importWizard can not call the right reader.
Please let me know if I still need to push the change or not.

fdlk: @QiaoN Ah yes, I had indeed been thinking of zipping a folder.

PR #4679, opened 2016-03-14
Fix #4123 Failed login prevents the login dialog from being closed

### LINKED ISSUE #4123: Failed login prevents the login dialog from being closed
The close button disappears if the login fails.
I'm told this is a feature for servers where you've no page to fall back to until you log in, but in the normal case this is weird behavior.

JanTimmermans:
Solves issue #[4123](https://github.com/molgenis/molgenis/issues/4123)
The old code removed the close button from the login-modal after a failed login attempt.
This prevents the user from being directed to /login?error (a blank fall back page).

The fix introduces a new button to (/) when the login fails (equal formatting).

The use of an alternative implementation, window.history.back(), could still cause problems when the user has multiple login failures. Because, the user will be redirected to the login-modal in that case.


fdlk:
Thanks, looks great!

PR #4684, opened 2016-03-15
Fix #4413 Menu overlaps top of page content

### LINKED ISSUE #4413: Menu overlaps top of page content
If the screen is narrower than the menu, but wider than the threshold for the hamburger button to appear, it becomes two (possibly more?) lines high.
But the page content doesn't scroll down, and gets hidden by the second line of the menu.
The top of the page becomes invisible and unreachable.

JanTimmermans:
Added some lines to molgenis.css.
The current threshold of the menu to collapse is 768px.
With this addition the user can change this threshold by altering one value (@media (max-width: 768px)) of Molgenis.css.

### DIFF
--- a/molgenis-core-ui/src/main/resources/css/molgenis.css
+++ b/molgenis-core-ui/src/main/resources/css/molgenis.css
@@ -177,7 +177,7 @@ body {
 }
 
 /* Change threshold for collapsing the menu, default is 768px*/
-@media (max-width: 768px) {
+@media (max-width: 1127px) {
     .navbar-toggle {
         display: block;
     }
--- a/molgenis-core-ui/src/main/resources/css/molgenis.css
+++ b/molgenis-core-ui/src/main/resources/css/molgenis.css
@@ -175,3 +175,28 @@ body {
 	padding: 7px 0px;
 	height: 26px;
 }
+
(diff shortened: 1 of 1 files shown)


molgenis-core-ui/src/main/resources/css/molgenis.css
```diff
@@ -174,4 +174,29 @@ body {
 	width: 125px;
 	padding: 7px 0px;
 	height: 26px;
-}
\ No newline at end of file
+}
+
+/* Change threshold for collapsing the menu, default is 768px*/
+@media (max-width: 768px) {
```
fdlk: Let's make this 1127, that should fix the issue for the standard case, even if the language pulldown is shown.

fdlk:
I've taken a look at http://stackoverflow.com/questions/20817956/change-collapse-breakpoint-in-bootstrap-3-0 where they suggest the slightly more advanced solution, i.e. to change the less variable `@grid-float-breakpoint`. But this looks like a simpler and better solution to me since the rest of our css is default. So please update the threshold, and I'll merge.

JanTimmermans:
@fdlk Thanks for your feedback. 
I've updated the threshold to 1127px.

fdlk:
Thanks for the bugfix!

Issue #4297, opened 2016-01-05
nonobvious ref and alt not working for GoNL


bartcharbon:
this line in the resource
1	231094050	.	GA	G	.	PASS	AC=15;AN=991;DB;GTC=7,11,33;set=SNP

should match on this one in the entity(both have a deletion of a single "A"):
1	G,GAAA,GA	231094050	GAA		85.84		

However no results are added for the GoNL annotator

bartcharbon:
GoNL annotator does not use the multiallelicResultFilter (where the nonobvious matching was implemented), so the solution needs to be added to GoNL still?

Issue #4098, opened 2015-10-29
GoNL annotator writes out INFO field with '.' when no results are found


joerivandervelde:
When no results are found within GoNL, the INFO fields should remain untouched. However, 'null' strings are written out as such:

```
AC=1;AF=0.5;AN=2;set=variant38;GoNL_GTC=.,.,.;GoNL_AF=.;
```
This should be:

```
AC=1;AF=0.5;AN=2;set=variant38;
```

bartcharbon:
Strange that this is annotator specific since the GoNL annotator does not do the writing itself, it just fills out columns like every other annotator. 
Does it also add a dot or space in the UI version to the entity?

Issue #3674, opened 2015-09-15
GoNL annotator annotates with a random match from the GoNL source


fdlk:
GoNL can have multiple rows for the same chrom/pos/ref combination, e.g.
```
1	80301888	rs11591201	G	A	.	PASS	AC=260;AN=996;DB;GTC=267,202,29;set=SNP
1	80301888	.	G	GT	.	PASS	AC=1;AN=996;GTC=497,1,0;set=INDEL
```
It matches these rows with the ALT column of the input entity.

Currently, if the input entity has ALT = A,GT or GT,A then the result value for GTC and AF is that of the last matching GoNL row. It should be a comma separated list of all matching rows.

PR #4687, opened 2016-03-16
Fix #4586 GoNL annotator writes non-standard separators and missing values

### LINKED ISSUE #4586: GoNL annotator writes non-standard separators and missing values
The GoNL annotator uses pipe ("|") to separate matches for multiple alleles, and prints no value when there a value is missing (""). Other allele frequency annotators use comma (",") to separate multiple alleles and print a full-stop (".") to denote there was no match for this particular allele. There is no reason for the GoNL annotator to be inconsistent, and it makes parsing/using the MAF information needlessly complex.

Example:

```
18	11753801	.	AT	GT,A	GoNL_AF=0.024096385542168676|;EXAC_AF=0.013,0.001244;Thousand_Genomes_AF=0.00559105,.;
```

Should become:

```
18	11753801	.	AT	GT,A	GoNL_AF=0.024096385542168676,.;EXAC_AF=0.013,0.001244;Thousand_Genomes_AF=0.00559105,.;
```

joerivandervelde:
Fix for issue "GoNL annotator writes non-standard separators and missing values". Updated writing of missing values, allele separators and genotype counts to be the same as other population based annotators.

### DIFF
--- a/molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/GoNLAnnotator.java
+++ b/molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/GoNLAnnotator.java
@@ -154,12 +154,12 @@ public class GoNLAnnotator
 					if (!Iterables.all(alleleMatches, Predicates.isNull()))
 					{
 						afs = alleleMatches.stream()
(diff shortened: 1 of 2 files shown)


fdlk:
So after #3674, #4098, #4297 looks like we finally are getting it right!

PR #4695, opened 2016-03-17
Get like and not like query working for rsql queries

marikaris:


molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
```diff
@@ -32,7 +32,6 @@
     <#if app_settings.cssHref?has_content>
         <link rel="stylesheet" href="<@resource_href "/css/${app_settings.cssHref?html}"/>" type="text/css">
     </#if> 
-
```
fdlk: Unrelated, I think?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestConfig.java
```diff
@@ -55,6 +55,8 @@ public RSQLParser rsqlParser()
 	{
 		Set<ComparisonOperator> operators = RSQLOperators.defaultOperators();
 		operators.add(new ComparisonOperator("=q=", false));
+		operators.add(new ComparisonOperator("=like=", false));
```
fdlk: Like is already listed

fdlk:
add to whitelist

Issue #4693, opened 2016-03-17
CADD annotator is not multi-allelic


joerivandervelde:
When annotating a multi-allelic VCF line, you only get back 1 CADD result, not knowing to which alternative allele it belongs. For example:
```
22	25599848	.	C	T,TA	.	.	.
```
Gets annotated as
```
22	25599848	.	C	T,TA	.	.	CADD=6.956883;CADD_SCALED=33.0;
```
But should be
```
22	25599848	.	C	T,TA	.	.	CADD=6.956883,.;CADD_SCALED=33.0,.;
```

Mark-de-Haan:
Fixed in #4696

PR #4696, opened 2016-03-18
Fix for #4693: CADD annotator is not multi-allelic

joerivandervelde:
Added merging of multiline resource output so that CADD (and other such annotators) can be multi allelic. Added extra constructor to explictly enable this behaviour. Updated tests and test files. Included one other small fix where a null output represented by '.' now also extends to any combination of '.,', as is the case for no hits for multi allelic annotation.


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/filter/MultiAllelicResultFilter.java
```diff
@@ -54,10 +55,18 @@
 {
 
 	private List<AttributeMetaData> attributes;
-
-	public MultiAllelicResultFilter(List<AttributeMetaData> attributes)
+	boolean mergeMultilineResourceResults;
```
Mark-de-Haan: make this private

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/filter/MultiAllelicResultFilter.java
```diff
@@ -118,12 +133,110 @@ private void filter(Entity annotatedEntity, List<Entity> processedResults, Entit
...
+	 * and it should become:
+	 * 
+	 * 3	300	G	A,T,X,C	0.2|23.1,-2.4|0.123,-0.002|2.3,0.5|14.5
+	 * 3	300	GC	A,T	0.2|23.1,-2.4|0.123
+	 * 3	300	C	GX,GC	-0.002|2.3,0.5|14.5
+	 * 
+	 * so that the multi-allelic filter can then find back the appropriate values as if it were a multi-allelic VCF line
+	 * 
+	 */
+	public Iterable<Entity> merge(Iterable<Entity> resourceEntities)
```
Mark-de-Haan: A unit test for this method would be cool. 
Testing this through the entire annotator path is a bit tricky IMO

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/filter/MultiAllelicResultFilter.java
```diff
@@ -118,12 +133,110 @@ private void filter(Entity annotatedEntity, List<Entity> processedResults, Entit
...
+			}
+			myList.add(resourceEntity);
+		}
+		
+		//now iterate over map with refs and merge entities per ref
+		for(String refKey : refToMergedEntity.keySet())
+		{
+			boolean first = true;
+			Entity mergeWithMe = null;
+			for(Entity e : refToMergedEntity.get(refKey))
```
Mark-de-Haan: one letter variables are evil!

PR #4698, opened 2016-03-18
Added row counts to entity select

Mark-de-Haan:
screen shot 2016-03-18 at 14 06 10

### DIFF
--- a/molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
+++ b/molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
@@ -1,7 +1,29 @@
 package org.molgenis.dataexplorer.controller;
 
-import com.google.gson.Gson;
-import freemarker.core.ParseException;
+import static com.google.common.collect.Maps.newHashMap;
+import static org.molgenis.dataexplorer.controller.DataExplorerController.ATTR_GALAXY_API_KEY;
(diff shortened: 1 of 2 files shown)


fdlk:
Does not update the counts when they change unless I reload the page.

Mark-de-Haan:
Dynamically updating our current select2 implementation is not possible, so this static implementation that only reloads on page refresh is the best thing we can do until we implement react-select

molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
```diff
@@ -129,8 +129,16 @@ public String init(@RequestParam(value = "entity", required = false) String sele
...
 		boolean entityExists = false;
 		boolean hasEntityPermission = false;
+
 		List<EntityMetaData> entitiesMeta = dataService.getEntityNames().map(dataService::getEntityMetaData)
 				.collect(Collectors.toList());
+		Map<String, Long> entityCounts = newHashMap();
+		for (EntityMetaData entityMetaData : entitiesMeta)
+		{
+			String entityName = entityMetaData.getName();
+			entityCounts.put(entityName, dataService.count(entityName, new QueryImpl()));
```
bartcharbon: This will probably become a big performance bottleneck quite quickly...

Mark-de-Haan:
We can revisit this feature once we have react-select implemented

PR #4714, opened 2016-03-24
Fix #4713: AUTO id columns overwrite specified id value

### LINKED ISSUE #4713: AUTO id columns overwrite specified id value
If you specify an ID value for an AUTO id column, the value gets overwritten when you call dataservice.add().

I'd expect the AUTO id only to work when you specify NULL as a value.

fdlk:


dennishendriksen:
What is the desired behavior here?
1) Auto id column do not overwrite id value
2) Auto id column do not overwrite id value, exception is thrown
3) Auto id column overwrites id value

I would expect 3) or alternatively 2) since this is probably a user error. Furthermore some validation steps are not executed for auto id columns, since we know that the system generates them.

fdlk:
Sorta job execution entity extends the abstract JobExecution which has auto ID.
BUT for the sorta job execution entity it happens to be pretty handy to be able to specify the ID before adding the entity to the repo.
So there we'd like to generate one of our own and would desire behavior nr 1 please!

dennishendriksen:
In case of 1) AUTO attributes need to be validated.

fdlk:
In particular uniqueness?

fdlk:
I can also change the way I solved it in sorta, but a different use case where users will run into this is when you import entities with references through emx.

dennishendriksen:
The same validation rules as when auto=false

fdlk:
Only special validation that I can find is that null checks aren't executed for auto attributes. Presumably cause they'll be filled in by the autoValue decorator. That behavior remains correct so no changes needed there.

Also found that the streaming version of add() already does 1).

@dennishendriksen can you point out an example of such a special validation rule that I missed?

dennishendriksen:
RepositoryValidationDecorator performs all entity validation, the corresponding test class contains about a hundred test cases that illustrate how entity validation works.

dennishendriksen:
You can not reference to auto ids in EMX, in this case use your own ids.

fdlk:
> Furthermore some validation steps are not executed for auto id columns, since we know that the system generates them.
I did not find a single case of that. Only that auto IDs don't get null checked. Which still is correct.

PR #4730, opened 2016-04-05
Fix #4729 MySqlRepository not QUERYABLE but answers queries

### LINKED ISSUE #4729: MySqlRepository not QUERYABLE but answers queries
### Reproduce
- Look at MysqlRepository.getCapabilities, confirm it is not QUERYABLE
- Put a breakpoint in MysqlRepository.findAll(Query q)

### Expected
- Method is not called from outside MysqlRepository

### Actual
- Method is called from ElasticsearchRepositoryDecorator.findAll(Query q)

dennishendriksen:


tommydeboer:
Deleting books is without problems now. However, when I type in the 'books' field, I get an error:

```
Error! Failed to execute phase [query], all shards failed; shardFailures 
{[JNozibH4SnSC90hK3WB9vw][molgenis][0]: SearchParseException[[molgenis][0]: 
from[0],size[25]: Parse Failure [Failed to parse source [{"from":0,"size":25,"query":
{"bool":{"should":[{"filtered":{"query":{"match_all":{}},"filter":{"term":{"id":"b"}}}},
{"match":"title.ngram":"query":"b","type":"boolean","analyzer":"default"}}}],
"minimum_should_match":"1"}},"sort":[{"id":{"order":"asc","mode":"min"}}]}]]]; nested: 
NumberFormatException[For input string: "b"]; }
```

dennishendriksen:
Which data set are you using? Can you confirm that this issue did not exist with this data set before the fix?

tommydeboer:
I was using the dataset from the issue in github. The issue is unrelated btw, I'll add it to the tracker.

tommydeboer:
Oops, all my comments in this PR should go under #4726 Updating MREF attrs for ref entity int attribute

Issue #4791, opened 2016-04-21
CmdLine Gavin overwrites SnpEff headers


bartcharbon:
## Reproduce
- run commandline SnpEff annotator
- run commandline Gavin annotator

## Expected
"EFFECT" in both header and data is the combination of Gavin and SnpEff

## Observed
"EFFECT" in header is the Gavin only

PR #4794, opened 2016-04-21
fix: #4791 CmdLine Gavin overwrites SnpEff headers AND #4795 CmdLineAnnotator with update=true removes all values if no result was found

bartcharbon:
PR accidentally also contains previous PR, please merge in right order.

### DIFF
--- a/molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/AnnotatorImpl.java
+++ b/molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/impl/AnnotatorImpl.java
@@ -46,7 +46,10 @@ public class AnnotatorImpl extends QueryAnnotatorImpl implements EntityAnnotator
 		{
 			for (AttributeMetaData attr : getInfo().getOutputAttributes())
 			{
-				resultEntity.set(attr.getName(), null);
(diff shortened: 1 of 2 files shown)


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/cmd/CmdLineAnnotator.java
```diff
@@ -306,9 +306,10 @@ else if (inputAttributeNames.contains(attrName))
...
 
-			System.out.println("update = " + options.has("u"));
+			boolean update = options.has("u");
+			System.out.println("update = " + update);
 
-			Iterator<Entity> annotatedRecords = annotator.annotate(entitiesToAnnotate);
+			Iterator<Entity> annotatedRecords = annotator.annotate(entitiesToAnnotate, update);
 
 			if (annotator instanceof RefEntityAnnotator || annotator instanceof EffectsAnnotator)
 			{
```
Mark-de-Haan: Pull from 1.20 branch

bartcharbon: Will do, but should not be needed, github used to understand this

PR #4800, opened 2016-04-21
Add IndexedRepositoryDecorator

tommydeboer:
- A ```Repository``` can now define which query operators it can handle
- ```IndexedRepositoryDecorator``` checks a query and either sends the query to the decorated repository or the index
- All queries that involve computed attributes are sent to the index


jjettenn:
👎 :)

PR #4805, opened 2016-04-26
M4060: As Joeri, I want to have a MOLGENIS app for my variant classification method

fdlk:
Also contains the FileUpload component.

#### Checklist
- [x] Unit tests
- [x] Updated testplan


molgenis-app/src/main/java/org/molgenis/app/WebAppSecurityConfig.java
```diff
@@ -57,14 +49,16 @@ protected void configureUrlAuthorization(
 
 				.antMatchers("/phenotips/**").authenticated()
 
+				.antMatchers("/gavin/**").authenticated()
```
Mark-de-Haan: These changes are not needed anymore because we made gavin a plugin

PR #4813, opened 2016-04-29
Build fixes

joerivandervelde:
Updated gavin molgenis dependency to 1.21.0-SNAPSHOT
Removed Maven repo from molgenis POM


fdlk:
Thanks!

PR #4821, opened 2016-05-03
Added bug description, fix duplicate tests

Mark-de-Haan:


molgenis-data-elasticsearch/src/test/java/org/molgenis/data/elasticsearch/request/SearchRequestGeneratorTest.java
```diff
@@ -81,21 +81,4 @@ public void testBuildSearchRequestNoFetchBackendElasticsearch()
...
-		SearchType searchType = SearchType.COUNT;
-
-		gen.buildSearchRequest(searchRequestBuilderMock, entityName, searchType, new QueryImpl<Entity>().search("test"), null,
-				null, null, entityMeta);
-		verify(searchRequestBuilderMock).setFrom(0);
-		verify(searchRequestBuilderMock).setSearchType(searchType);
-		verify(searchRequestBuilderMock).setTypes(entityName);
-		verify(searchRequestBuilderMock).setQuery(Matchers.<QueryBuilder> anyObject());
-		verifyNoMoreInteractions(searchRequestBuilderMock);
-	}
```
tommydeboer: Instead of removing this test, the test above it should be fixed

molgenis-integration-tests/src/test/java/org/molgenis/integrationtest/data/abstracts/query/AbstractSearchQueryIT.java
```diff
@@ -68,56 +75,68 @@ protected void testDate() throws ParseException
...
 
 	@Override
 	protected void testBool()
 	{
-		// UNSUPPORTED
+		try
+		{
+			String value = "true";
+			Query<Entity> query = new QueryImpl<Entity>().search(ACTIVE, value);
+			assertTrue(newHashSet(person2).contains(personsRepository.findOne(query)));
```
tommydeboer: ```assertTrue``` isn't necessary here because you expect the exception to be thrown when you call ```.findOne()```

Issue #4817, opened 2016-05-03
Numerous query operator integration tests fail


tommydeboer:
Uncomment tests in these classes:
 - [ ] ```AbstractInQueryIT``` (fails for Elasticsearch)
 - [x] ```AbstractLikeQueryIT``` (fails for PostgreSQL)
 - [ ] ```AbstractNotQueryIT``` (fails for PostgreSQL)
 - [ ] ```AbstractSearchQueryIT``` (fails for Elasticsearch)
 - [x] ```AbstractRangeQueryIT``` (not implented yet in PostgreSQL)

These issues need to be fixed.

tommydeboer:
I looked at the failing test in **```AbstractNotQueryIT```**. 

There are three person entities:

- Person 1: author of "MOLGENIS for dummies"
- Person 2: author of "MOLGENIS for dummies" and "Your database at the push of a button"
- Person 3: author of no books

We do a ```NOT EQUALS``` query on the ```mref```:

```java
new QueryImpl<Entity>().not().eq("authorOf", "MOLGENIS for dummies");
```

We would expect to see person 3 returned. But we get person 2. The PSQL query is as follows:

```sql
SELECT 
	this."id", "authorOf_filter1"."authorOf",
	(SELECT	
		array_agg(DISTINCT ARRAY["authorOf"."order"::TEXT, "authorOf"."authorOf"::TEXT])
	FROM
		"query_person_authorOf" AS "authorOf"
	WHERE
		this."id" = "authorOf"."id") AS "authorOf"
FROM
	"query_person" AS this
LEFT JOIN
	"query_person_authorOf" AS "authorOf_filter1"
ON	
	(this."id" = "authorOf_filter1"."id")
WHERE 
	NOT "authorOf_filter1"."authorOf" = 'MOLGENIS for dummies' LIMIT 1000;
```

The problem lies in the way the tables are joined:
![image](https://cloud.githubusercontent.com/assets/5362890/14982499/1de4f35c-1137-11e6-96e2-ca3da4b1856c.png)

1. Because person 3 has no entries in the ```authorOf``` table, it is ignored by the ```WHERE``` clause
2. Because the books are checked per joined row, person 2 is not excluded by the ```WHERE``` clause

Mark-de-Haan:
Looked at the failing  **AbstractSearchQueryIT** for _Elasticsearch_

[...]

PR #4823, opened 2016-05-04
Fix like query on mref not working in PostgreSQL backend

tommydeboer:
- This fixes like queries on mrefs in PostgreSQL backend
- See #4817


Mark-de-Haan:
test this please

PR #4825, opened 2016-05-04
Fix #4691 ImportApi check if repo exists based on filename, which is nonsense for EMX

### LINKED ISSUE #4691: ImportApi check if repo exists based on filename, which is nonsense for EMX
the check:
```java
if (dataService.hasRepository(FilenameUtils.getBaseName(file.getName())))
{
   throw new MolgenisDataException(
   "A repository with name " + FilenameUtils.getBaseName(file.getName()) + " already exists");
}
```

is valid for VCF's, but nonsense for EMX where the filename has no relations with the actual entity names.

Mark-de-Haan:


molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportWizardController.java
```diff
@@ -428,11 +429,13 @@ private ImportRun importFile(HttpServletRequest request, File file, String actio
 	{
 		// no action specified? default is ADD just like the importerPlugin
 		ImportRun importRun;
+		String fileExtension = getExtension(file.getName());
 		DatabaseAction databaseAction = getDatabaseAction(file, action);
-		if (dataService.hasRepository(FilenameUtils.getBaseName(file.getName())))
+		if (fileExtension.contains("vcf") &&
```
bartcharbon: please use the existing constants, and also check for vcf.gz

molgenis-data-import/src/main/java/org/molgenis/data/importer/ImportWizardController.java
```diff
@@ -1,5 +1,6 @@
 package org.molgenis.data.importer;
 
+import static org.apache.commons.io.FilenameUtils.*;
```
bartcharbon: there is also an exisiting import "import org.apache.commons.io.FilenameUtils".

PR #4855, opened 2016-05-11
fix 4847: Batch update rest api V2 fails

bartcharbon:
#### Checklist
- [x] Unit tests
- [ ] Updated testplan

### DIFF
--- a/molgenis-data-validation/src/test/java/org/molgenis/data/validation/RepositoryValidationDecoratorTest.java
+++ b/molgenis-data-validation/src/test/java/org/molgenis/data/validation/RepositoryValidationDecoratorTest.java
@@ -10,6 +10,7 @@ import static org.molgenis.MolgenisFieldTypes.MREF;
 import static org.molgenis.MolgenisFieldTypes.STRING;
 import static org.molgenis.MolgenisFieldTypes.XREF;
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.fail;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


molgenis-data-validation/src/test/java/org/molgenis/data/validation/RepositoryValidationDecoratorTest.java
```diff
@@ -2010,6 +2014,33 @@ public void update()
 	}
 
 	@Test
```
Mark-de-Haan: You can set a test to expect an exception:

`@Test(expectedExceptions = { MolgenisDataException.class }, expectedExceptionsMessageRegExp = "The entity you are trying to update [id999] does not exist.")`

fdlk: But then the regex probably should escape the `[` ;-)

Mark-de-Haan: True :p

bartcharbon: I know, however I find the current solution clearer to read

tommydeboer:
The new exception message is only thrown when you update a whole entity, not when you do an attribute specific update:

Calling ```/v2/org_molgenis_test_TypeTest/xdatetime``` with an invalid id gives:

```java
{
  "errors": [
    {
      "message": "Operation failed. Identifier: 77 is not valid"
    }
  ]
}
```

PR #4864, opened 2016-05-12
Add PostgreSQL RANGE query

tommydeboer:
#### Checklist
- [x] Unit tests

Fixes one of the issues in #4817

### DIFF
--- a/molgenis-data-postgresql/src/main/java/org/molgenis/data/postgresql/PostgreSqlQueryGenerator.java
+++ b/molgenis-data-postgresql/src/main/java/org/molgenis/data/postgresql/PostgreSqlQueryGenerator.java
@@ -505,8 +505,8 @@ class PostgreSqlQueryGenerator
 								range.getClass().getSimpleName()));
 					}
 					Iterator<Object> rangeValues = ((Iterable) range).iterator();
-					parameters.add(rangeValues.next()); // smaller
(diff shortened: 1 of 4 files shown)


molgenis-data-postgresql/src/main/java/org/molgenis/data/postgresql/PostgreSqlQueryGenerator.java
```diff
@@ -494,9 +494,33 @@ private static String getSqlDropTable(String tableName)
...
+					{
+						throw new MolgenisDataException("Missing value for RANGE query");
+					}
+					if (!(range instanceof Iterable<?>))
+					{
+						throw new MolgenisDataException(format("RANGE value is of type [%s] instead of [Iterable]",
+								range.getClass().getSimpleName()));
+					}
+					Iterator<Object> rangeValues = ((Iterable) range).iterator();
+					parameters.add(rangeValues.next()); // smaller
```
jjettenn: // Smaller -> From  
// Bigger -> To

tommydeboer: I used the variable names from the ```rng``` method of ```QueryImpl``` to stay consistent. I agree that from and to are better :)

Issue #4194, opened 2015-11-24
"Message:null" when running annotator with wrongly configured resource


tommydeboer:
## To reproduce:
1. Upload test VCF from _molgenis-data-vcf_
2. Go to annotators
3. Set CADD resource to _<your_resource_location>/cadd/cadd.tsv.gz.tbi_
4. Run the CADD annotator
## Expected:

Message saying the resource isn't configured properly (or even better: "please use **cadd.tsv.gz** instead of **cadd.tsv.gz.tbi**")
## Actual:

```
Error! An error occurred. Please contact the administrator.
Message:null
```

bartcharbon:
disagree with the "expected" here, if the resource is not configured properly it should not be "active" if should be in the unavailable list.

PR #4885, opened 2016-05-23
#4194 As freemarkertemplate developer I want to set custom headers before/after jquery

bartcharbon:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
```diff
@@ -31,8 +31,12 @@
 
     <#if app_settings.cssHref?has_content>
         <link rel="stylesheet" href="<@resource_href "/css/${app_settings.cssHref?html}"/>" type="text/css">
-    </#if> 
-    
+    </#if>
+        <#if app_settings.customJavascript?has_content>
+            <#list app_settings.customJavascript?split(r"\s*,\s*", "r") as js_file_name>
+                <script src="<@resource_href "/js/${js_file_name?html}"/>"></script>
```
Mark-de-Haan: What if I want to use JS from an external source?
e.g.: `<script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.6/d3.min.js" charset="utf-8"></script>`

bartcharbon: - A faulty piece of customJavascript will cause freemarker errors on every page
yes if you break stuff it will break...
- non existing files will cause freemarker errors on every page
nope

molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
```diff
@@ -34,7 +34,7 @@
     </#if>
         <#if app_settings.customJavascript?has_content>
             <#list app_settings.customJavascript?split(r"\s*,\s*", "r") as js_file_name>
-                <script src="<@resource_href "${js_file_name?html}"/>"></script>
+                <script src="${js_file_name?html}"></script>
```
fdlk: The @resource_href macro adds a hash to the URL.
On a production server, we add a cache header for the js files that says they can be cached indefinitely.
But then we do need the hash or else they'll never be refreshed, even if they've changed.

PR #4888, opened 2016-05-24
Give a warning instead of throwing an exception when type is not found + refactor

jjettenn:


fdlk:
Breaks ProgressImplTest, see pull request build.
I've added a 2.0_reindex build to jenkins to check that it is currently working alright.

PR #4896, opened 2016-05-26
fix and clean up migration code

bartcharbon:


molgenis-data-migrate/src/main/java/org/molgenis/migrate/version/v1_22/Step32AddRowLevelSecurityMetadata.java
```diff
@@ -40,32 +38,34 @@ public void upgrade()
 		LOG.info("Updating entities table ...");
 
 		// update existing settings table
-		jdbcTemplate.execute("ALTER TABLE entities ADD COLUMN `rowLevelSecured` boolean");
+		// jdbcTemplate.execute("ALTER TABLE entities ADD COLUMN `rowLevelSecured` boolean");
```
tommydeboer: Should this be commented?

PR #4901, opened 2016-05-26
Fix #4895 Migration from 1.16 to 1.17 fails if an attribute part is also a lookup attribute

### LINKED ISSUE #4895: Migration from 1.16 to 1.17 fails if an attribute part is also a lookup attribute
#### Reproduce
- Load the following mysql dump: 
[reproduce.zip](https://github.com/molgenis/molgenis/files/281372/reproduce.zip)
- Deploy 1.13.0-BBMRI-NL (Github release)
- Manually set molgenis.version to 21 (bug in the 1.13.0-BBMRI-NL release)
- Now deploy app-bbmri-nl (branch)

#### Expected
Migrates

#### Observed
`ERROR: column fullName can not be null`

With an empty database it does migrate. Doing it version by version will also cause the constraint error.
I suspect this is not BBMRI specific.

Mark-de-Haan:
Fix for #4895

### DIFF
--- a/molgenis-data-migrate/src/main/java/org/molgenis/migrate/version/v1_17/Step27MetaDataAttributeRoles.java
+++ b/molgenis-data-migrate/src/main/java/org/molgenis/migrate/version/v1_17/Step27MetaDataAttributeRoles.java
@@ -170,7 +170,7 @@ public class Step27MetaDataAttributeRoles extends MolgenisUpgrade
 
 	private boolean compoundAttrIsAttributePart(String compoundAttrName)
 	{
-		return getCompoundAttrName(compoundAttrName) != "";
+		return getCompoundAttrName(compoundAttrName).isEmpty();
 	}
 
 	private String getCompoundAttrName(String attrName)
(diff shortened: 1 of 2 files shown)


Mark-de-Haan:
test this please

Mark-de-Haan:
test this please

molgenis-data-migrate/src/main/java/org/molgenis/migrate/version/v1_17/Step27MetaDataAttributeRoles.java
```diff
@@ -146,6 +144,54 @@ public void processRow(ResultSet rs) throws SQLException
...
+		{
+			public String mapRow(ResultSet rs, int rowNum) throws SQLException {
+				return rs.getString("fullName");
+			}
+		});
+	}
+
+	private boolean compoundAttrIsAttributePart(String compoundAttrName)
+	{
+		return getCompoundAttrName(compoundAttrName) != "";
```
dennishendriksen: String is a object, do not use ==/!= to compare strings, use equals (or in this case isEmpty) instead.

PR #4903, opened 2016-05-26
Fix #4699 Error when calling RestClientV2 without '/api/v2/' in URI

### LINKED ISSUE #4699: Error when calling RestClientV2 without '/api/v2/' in URI
#### Reproduce:
Create a piece of javascript with a call to the REST client without the '/api/v2' prefix:

```javascript
import RestClientV2 from "rest-client/RestClientV2";

var api = new RestClientV2();
var data = api.get('TestEntity', {});
```

#### Expected:
No problems, the client should take care of it:
```javascript
if(!resourceUri.startsWith('/api/')) {
    // assume that resourceUri is a entity name
    resourceUri = apiBaseUri + htmlEscape(resourceUri);
}
```

#### Actual:
```Uncaught ReferenceError: htmlEscape is not defined```

### PR DESCRIPTION
**Pull request by Mark-de-Haan**

### CODE REVIEW

File: molgenis-core-ui/src/main/javascript/modules/rest-client/RestClientV2.js [hunk 1]
```diff
@@ -3,6 +3,12 @@ import $ from 'jquery';
 
     var apiBaseUri = '/api/v2/';
 
+    var htmlEscape = function(string) {
+        return String(string).replace(/[&<>"'\/]/g, function(s) {
+            return entityMap[s];
+        });
+    };
+
```
bartcharbon: Is this code duplication the only way to solve this?

molgenis-core-ui/src/main/javascript/molgenis-global-webpack.js
```diff
@@ -35,3 +35,6 @@ window.top.molgenis.isXrefAttr = isXrefAttr;
 window.top.molgenis.isMrefAttr = isMrefAttr;
 window.top.molgenis.isCompoundAttr = isCompoundAttr;
 
+import { htmlEscape } from 'utils/HtmlUtils';
+
+window.top.molgenis.htmlEscape = htmlEscape;
```
fdlk: I think that isn't quite right. The "old" code put the htmlEscape function on the global/window scope. Now it gets put in the molgenis object, but `molgenis-dataexplorer/src/main/resources/js/dataexplorer-filter.js` and `molgenis-core-ui/src/main/resources/js/jquery.molgenis.xrefmrefsearch.js ` assume it's in the global scope.

Mark-de-Haan:
test this please

PR #4940, opened 2016-06-03
Rowlevel: Fix import and create bugs, fix existing tests

bartcharbon:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


molgenis-core-ui/src/main/java/org/molgenis/ui/MolgenisRepositoryDecoratorFactory.java
```diff
@@ -34,11 +35,12 @@
 	private final DataService dataService;
 	private final ExpressionValidator expressionValidator;
 	private final RepositoryDecoratorRegistry repositoryDecoratorRegistry;
+	private final UserAccountService userAccountService;
```
tommydeboer: This is never used

PR #4950, opened 2016-06-06
Fix #4949 Settings filter is broken (again!)

### LINKED ISSUE #4949: Settings filter is broken (again!)
#### Reproduce
Log in as admin, go to the settings plugin, type a string into the search box

#### Expected
Settings options get filtered

#### Observed
Error message is displayed:
> Query with operator [LIKE] and data type [TEXT] not supported

Broken in master (tested on molgenis01)
Oddly enough: Works fine on 1.21.1 (tested on molgenis09)

Mark-de-Haan:


fdlk:
We think it's not dependent on the version, but breaks when you add languages!

Issue #4910, opened 2016-05-30
Internationalisation is broken for xmref and mref labels


Mark-de-Haan:
#### Reproduce
- Import the test questionnaire in molgenis-questionnaire/src/test/resources
- Go the the dataexplorer and select Test questionnaire
- Add a new Row and select the Gender mref from the tree

#### Expected
The labels in the **Genders?** column should be the one from the selected language

#### Observed
The labels are always Dutch, even if you switch languages

Note: The Table component uses the label attribute to determine which label to show. So if Label (en) is your label attribute in the EMX, then only the English labels will be shown for xref/mref fields

fdlk:
Still not quite okay when I open the form, I think:
<img width="725" alt="screen shot 2016-06-03 at 21 09 00" src="https://cloud.githubusercontent.com/assets/1523838/15790050/82b9a81c-29cf-11e6-8334-c831e4ef05b5.png">

fdlk:
I take it all back! Works fine.

PR #4954, opened 2016-06-08
Fix #4949 Settings filter is broken

### LINKED ISSUE #4949: Settings filter is broken (again!)

Mark-de-Haan:
The fix for #4910 causes issue #4949 
Might some slight changes to fix both issues.

## Note
It is not allowed to use multiple label attributes per entity!


molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntityMetaData.java
```diff
@@ -672,9 +672,14 @@ public void onChange(Package package_)
 	@Override
 	public AttributeMetaData getLabelAttribute(String languageCode)
 	{
-		String i18nLabelAttrName = getCachedAllAttrs().keySet().stream().filter(
-				attributeName -> attributeName.endsWith('-' + languageCode)).findFirst().orElse(null);
-		return i18nLabelAttrName != null ? getCachedAllAttrs().get(i18nLabelAttrName) : getLabelAttribute();
+		AttributeMetaData labelAttr = getLabelAttribute();
+		String labelAttributeName = labelAttr.getName();
+		if(labelAttributeName.endsWith("-en"))
```
bartcharbon: use "-"+LanguageService.FALLBACK_LANGUAGE to prevent the functionality from breaking when we change the default language

molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntityMetaData.java
```diff
@@ -672,9 +673,14 @@ public void onChange(Package package_)
 	@Override
 	public AttributeMetaData getLabelAttribute(String languageCode)
 	{
-		String i18nLabelAttrName = getCachedAllAttrs().keySet().stream().filter(
-				attributeName -> attributeName.endsWith('-' + languageCode)).findFirst().orElse(null);
-		return i18nLabelAttrName != null ? getCachedAllAttrs().get(i18nLabelAttrName) : getLabelAttribute();
+		AttributeMetaData labelAttr = getLabelAttribute();
+		String labelAttributeName = labelAttr.getName();
+		if(labelAttributeName.endsWith(FALLBACK_LANGUAGE))
```
bartcharbon: You'll probably want to check for the "-" as a prefix as well

bartcharbon:
test this please

PR #4969, opened 2016-06-17
Improve reindex performance. Query the SQL in batches, query the ES referencing entities using scan-scroll.

fdlk:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


molgenis-data-validation/src/main/java/org/molgenis/data/validation/EntityAttributesValidator.java
```diff
@@ -86,12 +86,12 @@
...
 				case CATEGORICAL:
 				case XREF:
-					violation = checkXref(entity, attr, meta);
+//					violation = checkXref(entity, attr, meta);
 					break;
 				case CATEGORICAL_MREF:
 				case FILE:
 				case MREF:
-					violation = checkMref(entity, attr, meta);
+//					violation = checkMref(entity, attr, meta);
```
fdlk: This is causing side-effects for null mref values.

PR #4983, opened 2016-06-24
Fix "app not starting" + W.I.P. Annotators + fix tests

bartcharbon:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/filter/Location.java
```diff
@@ -17,10 +17,10 @@
 {
 	public abstract String getChrom();
 
-	public abstract long getPos();
+	public abstract int getPos();
```
Mark-de-Haan: Why the type change?

molgenis-data/src/main/java/org/molgenis/data/meta/model/AttributeMetaData.java
```diff
@@ -536,4 +536,6 @@ private static String toEnumOptionsString(List<String> enumOptions)
 	{
 		return !enumOptions.isEmpty() ? enumOptions.stream().collect(joining(",")) : null;
 	}
+
+
```
Mark-de-Haan: No changes in this file

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/entity/AnnotatorConfig.java
```diff
@@ -0,0 +1,9 @@
+package org.molgenis.data.annotation.entity;
+
+/**
+ * Created by charbonb on 24/06/16.
+ */
```
Mark-de-Haan: IntelliJ comments

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/filter/Location.java
bartcharbon: because the type was wrong

PR #4990, opened 2016-06-27
Fix ReindexActionRegisterServiceTest.

fdlk:
Inline the createReindexAction method because it no longer does any type checking.
ACTION_ORDER attribute is INT not STRING, updated it.
N.B. I suspect this is an incompatible metadata change, so clean your database.


dennishendriksen:
> N.B. I suspect this is an incompatible metadata change, so clean your database.
@fdlk Meta data change is processed automatically during bootstrapping, cleanup not required ;-)

PR #5007, opened 2016-06-30
Refactoring + streaming aggregation of samples

tommydeboer:


molgenis-app/src/main/java/org/molgenis/app/promise/mapper/RadboudMapper.java
```diff
@@ -90,94 +57,36 @@ public MappingReport map(Entity project)
...
 			});
 
-			LOG.info("Adding {} entities to {}", sampleCollectionsToAdd.size(), SAMPLE_COLLECTIONS_ENTITY);
-			dataService.add(SAMPLE_COLLECTIONS_ENTITY, sampleCollectionsToAdd.stream());
-
-			LOG.info("Updating {} entities in {}", sampleCollectionsToUpdate.size(), SAMPLE_COLLECTIONS_ENTITY);
-			dataService.update(SAMPLE_COLLECTIONS_ENTITY, sampleCollectionsToUpdate.stream());
-
 			LOG.info("Finished mapping RADBOUD biobanks");
 			report.setStatus(Status.SUCCESS);
```
fdlk: Probably would be informative to add a message as well, detailing how many samples and biobanks were processed?

molgenis-app/src/main/java/org/molgenis/app/promise/mapper/RadboudMapper.java
```diff
@@ -90,94 +57,36 @@ public MappingReport map(Entity project)
...
-				String diseaseId = diseaseEntity.getString(XML_IDAA);
-				diseaseIdMap.putIfAbsent(diseaseId, newArrayList());
-				diseaseIdMap.get(diseaseId).add(diseaseEntity);
-			});
-
-			LOG.info("Downloading biobank data for " + project.getString(NAME));
-			promiseDataParser.parse(credentials, 0, biobankEntity -> {
-				LOG.info("Mapping biobank: " + biobankEntity.getString(XML_TITLE));
+			LOG.info("Generating RADBOUD sample");
+			promiseDataParser.parse(credentials, 1, samples::addSample);
```
fdlk: Nice!

molgenis-app/src/main/java/org/molgenis/app/promise/mapper/RadboudSampleMap.java
```diff
@@ -0,0 +1,433 @@
...
+		}
+		else
+		{
+			return null;
+		}
+	}
+
+	private class AggregatedSampleInfo
+	{
+		private Set<String> materialIds = newHashSet();
```
fdlk: This means the order of the Ids will be completely random, and probably confusing.
A LinkedHashSet would keep the order in which they were collected.
A TreeSet would sort alphabetically (probably the nicest?)

PR #5038, opened 2016-07-11
[MOLGENIS 2.0] Fix #5036 EntityListeners seem to be broken, restart needed to change settings

### LINKED ISSUE #5036: [MOLGENIS 2.0] Fix EntityListeners seem to be broken, restart needed to change settings
#### Reproduce
TODO

#### Expected
TODO

#### Observed
In existing platform integration test.

jjettenn:
#### Checklist
- [x] Unit tests
- [x] Updated testplan


molgenis-data/src/main/java/org/molgenis/data/listeners/EntityListenersService.java
```diff
@@ -0,0 +1,43 @@
...
+{
+	private static final ConcurrentHashMap<String, SetMultimap<Object, EntityListener>> entityListeners = new ConcurrentHashMap<String, SetMultimap<Object, EntityListener>>();
+
+	/**
+	 * returns a SetMultimap<Object, EntityListener> where the object is the id of the entity and the EntityListener is listener interface
+	 *
+	 * @param repoFullName
+	 * @return SetMultimap<Object, EntityListener>
+	 */
+	public synchronized SetMultimap<Object, EntityListener> getEntityListeners(String repoFullName)
```
fdlk: This one probably shouldn't be synchronized cause it'll get executed a LOT of times concurrently.

molgenis-data/src/main/java/org/molgenis/data/listeners/EntityListenersService.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.common.collect.SetMultimap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+
+import javax.annotation.PreDestroy;
+import java.util.concurrent.ConcurrentHashMap;
+
+@Service
+public class EntityListenersService
```
fdlk: Separate class deserves separate unit test, I think?

molgenis-data/src/main/java/org/molgenis/data/listeners/EntityListenersService.java
```diff
@@ -0,0 +1,43 @@
...
+	/**
+	 * returns a SetMultimap<Object, EntityListener> where the object is the id of the entity and the EntityListener is listener interface
+	 *
+	 * @param repoFullName
+	 * @return SetMultimap<Object, EntityListener>
+	 */
+	public synchronized SetMultimap<Object, EntityListener> getEntityListeners(String repoFullName)
+	{
+		if (!entityListeners.containsKey(repoFullName))
+			entityListeners.put(repoFullName, HashMultimap.<Object, EntityListener>create());
```
fdlk: Should also be concurrent

molgenis-data/src/main/java/org/molgenis/data/listeners/EntityListenersService.java
```diff
@@ -0,0 +1,105 @@
...
+		{
+			entityListeners.put(repoFullName, HashMultimap.create());
+		}
+	}
+
+	Stream<Entity> updateEntities(String repoFullName, Stream<Entity> entities)
+	{
+		verifyRepoRegistered(repoFullName);
+		SetMultimap<Object, EntityListener> entityListeners = this.entityListeners.get(repoFullName);
+		synchronized (entityListeners)
```
fdlk: I think it should be possible to concurrently read from the listeners, perhaps you need to create an immutable copy or so. But that'll help prevent synchronization issues. This synchronized block effectively locks the repository.

molgenis-data/src/main/java/org/molgenis/data/listeners/EntityListenersService.java
```diff
@@ -0,0 +1,140 @@
...
+	 * @return Stream<Entity>
+	 */
+	Stream<Entity> updateEntities(String repoFullName, Stream<Entity> entities)
+	{
+		verifyRepoRegistered(repoFullName);
+		SetMultimap<Object, EntityListener> entityListeners = this.entityListeners.get(repoFullName);
+		return entities.filter(entity -> {
+			Set<EntityListener> entityEntityListeners = entityListeners.get(entity.getIdValue());
+			entityEntityListeners.forEach(entityListener -> {
+				entityListener.postUpdate(entity);
```
fdlk: This one reads through the listeners without locking it (which is good cause otherwise you cannot do concurrent updates) but now I *think* it may still happen that the addListener() method is concurrently adding entries to the map while you iterate through it with this forEach.

In practice I suspect the add() will only be called at startup so we won't run into this bug.

molgenis-data/src/main/java/org/molgenis/data/listeners/EntityListenersService.java
```diff
@@ -0,0 +1,140 @@
...
+	 * @param repoFullName
+	 * @param entityListener entity listener for a entity
+	 */
+	public void addEntityListener(String repoFullName, EntityListener entityListener)
+	{
+		verifyRepoRegistered(repoFullName);
+		SetMultimap<Object, EntityListener> entityListeners = this.entityListeners.get(repoFullName);
+		synchronized (entityListeners)
+		{
+			entityListeners.put(entityListener.getEntityId(), entityListener);
```
fdlk: Here's where the insert happens. This method synchronizes on the SetMultimap but the reading threads don't.

molgenis-data/src/main/java/org/molgenis/data/listeners/EntityListenersService.java
fdlk: After some googling: How about we use a http://tutorials.jenkov.com/java-util-concurrent/readwritelock.html then we can let all read access to the listeners happen in a block that does lock.readLock().lock() and inserting or removing listeners happen in a block that does lock.writeLock().lock() !

molgenis-data/src/main/java/org/molgenis/data/listeners/EntityListenersService.java
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	/**
+	 * Verify that the repository is registered
+	 *
+	 * @param repoFullName
+	 */
+	protected void verifyRepoRegistered(String repoFullName)
+	{
+		if (!entityListeners.containsKey(requireNonNull(repoFullName)))
```
fdlk: readlock!!

PR #5046, opened 2016-07-14
move cmdLineAnnotator code to seperate module

bartcharbon:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


molgenis-annotators-cmd/src/main/java/org/molgenis/annotation/cmd/CmdLineAnnotator.java
```diff
@@ -1,4 +1,4 @@
-package org.molgenis.data.annotation.cmd;
+package org.molgenis.annotation.cmd;
```
Mark-de-Haan: /**
  * Build JAR file...............: mvn clean install -pl molgenis-data-annotators/ -am -DskipTests -P create-delivery
  * Run..........................: java -jar molgenis-data-annotators/target/CmdLineAnnotator.jar
  */

Change javadoc example commands to correct module paths

molgenis-data-annotators/pom.xml
```diff
@@ -86,45 +86,6 @@
...
-                                    <goal>single</goal>
-                                </goals>
-                            </execution>
-                        </executions>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
-    </profiles>
-
```
Mark-de-Haan: This was removed but not added to the new POM. How to create runnable jar?

molgenis-annotators-cmd/src/main/java/org/molgenis/annotation/cmd/CmdLineAnnotator.java
bartcharbon: this is fixed, however the line note is not on the line of the problem, so github keeps showing this comment.

molgenis-data-annotators/pom.xml
bartcharbon: this is fixed, however the line note is not on the line of the problem, so github keeps showing this comment.

molgenis-annotators-cmd/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+        <profile>
+            <id>create-delivery</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-assembly-plugin</artifactId>
+                        <configuration>
+                            <archive>
+                                <manifest>
+                                    <mainClass>org.molgenis.data.annotation.cmd.CmdLineAnnotator</mainClass>
```
Mark-de-Haan: this should be **org.molgenis.annotation.cmd.CmdLineAnnotator**
Right now the Main class is not properly packaged in the Jar

Issue #4495, opened 2016-02-15
Duplicate key error when importing emx_deep_nesting.xlsx


tommydeboer:
#### Reproduce
1. Import *molgenis/molgenis-app/src/test/resources/xls/***_emx_deep_nesting.xlsx_**

#### Expected
File imports without errors.

#### Actual:
```java
java.lang.IllegalStateException: Duplicate key AttributeMetaData(name='Identifier' dataType='string' 
description='null' fieldType='string' nillable='false' readOnly='true' defaultValue='null' expression='null' 
label='Identifier' visible='true' unique='true' visible='true' auto='false' attributesMetaData='null' 
aggregateable='false' range='null' visibleExpression='null' validationExpression='null')
```

PR #5048, opened 2016-07-15
Fix #5024 Lookup attribute, label attribute is not inherited by entities that implement an abstract entity

### LINKED ISSUE #5024: Lookup attribute, label attribute is not inherited by entities that implement an abstract entity
#### Reproduce
1.  Create an abstract entity OntologyTerm with an attribute code that is a lookup attribute and an attribute term that is both a lookup attribute and a label attribute
2. Create an entity Disease that extends OntologyTerm
3. Create an entity Sample that has an XREF attribute disease that refers to the disease entity
4. Upload your model

#### Expected
I can filter on disease in the sample entity

#### Observed
1. When I try to set a filter on disease I get an error 'lookup attribute not specified'
2. In the model registry my Disease entity does not have the lookup attribute or label attribute set

Issue reproduced on 1.21.1

tommydeboer:

### DIFF
--- a/molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntityMetaData.java
+++ b/molgenis-data/src/main/java/org/molgenis/data/support/DefaultEntityMetaData.java
@@ -1,5 +1,6 @@
 package org.molgenis.data.support;
 
+import autovalue.shaded.com.google.common.common.collect.Maps;
 import com.google.common.collect.Iterables;
 import org.molgenis.data.*;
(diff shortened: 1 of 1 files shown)


bartcharbon:
test this please

bartcharbon:
@tommydeboer did you check if the file metioned in "https://github.com/molgenis/molgenis/issues/4495" still uploads? this issue was also related to lookup attributes and abstract entities, might be useful to check for regression here.

PR #5064, opened 2016-07-20
Bugfixes #5061, #5063

### LINKED ISSUE #5061: Entities that are only specified in the attributes sheet do not get their backend set, resulting in a null pointer during importing
#### Reproduce
Import TypeTest
Add/Update TypeTest

#### Expected
Import succeeds

#### Observed
Null pointer

### LINKED ISSUE #5063: Performing an add/update with compounds results in the attribute_parts table being updated with null values
#### Reproduce
Import TypeTest
Add/Update with TypeTest

#### Expected
Import succeeds

#### Observed
The attribute 'parts' of entity 'sys_md_attributes_parts' can not be null.

Mark-de-Haan:
- Fix #5061 Entities that are only specified in the attributes sheet do not get their backend set, resulting in a null pointer during importing
- Fix #5063 Performing an add/update with compounds results in the attribute_parts table being updated with null values


molgenis-data/src/main/java/org/molgenis/data/meta/MetaDataServiceImpl.java
```diff
@@ -272,8 +272,10 @@ public Package getPackage(String string)
 			throw new UnknownEntityException(format("Unknown entity [%s]", entityMeta.getName()));
 		}
 
-		// add/update attributes, attributes are deleted when deleting entity meta data if no more references exist
-		Iterable<AttributeMetaData> ownAttrs = entityMeta.getOwnAttributes();
+		// FIXME Is this true??
```
jjettenn: Remove this comment from code and add please add this comment into the TODO list MOLGENIS 2.0.

molgenis-data/src/main/java/org/molgenis/data/meta/MetaDataServiceImpl.java
```diff
@@ -272,8 +272,10 @@ public Package getPackage(String string)
...
+		// FIXME add/update attributes, attributes are deleted when deleting entity meta data if no more references exist
+		Iterable<AttributeMetaData> ownAttrs = entityMeta.getCompoundOrderedAttributes();
```
jjettenn: rename ownAttr -> suggestion: compoundOrdedAttributes

molgenis-data/src/main/java/org/molgenis/data/meta/MetaUtils.java
```diff
@@ -29,7 +30,7 @@ public static Fetch getEntityMetaDataFetch()
 				.getName();
 
 		EntityMetaData existingEntityMetaData = metaDataService.getEntityMetaData(entityMeta.getName());
-		if (!existingEntityMetaData.getBackend().equals(backend))
+		if (!Objects.equal(backend, existingEntityMetaData.getBackend()))
```
jjettenn: Please check not null
Objects.equal is null tolerant

molgenis-data/src/main/java/org/molgenis/data/meta/model/EntityMetaData.java
```diff
@@ -428,6 +429,32 @@ public EntityMetaData setExtends(EntityMetaData extends_)
...
+	 */
+	public List<AttributeMetaData> getCompoundOrderedAttributes()
+	{
+		List<AttributeMetaData> attributes = newArrayList();
+		getEntities(ATTRIBUTES, AttributeMetaData.class).forEach(attribute -> {
+			if (attribute.getDataType() == COMPOUND)
+			{
+				attribute.getAttributeParts().forEach(attributePart -> attributes.add(attributePart));
+			}
+			attributes.add(attribute);
```
jjettenn: please check if attribute is not already added

molgenis-data/src/main/java/org/molgenis/util/EntityUtils.java
```diff
@@ -356,10 +360,39 @@ public static boolean equals(AttributeMetaData attr, AttributeMetaData otherAttr
...
+	 *
+	 * @param attr
+	 * @param otherAttr
+	 * @return
+	 */
+	public static boolean equals(AttributeMetaData attr, AttributeMetaData otherAttr)
+	{
+		if (attr == null || otherAttr == null)
+		{
+			if (attr == null && otherAttr == null) return true;
```
jjettenn: remove this: if (attr == null && otherAttr == null) return true;
 if (attr == null || otherAttr == null) return false is enough

PR #5071, opened 2016-07-25
Fix unit test 2.0

Mark-de-Haan:


molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/utils/VcfWriterUtils.java
```diff
@@ -368,8 +368,9 @@ private static boolean writeSingleInfoField(Entity vcfEntity, BufferedWriter wri
 				hasInfoFields = true;
 			}
 		}
-		else
+		else if (dataType != MREF && dataType != XREF && dataType != CATEGORICAL_MREF && dataType != CATEGORICAL)
```
dennishendriksen: Please add FILE data type or even better call EntityMetaDataUtils.isReferenceType(attributeMetaData).

bartcharbon: besides that, shouldn't we actually do something for those types instead of ignoring them completely?

PR #5072, opened 2016-07-25
Fix EMD test + Use isReferenceType everywhere in the VCF writer

Mark-de-Haan:


molgenis-data/src/main/java/org/molgenis/data/meta/model/EntityMetaData.java
```diff
@@ -443,28 +442,28 @@ public EntityMetaData setExtends(EntityMetaData extends_)
...
-		Set<AttributeMetaData> attributes = newHashSet();
+		List<AttributeMetaData> attributes = newArrayList();
 		getEntities(ATTRIBUTES, AttributeMetaData.class).forEach(attribute -> {
 			if (attribute.getDataType() == COMPOUND)
 			{
 				attribute.getAttributeParts()
 						.forEach(attributePart -> resolvePossibleNestedCompounds(attributePart, attributes));
 			}
-			attributes.add(attribute);
+			if (!attributes.contains(attribute)) attributes.add(attribute);
```
dennishendriksen: 1) Potentially expensive check for large lists. Use a Set or a LinkedSet in case you want to preserve order. 2) Coding standard: add curly braces for one line if statements

PR #5073, opened 2016-07-25
Fix mapping service unit tests

dennishendriksen:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


molgenis-data-mapper/src/test/java/org/molgenis/data/mapper/service/impl/MappingServiceImplTest.java
```diff
@@ -186,367 +242,268 @@ public void testGetMappingProject()
...
-		try
-		{
-			retrieved.getMappingTarget(HOP_ENTITY).addSource(geneMetaData);
-			fail("Expected exception");
-		}
-		catch (IllegalStateException ignored)
-		{
-		}
-	}
+		assertTrue(EntityUtils.entitiesEquals(streamCaptor.getValue(), expectedEntities.stream()));
```
Mark-de-Haan: entitiesEquals(Stream<Entity>, Stream<Entity>) does not exist in EntityUtils. Forgot to commit?

Mark-de-Haan:
Merge with 2.0 and add missing methods, then its OK

PR #5074, opened 2016-07-26
Fix non-superuser is not allowed to view meta data entities

dennishendriksen:


Mark-de-Haan:
Can you merge with 2.0 to remove changes that are already present in the 2.0 branch?

PR #5093, opened 2016-07-29
Improve indexing and bootstrapping performance

dennishendriksen:


molgenis-data-elasticsearch/src/main/java/org/molgenis/data/elasticsearch/ElasticsearchEntityFactory.java
```diff
@@ -1,42 +1,227 @@
...
+					generator.writeEndArray();
+				}
+				else
+				{
+					generator.writeNull();
+				}
+				break;
+			}
+			case COMPOUND:
+				throw new RuntimeException(format("Illegal attribute type [%s]", attrType.toString()));
```
jjettenn: Why is the compound not supported?

PR #5108, opened 2016-08-04
UI + CLI annotator exceptions + GAVIN fixes

tommydeboer:
This PR adds:
- Two fixes where GAVIN code was pointing to old entity name (in root package).
- Annotators that break during annotation report information about the failed entity.


tommydeboer:
test this please

PR #5111, opened 2016-08-04
Fix mapping service create integrated data set

dennishendriksen:
#### Checklist
- [x] Unit tests

### DIFF
--- a/molgenis-data/src/main/java/org/molgenis/data/meta/model/AttributeMetaData.java
+++ b/molgenis-data/src/main/java/org/molgenis/data/meta/model/AttributeMetaData.java
@@ -1,13 +1,12 @@
 package org.molgenis.data.meta.model;
 
-import com.google.common.collect.Maps;
+import autovalue.shaded.com.google.common.common.collect.Lists;
 import org.molgenis.MolgenisFieldTypes.AttributeType;
 import org.molgenis.data.Entity;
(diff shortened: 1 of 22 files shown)


molgenis-data/src/main/java/org/molgenis/data/meta/model/AttributeMetaData.java
```diff
@@ -60,46 +60,21 @@ public AttributeMetaData(String attrId, EntityMetaData entityMeta)
...
-		requireNonNull(attrMeta.getName());
-
-		if(copied.containsKey(attrMeta.getName()))
-		{
-			return (AttributeMetaData) copied.get(attrMeta.getName());
-		}
-
-		EntityMetaData entityMeta = attrMeta.getEntityMetaData();
-		AttributeMetaData attrMetaCopy = new AttributeMetaData(entityMeta);
+		AttributeMetaData attrMetaCopy = new AttributeMetaData(attrMeta.getEntityMetaData()); // do not deep-copy
```
tommydeboer: The AttributeMetaData should not refer to existing attributes right? So why not deep copy?

dennishendriksen: All AttributeMetaData should have the same reference to AttributeMetaDataMetaData

PR #5116, opened 2016-08-05
Fix importing entity sheets with language columns like 'description-nl'

Mark-de-Haan:


molgenis-data/src/main/java/org/molgenis/data/support/DynamicEntity.java
```diff
@@ -199,25 +198,24 @@ public void set(Entity values)
...
-		{
-			return;
-		}
+		if (value == null) return;
 
 		AttributeMetaData attr = entityMeta.getAttribute(attrName);
 		if (attr == null)
 		{
-			throw new UnknownAttributeException(format("Unknown attribute [%s]", attrName));
+			// This attribute does not exist yet, this can be because of a i18n situation
```
dennishendriksen: The validation is meant to check for cases where you set data for non-existing attributes. Consider fixing this problem upstream such that setting data for non-existing attributes does not occur.

PR #5190, opened 2016-08-16
M4419: Apply metadata compatibility check when mapping to an existing repository

Mark-de-Haan:
#### Checklist
- [x] Unit tests


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
```diff
@@ -183,6 +188,14 @@ public String applyMappings(MappingTarget mappingTarget, String entityName, bool
 			// Get an existing repository
 			targetRepo = dataService.getRepository(entityName);
 
+			// Compare the metadata between the target repository and the mapping target
+			// Returns detailed information in case something is not compatible
+			String mappingException = compareTargetMetaDatas(targetRepo.getEntityMetaData(), targetMetaData);
+			if (mappingException != null)
```
fdlk: Let the checker method throw the exception then, I think. Returning null doesn't make things clearer.

fdlk:
cannot test due to #5192

PR #5205, opened 2016-08-18
M4187 L3 (query) cache

fdlk:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


dennishendriksen:
@fdlk How does the query cache deal with repositories that return different results for different users for the same query? Specifically for the current temporary solution for entity-permission based row-level security for entities and attributes.

fdlk:
@dennishendriksen The query cache sits right above the IndexedRepositoryDecorator. It behaves the same as that decorator in the sense that it does not care about the user that calls it.

A future, different, implementation of row-level-security may perhaps result in the IndexedRepositoryDecorator returning different IDs for the same query, depending on who's asking it. In that case we can/should add the caller to the cache key for row-level-secured repositories. It's something to keep in mind when designing the row-level-security.

PR #5207, opened 2016-08-18
Fix #5206 Entity meta data validation missing checks

### LINKED ISSUE #5206: Entity meta data validation missing checks
Entity meta data validation is missing the following checks:
- Qualified name equal to package name + underscore + name
- Own attributes not owned by other entities
- Own attribute names don't exist in parent entities
- Own ID attribute is in the list of attributes
- Own ID attribute is unique
- Own ID attribute is non-nillable
- Own ID attribute has allowed data type
- Own label attribute is in the list of attributes
- Own lookup attributes are in in the list of attributes
- ID attribute exists for non-abstract entity with no parent entity
- ID attribute exists for non-abstract entity with parent entity without ID attribute
- Backend exists

dennishendriksen:
#### Checklist
- [x] Unit tests


molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/EntityMetaDataValidator.java
```diff
@@ -0,0 +1,212 @@
...
+{
+	private final DataService dataService;
+
+	@Autowired
+	public EntityMetaDataValidator(DataService dataService)
+	{
+		this.dataService = requireNonNull(dataService);
+	}
+
+	public void validate(EntityMetaData entityMeta)
```
jjettenn: Can you split this method into smaller methods? It is beter for the Cohesion

dennishendriksen:
test this please

PR #5209, opened 2016-08-18
Fix #5208 Attribute ref entity updates succeed if they shouldn't

### LINKED ISSUE #5208: Attribute ref entity updates succeed if they shouldn't
#### Reproduce
- Import ```molgenis-app\src\test\resources\xls\emx_all_datatypes.xlsx```
- In the entities table edit the attribute ```xxref_value```
- Update the Ref entity from TypeTestRef to something else
- Save changes

#### Expected
- Error that ref entity could not be updated due to unknown value

#### Observed
- Update successful, data explorer displays ```xxref_value``` column with empty data

dennishendriksen:
#### Checklist
- [x] Unit tests


fdlk:
If I import a TypeTestRef2 entity with same values as in TypeTestRef, I can change the refEntity of xref and mref attributes in TypeTest from TypeTestRef to TypeTestRef2. 
For xref attributes this works great, and even the values are preserved, so an entity that referred to ref1 in TypeTestRef now refers to ref1 in TypeTestRef2.
For mref attributes this also works but the values of the mref get cleaned. I'd expect them to behave the same, so either both get cleared.

PR #5225, opened 2016-08-29
Fix #5230: QueryRule equals + integration test method for caching performance measurement

### LINKED ISSUE #5230: QueryRule equals method too strict
#### Reproduce
Compare two identical QueryRule instances, but with Entity value set to two instances of an entity with the same idValue.

#### Expected
equals() returns true

#### Observed
equals() returns false

This causes unnecessary misses in the L3 / query cache.

Mark-de-Haan:
#### Checklist
- [x] Unit tests

### DIFF
--- a/molgenis-data-cache/src/main/java/org/molgenis/data/cache/l3/L3CacheRepositoryDecorator.java
+++ b/molgenis-data-cache/src/main/java/org/molgenis/data/cache/l3/L3CacheRepositoryDecorator.java
@@ -3,6 +3,7 @@ package org.molgenis.data.cache.l3;
 import org.molgenis.data.*;
 import org.molgenis.data.support.QueryImpl;
 import org.molgenis.data.transaction.TransactionInformation;
+import org.slf4j.Logger;
 
 import java.util.List;
 import java.util.stream.Stream;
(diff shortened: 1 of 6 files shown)


bartcharbon:
please fix the failing test

molgenis-data-cache/src/main/java/org/molgenis/data/cache/l3/L3CacheRepositoryDecorator.java
```diff
@@ -53,11 +57,19 @@ public L3CacheRepositoryDecorator(Repository<Entity> decoratedRepository, L3Cach
 	@Override
 	public Stream<Entity> findAll(Query<Entity> query)
 	{
-		if (transactionInformation.isRepositoryCompletelyClean(getName()) && cacheable && query.getPageSize() > 0
-				&& query.getPageSize() <= MAX_PAGE_SIZE)
+		if (transactionInformation.isRepositoryCompletelyClean(getName()))
+		{
+			// FIXME page size for metadata is always 0, and batching is done by the postgres repository
```
bartcharbon: is there a story or issue to make sure these FIXME's get addressed?

fdlk: I've already added it under the header "Issues" in the FD/TD. Can also file a bug here.

fdlk:
Closed for #5239

PR #5239, opened 2016-09-01
Fix #5230: QueryRule equals + integration test method for caching performance measurement

### LINKED ISSUE #5230: QueryRule equals method too strict

fdlk:
#### Checklist
- [x] Unit tests
- [ ] Updated testplan


bartcharbon:
integration test: testFindQueryOperatorRange[10,10,[0]] fails

fdlk:
For the range operator, storing the values in a set is not particularly wise :)

fdlk:
Test this please

fdlk:
test this please

fdlk:
test this please

PR #5249, opened 2016-09-06
Introduce One-to-many attribute type

dennishendriksen:
#### Checklist
- [x] Unit tests
- [ ] Updated testplan


molgenis-app/src/main/resources/css/gaflistimporter.css
```diff
@@ -1,15 +0,0 @@
-.table th, .table td {
```
fdlk: Does not belong in this PR.

fdlk:
Is there a way we can e.g. pre-format the js in the master so that the .js changes are a bit easier to review?

molgenis-data-postgresql/src/main/java/org/molgenis/data/postgresql/PostgreSqlQueryGenerator.java
```diff
@@ -127,8 +127,23 @@ static String getSqlSetDataType(EntityMetaData entityMeta, AttributeMetaData att
...
+		{
+			tableEntityMeta = attr.getRefEntity();
+			columnSql = getSqlOrderColumn(attr);
+		}
+		else
+		{
+			tableEntityMeta = entityMeta;
+			columnSql = getSqlColumn(entityMeta, attr);
+		}
+		sql.append(getTableName(tableEntityMeta)).append(" ADD ").append(columnSql);
```
fdlk: What's the final verdict on cross-repo one-to-many?

Mark-de-Haan:
When applied over an existing database: **unknown column mappedBy**
This should be allowed with meta data editing right?

dennishendriksen:
@Mark-de-Haan Meta meta data editing is not supported in 2.0

molgenis-data-postgresql/src/main/java/org/molgenis/data/postgresql/PostgreSqlQueryGenerator.java
dennishendriksen: Bidirectional cross-backend one-to-many should be supported, I'll fix it.

Mark-de-Haan:
If I create a oneToMany attribute, and then try to add it to TypeTest:
_Caused by: org.postgresql.util.PSQLException: ERROR: relation "org_molgenis_test_TypeTest_oneToMany" does not exist_

dennishendriksen:
@Mark-de-Haan What does the content of the created attribute look like?

Mark-de-Haan:
@dennishendriksen 

image

Mark-de-Haan:
After adding the oneToMany attribute to TypeTest (which fails) and then going to the TypeTest entity:

`Error! org.molgenis.data.MolgenisDataException: ; nested exception is org.postgresql.util.PSQLException: ERROR: operator does not exist: integer = character varying Hint: No operator matches the given name and argument type(s). You might need to add explicit type casts. Position: 2038`

So I guess failing to add attributes is not roll backed?

molgenis-app/src/main/resources/css/gaflistimporter.css
dennishendriksen: Done, see #5260

dennishendriksen:
@Mark-de-Haan your data was successfully persisted so no rollback was required. I've added both a validation check and a fix which results in being able to add a onetomany in TypeTestRef mappedBy TypeTest.xref_value.

dennishendriksen:
Repository.forEachBatched broken if entity meta data contains one-to-many attributes

dennishendriksen:
Querying one-to-many attributes broken in some cases

tommydeboer:
Sort order not updated after changing the label of referenced attribute.

Reproduce using the Author/Book entities with sort order ASC on books attribute:
- Add two books: abook, xbook
- Add author with these two books
- Change label of xbook to axbook

Order is updated when removing one of the books from author, and then re-adding it.

tommydeboer:
Newly added book doesn't show up in entity select box.

Reproduce:
- Edit an existing author.
- Click the '+' sign next to the books entity select box
- Add a new book, and set the author to the author you are editing. Click Create
- Still in the 'edit author' form, click '**Save changes**'
- *Book not added to author. Book entity exists but has no author*

Similarly:
- Edit an existing author.
- Click the '+' sign next to the books entity select box
- Add a new book, and set the author to the author you are editing. Click Create
- Still in the 'edit author' form, click '**Cancel**'
- *Book apparently not added to author. Book entity exists and has author. Going back to Author entity: book appears*

tommydeboer:
Filtering on book label throws error:

```Error! ; nested exception is org.springframework.jdbc.BadSqlGrammarException: PreparedStatementCallback; bad SQL grammar [SELECT COUNT(DISTINCT this."id") FROM "sys_Author" AS this LEFT JOIN "sys_Author_books" AS "books_filter1" ON (this."id" = "books_filter1"."id") WHERE "books_filter1"."books" = ?]; nested exception is org.postgresql.util.PSQLException: ERROR: relation "sys_Author_books" does not exist Position: 70```

Reproduce:
- Add book to author
- Click on 'books' attribute in attribute tree
- Filter on book label

**Comment by tommydeboer** [comment]
Can't edit or delete many_to_one value when nillable is set to false.

Reproduce:
- Set attribute *author* of entity *Book* to **nillable=false**
- Add two *Author* entities without books: **author1, author2**
- Add a *Book* entity with author set to **author1** and click 'Create'
- Try to change the author or remove te book completely. This will result in an error:

```Error! The attribute 'author' of entity 'sys_Book' can not be null.```

**Comment by Mark-de-Haan** [comment]
Using the Author and Book metadata classes:
I want to create an Author:
```
Error! An error occurred. Please contact the administrator.
Uncaught Unknown data type: ONE_TO_MANY
```

I ignore the error, create a book and author combination
the resulting book value in the Author table looks like this:
screen shot 2016-09-19 at 15 25 53

dennishendriksen:
@Mark-de-Haan I cannot reproduce your problem. Could you please ensure that you have the latest code  and that you have no index and an empty database? Next step is to add the following resources:
[onetomanyentities.zip](https://github.com/molgenis/molgenis/files/481695/onetomanyentities.zip)
which I added in org.molgenis.app in module molgenis-app and afterwords launch the application.

fdlk:
We're going to integrate with languages PR and integration tests on a branch first and send to master from there.

PR #5250, opened 2016-09-06
Feature/m4237 refactor index algorithm

jjettenn:
To react on the current state of the data and not on the past commands that are probably not relevant anymore by a system recovery.

#### Checklist
- [x] Unit tests


molgenis-data-elasticsearch/src/main/java/org/molgenis/data/elasticsearch/index/job/IndexJob.java
```diff
@@ -16,6 +17,7 @@
 import org.slf4j.LoggerFactory;
 import org.springframework.security.core.Authentication;
 
+import javax.management.*;
```
bartcharbon: unused import according to intellij

PR #5267, opened 2016-09-15
Fix #5266 updating an user hashes the hashed password

### LINKED ISSUE #5266: updating an user results in your password hash being hashed
#### Reproduce
- update for example your phone number in the "account" plugin
- logout
- login

#### Expected
I can login

#### Observed
password is invalid due to the hashed password being hashed when the user was updates

bartcharbon:
#### Checklist
- [x] Unit tests
- [ ] Updated testplan


fdlk:
Shouldn't we simply hash the password client-side when the user enters one?
That would also help with the uploads from emx / migration.

PR #5272, opened 2016-09-20
add diagnositics module + first version of GeneNetwork import controller

bartcharbon:
#### Checklist
- [x] Unit tests
- [ ] Updated testplan


molgenis-genetics-diag/src/test/resources/testfile.txt
```diff
@@ -0,0 +1,4 @@
+12-12-2016  hpo1    hpo2    hpo3
```
Mark-de-Haan: A data as column name is correct?

bartcharbon: yes this is similar to the input file we received

PR #5274, opened 2016-09-20
Fix/adding deleting languages static

jjettenn:
#### Checklist
- [x] Unit tests
- [ ] Updated testplan

There are 8 standard languages: "en", "nl", "de", "es", "it", "pt", "fr", "xx"
en: is default
xx: is a placeholder for my language


molgenis-core-ui/src/main/javascript/modules/react-components/LanguageSelectBox.js
```diff
@@ -1,76 +1,81 @@
...
+            type: 'POST',
+            url: '/plugin/useraccount/language/update',
+            data: 'languageCode=' + language.id,
+            success: function () {
+                location.reload(true);
+            }
+        });
+    }
+});
+
```
fdlk: let's try to update formatting settings

molgenis-data-i18n/src/main/java/org/molgenis/data/i18n/LanguageRepositoryDecorator.java
```diff
@@ -247,88 +176,17 @@ public void deleteAll()
 	@Override
 	public void add(Language language)
 	{
-		// Add language
-		decorated.add(language);
 
-		if (!language.getCode().equals(DEFAULT_LANGUAGE_CODE))
+		if (!languageService.getLanguageCodes().anyMatch(e -> e.equals(language.getCode())))
```
fdlk: add method to languageService that does this

molgenis-data-i18n/src/main/java/org/molgenis/data/i18n/model/LanguageFactory.java
```diff
@@ -18,12 +18,14 @@
 	 *
 	 * @param code language code, e.g. "en"
 	 * @param name language name, e.g. "English"
+	 * @Param active language active, e.g "true"
```
fdlk: little p

PR #5304, opened 2016-10-03
M4548 As Chao I want support to merge three pull requests with MOLGENIS 2.0, Part-1

ChaoPang:
This one contains:
* New version of Explain API, less intertwined with ElasticSearch, relying on the stemming but no longer based on the ElasticSearch scores
* Extract two new classes TagGroupGenerator and QueryExpansionService out of SemanticSearchService
* ValueObjects OntologyTerm and CombinedOntologyTerm with interface OntologyTagObject that allows both to be used as tag objects
* ExplainMatchCandidate can explain matches to String, used by new Explain API, and then the SemanticSearchService turns the explanations into matches to AttributeMetaData for use externally.


fdlk:
Will take a look at the formatting first.

fdlk:
test this please

fdlk:
I've put up a pull request on Chao's repo with some cleanup and minor fixes

molgenis-ontology-core/src/main/java/org/molgenis/ontology/core/repository/OntologyTermRepository.java
```diff
@@ -126,217 +148,433 @@ private boolean isOntologyTermExactMatch(Set<String> terms, OntologyTerm ontolog
...
+		List<String> nodePaths = ontologyTerm.getNodePaths();
+
+		for (int i = 0; i < maxLevel; i++)
+		{
+			nodePaths = nodePaths.stream().map(this::getParentNodePath).filter(StringUtils::isNotBlank)
+					.collect(toList());
+
+			if (nodePaths.size() > 0)
+			{
+				// TODO: my teammates will take look
```
fdlk: Explanation from Chao:
Node paths are large and therefore stored as text.
Searching for node paths breaks because the separator `.` is a token that gets chopped up by ElasticSearch.

Consider
* Using unique identifiers for the OntologyTerms in the NodePath instead of the 0[1] stuff, so that they're unique
* create a bigger AND AND AND query for each parent node, then the tokenizing shouldn't matter
* Fixing this in a separate story

dennishendriksen:
@fdlk @ChaoPang What is the status of this PR? What is required in order to merge it? When is this expected to happen?

fdlk:
I unassigned myself cause the story is not in the sprint.

dennishendriksen:
Reopen once we start working on the story.

PR #5323, opened 2016-10-10
Update version to 2.0.0-BBMRI_DIRECTORY-SNAPSHOT

Mark-de-Haan:


Mark-de-Haan:
It says checks failed, but Jenkins says all tests passed

PR #5336, opened 2016-10-11
Hackathon/directory negotiator

jjettenn:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


molgenis-bbmri-directory/src/main/java/org/molgenis/bbmri/directory/controller/DirectoryController.java
```diff
@@ -69,6 +71,7 @@ public String init(Model model)
 		headers.set("Authorization", "Basic bW9sZ2VuaXM6Z29nb2dv");
```
Mark-de-Haan: This should be in the settings as well

PR #5342, opened 2016-10-12
Set backend field type to enum 

tommydeboer:


dennishendriksen:
@tommydeboer can you set the default repository collection as default value for the backend attribute?

PR #5343, opened 2016-10-13
Refactor entity.attribute mref --> one-to-many

dennishendriksen:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


molgenis-data/src/main/java/org/molgenis/data/meta/model/Attribute.java
```diff
@@ -135,6 +138,76 @@ public Attribute setName(String name)
...
+		return getInt(SEQUENCE_NR);
+	}
+
+	public Attribute setSequenceNumber(int seqNr)
+	{
+		set(SEQUENCE_NR, seqNr);
+		return this;
+	}
+	
+	// FIXME rename to getEntityType
```
tommydeboer: Can you add sticky on the board or do this refactoring?

molgenis-data/src/main/java/org/molgenis/data/meta/model/Attribute.java
```diff
@@ -135,6 +138,76 @@ public Attribute setName(String name)
...
+	public EntityType getEntity()
+	{
+		return getEntity(ENTITY, EntityType.class);
+	}
+
+	// FIXME rename to setEntityType
```
tommydeboer: Can you add sticky on the board or do this refactoring?

molgenis-data/src/main/java/org/molgenis/data/meta/model/Attribute.java
```diff
@@ -224,15 +297,10 @@ public Attribute setDataType(AttributeType dataType)
 	 *
 	 * @return Iterable of attributes or empty Iterable if no attribute parts exist
 	 */
+	// FIXME rename to getChildren
```
tommydeboer: Can you add sticky on the board or do this refactoring?

tommydeboer:
Removing one of the attributes of an EntityType gives:
Error! entity [sys_md_Attribute] with id [AAAACV5564YDKGWXPB5UYUAAAI] does not exist

tommydeboer:
When removing a compound attribute from an EntityType:

Error! Could not commit JDBC transaction; nested exception is org.postgresql.util.PSQLException: ERROR: update or delete on table "sys_md_Attribute" violates foreign key constraint "sys_md_Attribute_parent_fkey" on table "sys_md_Attribute" Detail: Key (id)=(AAAACV5564YDGGWXPB5UYUAAAM) is still referenced from table "sys_md_Attribute".

dennishendriksen:
See meta model refactoring pull request

PR #5344, opened 2016-10-13
Fix #5288 Repository with one-to-many copy broken

### LINKED ISSUE #5288: Entities with an one-to-many attribute cannot be copied with the copy functionality.
#### Reproduce
Select a static entity with a one to many attribute try to copy it via the Data Explorer.

#### Expected
New copied entity

#### Observed
Error! ; nested exception is org.postgresql.util.PSQLException: ERROR: constraint "sys_Author_books_fkey" for relation "sys_Book_author" already exists

dennishendriksen:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


dennishendriksen:
See meta model refactoring pull request

PR #5347, opened 2016-10-14
Allow http access to ElasticSearch, we need it!

fdlk:


fdlk:
@mswertz I've tested on local VM and can confirm that setting elasticsearch property `network.host=localhost` makes ElasticSearch refuse connections from my desktop to my VM.

PR #5386, opened 2016-10-24
Fix Dependencies for DAS when using Jetty

### LINKED ISSUE #4926: Building MOLGENIS with empty local Maven repository fails
#### Reproduce
- Create a backup of your local Maven repository
- Empty your local Maven repository
- mvn clean install 
#### Expected
- Build succes
#### Observed
- Build failure
  `[ERROR] Failed to execute goal on project molgenis-das: Could not resolve dependencies for project org.molgenis:molgenis-das:jar:1.22.0-SNAPSHOT: Could not transfer artifact javax.jms:jms:jar:1.1 from/to java.net (https://maven-repository.dev.java.net/nonav/repository): Cannot access https://maven-repository.dev.java.net/nonav/repository with type legacy using the available connector factories: BasicRepositoryConnectorFactory: Cannot access https://maven-repository.dev.java.net/nonav/repository with type legacy using the available layout factories: Maven2RepositoryLayoutFactory: Unsupported repository layout legacy -> [Help 1]`

Mark-de-Haan:

### DIFF
--- a/molgenis-app/pom.xml
+++ b/molgenis-app/pom.xml
@@ -65,9 +65,9 @@
 						<version>1.1.3.4.O</version>
 					</dependency>
 					<dependency>
-						<artifactId>log4j</artifactId>
-						<groupId>log4j</groupId>
-						<version>1.2.17</version>
+						<groupId>org.slf4j</groupId>
+						<artifactId>log4j-over-slf4j</artifactId>
+						<version>1.7.21</version>
 					</dependency>
 				</dependencies>
 			</plugin>
--- a/molgenis-app/pom.xml
+++ b/molgenis-app/pom.xml
@@ -52,6 +52,23 @@
 						<artifactId>mysql-connector-java</artifactId>
 						<version>${mysql.connector.java.version}</version>
 					</dependency>
+					<dependency>
+						<groupId>uk.ac.ebi.mydas</groupId>
+						<artifactId>mydas</artifactId>
+						<version>1.6.7</version>
+						<scope>system</scope>
+						<systemPath>${basedir}/../molgenis-das/src/main/resources/mydas-1.6.7.jar</systemPath>
+					</dependency>
+					<dependency>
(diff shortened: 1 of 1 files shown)


molgenis-app/pom.xml
```diff
@@ -52,6 +52,23 @@
...
+						<version>1.6.7</version>
+						<scope>system</scope>
+						<systemPath>${basedir}/../molgenis-das/src/main/resources/mydas-1.6.7.jar</systemPath>
+					</dependency>
+					<dependency>
+						<groupId>xpp3</groupId>
+						<artifactId>xpp3</artifactId>
+						<version>1.1.3.4.O</version>
+					</dependency>
+					<dependency>
```
fdlk: I'd expect log4j-over-slf4j instead

dennishendriksen:
@fdlk Are you reviewing this pull request? You added comments without assigning yourself to this PR.

molgenis-app/pom.xml
Mark-de-Haan: Jetty does not run, stating it is missing the Log4j class dependency
I suspect this is caused due to MyDas being added here

dennishendriksen: Shouldn't the dependency be excluded instead?

Mark-de-Haan: Excluding the dependency will trigger a noClassDefFound exception

Mark-de-Haan: The problem is, is that tomcat uses the created WAR, which contains all the correct dependencies.

Jetty does not use the war, but some fancy regex to collect all the dependencies. MyDas is not included in this dependency grab. This is why I added it here explicitly

Mark-de-Haan:
@fdlk or @dennishendriksen ? It does not impact our own way of working seeing as I only add 3 dependencies to get Jetty running.

The WAR we deploy to servers is not affected

PR #5387, opened 2016-10-24
Fix Dependencies for DAS when using Jetty

Mark-de-Haan:


molgenis-app/pom.xml
```diff
@@ -9,56 +9,74 @@
...
+						<systemPath>${basedir}/../molgenis-das/src/main/resources/mydas-1.6.7.jar</systemPath>
+					</dependency>
+					<dependency>
+						<groupId>xpp3</groupId>
+						<artifactId>xpp3</artifactId>
+						<version>1.1.3.4.O</version>
+					</dependency>
+					<dependency>
+						<groupId>org.slf4j</groupId>
+						<artifactId>log4j-over-slf4j</artifactId>
```
dennishendriksen: MOLGENIS uses Logback instead of log4j. Why is this dependency required?

fdlk:
Looks like this commit already made its way to the master in another pull request?

PR #5390, opened 2016-10-25
Reduce test run logging output from 20MB -> 2MB

dennishendriksen:
#### Checklist
- [x] Unit tests


molgenis-app/src/main/resources/logback.xml
```diff
@@ -9,7 +9,4 @@
 	<root level="WARN">
 		<appender-ref ref="STDOUT" />
 	</root>
-	<logger name="org.molgenis.data.jobs.model.JobExecution" level="ALL">
```
bartcharbon: this logger is used to fill the "log" attribute in the "Annotation job execution (sys_AnnotationJobExecution)". Removing it will result in an empty column after a job execution.

dennishendriksen: Reverted removed logger tag

bartcharbon:
test this please

PR #5394, opened 2016-10-26
Allow http access, we need it!

fdlk:
Tested on local VM and can confirm that setting elasticsearch property network.host=localhost makes ElasticSearch refuse connections from my desktop to my VM.

New PR cause I need my master branch even more than http access to elasticsearch :)


Mark-de-Haan:
Closing, if needed, we can restart tomcat with a JVM option to enable elastic search debugging.

fdlk:
@Mark-de-Haan do you think that's the better way to do this?

PR #5414, opened 2016-10-28
M4479 Tag metadata changes

dennishendriksen:
- Rename class
- Rename entity name
- Rename attribute identifier -> id

#### Checklist
- [x] Unit tests


Mark-de-Haan:
When I try to run the app, I get the error:

`Updating entity [sys_md_Package] attribute [tags] referenced entity from [sys_md_tags] to [sys_md_Tag] not allowed for type [MREF]`

PR #5429, opened 2016-10-31
Fix #5189 Transactionality of PostgreSqlRepository is incorrect in places

### LINKED ISSUE #5189: Transactionality of PostgreSqlRepository is incorrect in places
Code inspection:
- findAll(Query) creates a Stream<Entity> within a ReadOnly transaction but the actual queries get executed when the stream is consumed.
- Plenty FIXMEs for commented out @Transactional annotations on inserts, updates and deletions
- Batched updates should probably execute in one single transaction.

dennishendriksen:
#### Checklist
- [x] Unit tests


bartcharbon:
the field: private final PlatformTransactionManager transactionManager; in the PostgreSqlRepositoryCollection is not used anymore.

PR #5474, opened 2016-11-04
Fix #5473 and #5479 delete package with entities

### LINKED ISSUE #5473: Delete package with multiple entities results in an error
#### Reproduce
Load data with multiple entities (some entities depends others)
Try to delete from package forms (Data Explorer)

#### Expected
packages and entities are deleted

#### Observed
packages are not deleted

jjettenn:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan


molgenis-data/src/main/java/org/molgenis/data/meta/EntityTypeDependencyResolver.java
```diff
@@ -1,7 +1,13 @@
 package org.molgenis.data.meta;
 
+import autovalue.shaded.com.google$.common.collect.$Lists;
```
Mark-de-Haan: Wrong import, use google.common.collect

PR #5507, opened 2016-11-10
Improve validation expression error messages

tommydeboer:


molgenis-data-validation/src/main/java/org/molgenis/data/validation/EntityAttributesValidator.java
```diff
@@ -422,9 +422,10 @@ private ConstraintViolation createConstraintViolation(Entity entity, Attribute a
 			String message)
 	{
 		String dataValue = getDataValuesForType(entity, attribute).toString();
-		String fullMessage = format("Invalid %s value '%s' for attribute '%s' of entity '%s'.",
+		String fullMessage = format(
+				"Invalid [%s] value [%s] for attribute [%s] of entity label [%s] with type [%s].",
```
bartcharbon: remove "label" from this message.
"Invalid [%s] value [%s] for attribute [%s] of entity [%s] with type [%s]."

PR #5561, opened 2016-11-16
Fix #5536 Changing the datatype of xstring (TypeTest) to int results in wrong error message

### LINKED ISSUE #5536: Changing the datatype of xstring (TypeTest) to int results in wrong error message
#### Reproduce
LOAD: TypeTest
In the Attribute table, change the datatype of xstring (TypeTest) to int

#### Expected
Error! Value [xxx] of this entity attribute is not of type [INT or LONG].

#### Observed
exclamation-sign Attribute changes could not be saved. NumberFormatException For input string: "xstring"

tommydeboer:
#### Checklist
- [x] Unit tests

### DIFF
--- a/molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/AttributeValidator.java
+++ b/molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/AttributeValidator.java
@@ -190,7 +190,9 @@ public class AttributeValidator
 			}
 			catch (NumberFormatException e)
 			{
-				throw new MolgenisDataException(format("%s %s", e.getClass().getSimpleName(), e.getMessage()));
+				throw new MolgenisValidationException(new ConstraintViolation(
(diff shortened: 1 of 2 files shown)


molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/AttributeValidator.java
```diff
@@ -190,7 +190,9 @@ void validateDefaultValue(Attribute attr)
 			}
 			catch (NumberFormatException e)
 			{
-				throw new MolgenisDataException(format("%s %s", e.getClass().getSimpleName(), e.getMessage()));
+				throw new MolgenisValidationException(new ConstraintViolation(
+						format("Invalid default value [%s] for data type [%s]", value, attr.getDataType())
```
Mark-de-Haan: Default value has nothing to do with it right?

PR #5569, opened 2016-11-17
Fix #5527 Multiple attributes with same name in entity allowed

### LINKED ISSUE #5527: Multiple attributes with same name in entity allowed
#### Reproduce
In the Attribute table, create a nillable string attribute, "newString" 
1. Set the EntityType to TypeTest
2. Set the Sequence Order to 1

And create a new one

#### Expected
Error message

#### Observed
Both are created one is shown. Deleting one attribute and you cannot use the entity anymore

dennishendriksen:
#### Checklist
- [x] Unit tests


fdlk:
This fix breaks Cmdline annotators running in update mode, I think.

PR #5589, opened 2016-11-21
Refactor allowed type conversions. Add getter for valid ID attribute types

Mark-de-Haan:


molgenis-data/src/main/java/org/molgenis/data/support/AttributeUtils.java
```diff
@@ -25,31 +27,18 @@ public static String getI18nAttributeName(String attrName, String languageCode)
 	public static boolean isIdAttributeTypeAllowed(Attribute attr)
 	{
 		AttributeType attrType = attr.getDataType();
-		switch (attrType)
+		if (getValidIdAttributeTypes().contains(attrType))
```
tommydeboer: You can just do ```return getValidIdAttributeTypes().contains(attrType)``` here

molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/AttributeValidator.java
```diff
@@ -308,57 +309,87 @@ private static void validateOrderBy(Attribute attr, Sort orderBy)
...
+
+		DATA_TYPE_ALLOWED_TRANSITIONS
+				.put(STRING, EnumSet.of(BOOL, DATE, DATE_TIME, DECIMAL, INT, LONG, HTML, SCRIPT, TEXT, ENUM, COMPOUND));
+
+		DATA_TYPE_ALLOWED_TRANSITIONS
+				.put(TEXT, EnumSet.of(BOOL, DATE, DATE_TIME, DECIMAL, INT, LONG, HTML, SCRIPT, STRING, ENUM, COMPOUND));
+
+		DATA_TYPE_ALLOWED_TRANSITIONS.put(ENUM, EnumSet.of(STRING, INT, LONG, TEXT));
+
+		// FIXME Why let compound become something else?
```
tommydeboer: has this question been answered yet?

dennishendriksen:
Please rebase with molgenis/2.0 picking 1a8a917 to reduce history polution.

molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/AttributeValidator.java
```diff
@@ -308,57 +309,87 @@ private static void validateOrderBy(Attribute attr, Sort orderBy)
...
-		DATA_TYPE_DISALLOWED_TRANSITIONS
-				.put(TEXT, EnumSet.of(CATEGORICAL_MREF, MREF, ONE_TO_MANY, EMAIL, HYPERLINK, FILE));
-		DATA_TYPE_DISALLOWED_TRANSITIONS
-				.put(XREF, EnumSet.of(CATEGORICAL_MREF, MREF, ONE_TO_MANY, EMAIL, HYPERLINK, FILE));
+		DATA_TYPE_ALLOWED_TRANSITIONS = new EnumMap<>(AttributeType.class);
+		EnumSet<AttributeType> allowedIdAttributeTypes = getValidIdAttributeTypes();
+
+		// TRUE and FALSE can either be expressed in string or 0 and 1
+		// Postgres does not support boolean to bigint or double precision (LONG and DECIMAL)
+		DATA_TYPE_ALLOWED_TRANSITIONS.put(BOOL, EnumSet.of(STRING, TEXT, INT));
```
Mark-de-Haan: BOOL to DECIMAL will throw an error, why support it?
BOOl to ENUM does not make sense, why would you create an ENUM for the two options BOOL already supports?
BOOL to LONG will throw an error, why support it?

To convert to HTML and SCRIPT just feels as if we are doing it because they are text fields. You might as well convert to STRING and then convert to HTML or SCRIPT

molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/AttributeValidator.java
```diff
@@ -308,57 +309,87 @@ private static void validateOrderBy(Attribute attr, Sort orderBy)
...
+							currentDataType.toString(), newDataType.toString(), allowedDatatypes.toString()));
 		}
 	}
 
-	private static EnumMap<AttributeType, EnumSet<AttributeType>> DATA_TYPE_DISALLOWED_TRANSITIONS;
+	private static EnumMap<AttributeType, EnumSet<AttributeType>> DATA_TYPE_ALLOWED_TRANSITIONS;
 
 	static
 	{
 		// transitions to EMAIL and HYPERLINK not allowed because existing values not checked by PostgreSQL
```
Mark-de-Haan: You initially wrote this, I can remove the postgres specific part though

molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/AttributeValidator.java
```diff
@@ -308,57 +309,87 @@ private static void validateOrderBy(Attribute attr, Sort orderBy)
...
 		// transitions to CATEGORICAL_MREF and MREF not allowed because junction tables updated not implemented
```
dennishendriksen: The important thing here it not the comments (which should not refer to PostgreSQL, my bad) but: The transitions should be supported by molgenis-data-validation but blocked by molgenis-data-postgresql until we implement them.

molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/AttributeValidator.java
dennishendriksen: > BOOL to DECIMAL will throw an error, why support it?

If BOOL to DECIMAL throws an error for the PostgreSQL repository collection then this conversion should be blocked in molgenis-data-postgresql instead of molgenis-data-validation since other repository collections might convert true to 1.0 and false to 0.0.

> BOOl to ENUM does not make sense, why would you create an ENUM for the two options BOOL already supports?

We should not make assumptions on if it makes sense, only look at technical possibilities. In this case in can envision a use case where the data modeler wants to change an attribute from true false (modeled as BOOL) to true/false/maybe/don't know (modeled as ENUM) by first converting BOOL to ENUM and later adding enum options for maybe/don't know.

> BOOL to LONG will throw an error, why support it?

See comment about BOOL to DECIMAL

> You might as well convert to STRING and then convert to HTML or SCRIPT
Why use two steps if one step is sufficient?

PR #5590, opened 2016-11-21
Use the list of allowed ID attribute types to create validation expression

Mark-de-Haan:
Both this PR as well as #5589 use the added getter in the AttributeUtils class


dennishendriksen:
See related comments in other PR. Please rebase with molgenis/2.0 picking e3e5efd to reduce history polution.

PR #5599, opened 2016-11-22
Fix/5573 Sort Data type results in error

jjettenn:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan

### DIFF
--- a/molgenis-data/src/main/java/org/molgenis/data/meta/model/AttributeMetadata.java
+++ b/molgenis-data/src/main/java/org/molgenis/data/meta/model/AttributeMetadata.java
@@ -26,7 +26,7 @@ public class AttributeMetadata extends SystemEntityType
 	public static final String NAME = "name";
 	public static final String ENTITY = "entity";
 	public static final String SEQUENCE_NR = "sequenceNr";
-	public static final String TYPE = "type";
+	public static final String TYPE = "dataType";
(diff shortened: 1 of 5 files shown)


molgenis-data/src/main/java/org/molgenis/data/meta/model/AttributeMetadata.java
```diff
@@ -26,7 +26,7 @@
 	public static final String NAME = "name";
 	public static final String ENTITY = "entity";
 	public static final String SEQUENCE_NR = "sequenceNr";
-	public static final String TYPE = "type";
+	public static final String TYPE = "dataType";
```
dennishendriksen: I'm quiet sure ```type``` is used in a lot of freemarker templates and JavaScript files. Did you check if there are references that require updating?

dennishendriksen:
A lot of bugs fixed in 2.0.0 RC-1 were related to code changes similar to this one. We might consider not fixing this issue in 2.0.0 and applying the fix to 2.1.0.

Did you check if the same issue exists in the most recent version of Elasticsearch?

jjettenn:
It is now clear that it is the same issue as in the past: the field name should have the same mapping as in Elasticsearch.

jjettenn:
This issue will be implemented/solved in an future version of MOLGENIS

PR #5597, opened 2016-11-22
Refactor annotator code for snpEff/Gavin and commandline annotators + fix "zipped" files not working

bartcharbon:
#### Checklist
- [x] Unit tests
- [ ] Updated testplan

PR #5604, opened 2016-11-23
M4616: Update Gavin UI

fdlk:
I've merged with #5597.
We'd both added gavinJobExecution as a parameter to the GavinJob :)


fdlk:
Failing test is due to changes in other PR.

PR #5637, opened 2016-11-28
Fix #5606 REST v2 batch update single attribute error for non-string ids

### LINKED ISSUE #5606: Updating an Entity via de REST API V2 will not work with INT or LONG identifiers
### Reproduce
- Import TypeTest
- Perform the HTTP request: PUT http://molgenisXX.gcc.rug.nl/api/v2/org_molgenis_test_TypeTest/xbool
  - Headers: ```x-molgenis-token``` with ```<token>```
  - Content-Type: application/json
  - Body:
```
{
    "entities": [
        {
            "id": 1,
            "xbool": false
        },
        {
            "id": 2,
            "xbool": false
        }
    ]
}
```
### Expected
Update succeeds, TypeTest id is INT, and the id values in the json post are also INT

### Observed
`"message": "org.molgenis.data.MolgenisDataException: Attribute [id] query value is of type [String] instead of [Integer]"`

dennishendriksen:
#### Checklist
- [x] Unit tests

### DIFF
--- a/molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
+++ b/molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
@@ -113,9 +113,10 @@ class RestControllerV2
 		return new MolgenisDataException("Operation failed. Entities must provide only an identifier and a value");
 	}
 
-	static UnknownEntityException createUnknownEntityExceptionNotValidId(String id)
+	static UnknownEntityException createUnknownEntityExceptionNotValidId(Object id)
 	{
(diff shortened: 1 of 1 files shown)


Mark-de-Haan:
Using the advanced REST API, I login and with the token
I do a PUT with the following JSON:
```
{
  "entities": [
        {
            "id": 1,
            "xbool": false
        },
        {
            "id": 1,
            "xbool": false
        }
    ]
}
```
The error I receive is the following:
`"message": "Duplicate value '1, ref1' for unique attribute 'id, xcategoricalmref_value' from entity 'org_molgenis_test_TypeTest_xcategoricalmref_value'."`

PR #5662, opened 2016-11-30
M4616: When parsing VCF, if the ID is the empty string, replace it with “.”

fdlk:
#### Checklist
- [x] Unit tests
- [ ] Updated testplan


bartcharbon:
test this please

PR #5676, opened 2016-11-30
Fix #5655. Switch from guava to caffeine for the L2 cache.

### LINKED ISSUE #5655: L2 Cache contains second-last state of JobExecution entity
### How to Reproduce
Unsure. Run a lot of Gavin jobs.

### Expected behavior
L2Cache always contains the same version as the database.

### Observed behavior
Sometimes the gavin job as retrieved from the L2Cache still has status RUNNING and log ending with "File is ready for download." But in the database status has already been updated to FINISHED and log ends with "That took 20m12s."

fdlk:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan

### DIFF
--- a/molgenis-data-cache/pom.xml
+++ b/molgenis-data-cache/pom.xml
@@ -18,6 +18,16 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>2.3.5</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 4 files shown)


dennishendriksen:
Are L1, L2 and L3 cache are all using the new cache implementation now?

fdlk:
@dennishendriksen Right now this only changes the L2 cache implementation, I thought L3 wasn't loading but I see that it is so probably also suffers from the same bug. Will update that too. What do we want for the L1 cache? Switch to have everything use the same library? Or stick to guava cause that's the one we used in the manual tests?

dennishendriksen:
Switching to have everything use the same library is preferable.

PR #5680, opened 2016-12-01
Fix #5674 DATE and DATE_TIME persisted with too much detail in PostgreSQL

### LINKED ISSUE #5674: DATE and DATE_TIME persisted with too much detail in PostgreSQL
### How to Reproduce
1. Open data explorer, Token entity
2. Select a row to edit
3. Save without changing value
4. Save

### Expected behavior
Is saved

### Observed behavior
Error! The attribute 'creationDate' of entity 'sys_sec_Token' can not be changed it is readonly. (entity 1)

jjettenn:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan

### DIFF
--- a/molgenis-core/src/main/java/org/molgenis/util/MolgenisDateFormat.java
+++ b/molgenis-core/src/main/java/org/molgenis/util/MolgenisDateFormat.java
@@ -1,7 +1,11 @@
 package org.molgenis.util;
 
+import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.Date;
 import java.util.TimeZone;
+
+import static java.lang.String.format;
 
 public class MolgenisDateFormat
 {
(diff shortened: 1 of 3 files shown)


jjettenn:
test me please

jjettenn:
test this please

fdlk:
Couple remarks about this:

* If this change only aims to reduce the precision stored to seconds, then wouldn't it be better to do so in the table definition by specifying the column datatype as TIMESTAMP(0)?
* It's very inefficient to create a new SimpleDateFormatter for each time-valued cell we store
* Fiddling with Europe/Amsterdam timezone is a bit imperialistic. But worse, it means that the values will not be continuous around moments when daylight savings time starts or ends. Just wait for the index job that lasted -3600 seconds.
* Wouldn't molgenis compute type users actually enjoy millisecond precision?

dennishendriksen:
@fdlk Good remarks that we should definitely discuss. can you create a story in the backlog? Else your remarks might never be addressed.
@fdlk Do you think we should revert the fix?

fdlk:
@dennishendriksen I've created http://www.molgenis.org/ticket/4691 and no, let's not revert this.

PR #5686, opened 2016-12-01
Fix #5683 Fix #5685 Missing package validation checks

### LINKED ISSUE #5683: It is possible to add packages to system packages
### How to Reproduce
1. Go to Package in the data explorer
2. Add a new package
3. For parent, choose 'sys'
4. Submit form

### Expected behavior
Friendly message explaining it's not possible to add packages to system packages

### Observed behavior
Package is added to system package and thus can not be removed anymore.

### LINKED ISSUE #5685: Creating package with simpleName/name+parent mismatch allowed
### How to Reproduce
- Login as superuser
- Select the entity type ```Package``` in the ```Data explorer ``` plugin
- Select the '+' icon and enter the following values
  - Qualified name: parent_test
  - Name: test
- Select 'Create'

### Expected behavior
- Error that name/parent do not match with qualified name

### Observed behavior
- Package created successfully

dennishendriksen:
#### Checklist
- [ ] Unit tests
- [ ] Updated testplan

### DIFF
--- a/molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/PackageRepositoryValidationDecorator.java
+++ b/molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/PackageRepositoryValidationDecorator.java
@@ -1,0 +1,112 @@
+package org.molgenis.data.validation.meta;
+
+import org.molgenis.data.AbstractRepositoryDecorator;
+import org.molgenis.data.Repository;
+import org.molgenis.data.UnknownEntityException;
+import org.molgenis.data.meta.model.Package;
+
+import java.util.stream.Stream;
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+/**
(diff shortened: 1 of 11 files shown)


molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/PackageRepositoryValidationDecorator.java
```diff
@@ -0,0 +1,112 @@
...
+
+	@Override
+	public void deleteById(Object id)
+	{
+		Package package_ = findOneById(id);
+		if (package_ == null)
+		{
+			throw new UnknownEntityException(format("Unknown package [%s]", id.toString()));
+		}
+		packageValidator.validate(package_);
```
dennishendriksen: I agree with renaming, but I would like to keep it in sync with EntityType/Attribute/Tag validators and have some ideas on reusing validation infrastructure between these entities. I propose to address this after the 2.0 release.

PR #5708, opened 2016-12-08
Make mail settings configurable through a settings entity

fdlk:
#### Checklist
- [x] Unit tests
- [ ] Updated testplan

Use MailSender interface to make the mail interaction more testable.
Adds tests for the mails that get sent.
Makes MailSender available through a Supplier<MailSender> so that changes in the settings are reflected in the sender.
Moves generation of secure tokens to a SecureIdGenerator class.


molgenis-core-ui/src/main/java/org/molgenis/ui/controller/FeedbackController.java
```diff
@@ -51,17 +51,17 @@
 	private final UserService userService;
 	private final AppSettings appSettings;
 	private final CaptchaService captchaService;
-	private final JavaMailSender mailSender;
+	private final Supplier<MailSender> mailSenderSupplier;
```
dennishendriksen: What do you think about hiding the Supplier behind our own JavaMailSender implementation?
- Easier to use, easier to change
- Requires less changes in existing code
- Expose JavaMailSender bean instead of MailSender such that we can build messages with attachments, inline resources etc.

molgenis-data-elasticsearch/pom.xml
```diff
@@ -55,5 +55,10 @@
             <artifactId>protonpack</artifactId>
             <version>1.8</version>
         </dependency>
+        <dependency>
+            <groupId>org.molgenis</groupId>
+            <artifactId>molgenis-mail</artifactId>
```
dennishendriksen: The mail module will be a very small Maven module. Maybe this is the time to introduce molgenis-utils and add the mail package there? Afterward we can migrate and ditch molgenis-core.

molgenis-gavin/src/main/java/org/molgenis/gavin/controller/GavinController.java
```diff
@@ -78,7 +80,7 @@ public GavinController(@Qualifier("gavinExecutors") ExecutorService executorServ
 		this.fileStore = requireNonNull(fileStore);
 		this.userAccountService = requireNonNull(userAccountService);
 		this.menuReaderService = menuReaderService;
-		secureIdGenerator = new SecureIdGenerator();
+		this.secureIdGenerator = secureIdGenerator;
```
dennishendriksen: requireNonNull(secureIdGenerator)

molgenis-mail/src/main/java/org/molgenis/mail/MailSettings.java
```diff
@@ -0,0 +1,20 @@
+package org.molgenis.mail;
+
+interface MailSettings
+{
+	String getHost();
```
dennishendriksen: Are we missing out on settings that would require configuring a sendmail server on CentOS?

molgenis-security-core/src/main/java/org/molgenis/security/core/SecureIdGenerator.java
```diff
@@ -0,0 +1,19 @@
+package org.molgenis.security.core;
+
+public interface SecureIdGenerator
+{
+	/**
+	 * Generates a secure identifier.
```
dennishendriksen: Can you document what makes an identifier secure? Should a secure identifier be unique?

pom.xml
```diff
@@ -708,5 +708,6 @@
         <module>molgenis-data-cache</module>
         <module>molgenis-data-platform</module>
         <module>molgenis-annotators-cmd</module>
+		<module>molgenis-mail</module>
```
dennishendriksen: This PR will silently change your working mail server to a non-working mail server since username/password are not migrated from molgenis-server.properties. As an admin you might not notice this until users start telling you that they cannot register, change password etc.

How about extending MailSettings with a class that returns the corresponding property from molgenis-server.properties in case the value in the database is null? We could ditch this class later when we have a way to properly initialize the system.

molgenis-core-ui/src/main/java/org/molgenis/ui/controller/FeedbackController.java
fdlk: MailSender >> JavaMailSender cause it's way easier to test. I checked and we do not use JavaMailSender capabilities in any of our mail interactions right now so I'd like to switch to the simpler interface.

Will make the bean adhere to MailSender interface and hide the dynamic generation behaviour behind it.

molgenis-mail/src/main/java/org/molgenis/mail/MailSettings.java
fdlk: You mean, can we configure a running CentOS sendmail server using these parameters?

Mark-de-Haan:
You might want to merge with master because of some UUID changes that are present in this PR

fdlk:
@Mark-de-Haan Already is merged with master.

molgenis-data-elasticsearch/pom.xml
fdlk: elasticsearch no longer needs to depend on molgenis-mail, so I've removed these lines
molgenis-mail still is a rather small module, it depends on molgenis-data so is a bit higher than what I'd expect in molgenis-util or molgenis-core
Could move the mail classes to molgenis-data but that's already a bit beefy class.
Or start a molgenis-settings package. Unsure.

PR #5712, opened 2016-12-12
Revert "Exclude GavinJobExecution entity from caching."

fdlk:
Reverts workaround for #5655 which got properly fixed in #5676

### DIFF
--- a/molgenis-data-cache/src/main/java/org/molgenis/data/cache/l2/L2CacheRepositoryDecorator.java
+++ b/molgenis-data-cache/src/main/java/org/molgenis/data/cache/l2/L2CacheRepositoryDecorator.java
@@ -13,6 +13,7 @@ import java.util.Map;
 import java.util.stream.Stream;
 
 import static com.google.common.collect.Iterators.partition;
+import static com.google.common.collect.Lists.newArrayList;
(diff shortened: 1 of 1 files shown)


molgenis-data-cache/src/main/java/org/molgenis/data/cache/l2/L2CacheRepositoryDecorator.java
```diff
@@ -48,8 +49,7 @@ public L2CacheRepositoryDecorator(Repository<Entity> decoratedRepository, L2Cach
 	{
 		this.decoratedRepository = requireNonNull(decoratedRepository);
 		this.l2Cache = requireNonNull(l2Cache);
-		this.cacheable = decoratedRepository.getCapabilities().contains(CACHEABLE) && !"sys_idx_GavinJobExecution"
-				.equals(decoratedRepository.getName());
+		this.cacheable = decoratedRepository.getCapabilities().containsAll(newArrayList(CACHEABLE));
```
fdlk: Just reverting the workaround.

dennishendriksen: Nah, not a big deal.

PR #5746, opened 2017-01-05
[FEATURE] RSQL parser for javascript

fdlk:
#### Checklist
- [x] Unit tests
- [ ] Updated testplan


fdlk:
test this please

molgenis-core-ui/src/test/javascript/modules/rest-client/rsql/transformerTest.js
```diff
@@ -0,0 +1,219 @@
...
+    toComplexRef,
+    toComplexLine
+} from "rest-client/rsql/transformer";
+import test from "tape";
+
+test('Test that a single comparison gets mapped', assert => {
+    const actual = groupBySelector(parser.parse("xbool==false"))
+    const expected = {
+        "xbool": {
+            "selector": "xbool",
```
Mark-de-Haan: Shouldnt there be a type here as well? like RANGE or SIMPLE_REF
Else I still need to manually check for types and / or operators

molgenis-core-ui/src/main/javascript/modules/rest-client/rsql/transformer.js
```diff
@@ -0,0 +1,84 @@
+/**
+ * Transforms parsed filter RSQL to a map
+ */
+function groupBySelector(tree) {
+    return (tree.operands || [tree]).reduce(combine, {})
```
Mark-de-Haan: (string=q=1,string=q=2) results in a  model with only string=q=2

PR #5762, opened 2017-01-19
M4678 Mail server settings viewable/editable in UI

fdlk:
* (feature) Test connection on save, but only if username and password are provided.
* (feature) Use ONE_TO_MANY between MailSettings and JavaMailProperties to show the link.
* (refactor) Don’t create specific mail module. Create settings module and util module instead.

#### Checklist
- [x] Unit tests
- [ ] Updated testplan


fdlk:
test this please

PR #5768, opened 2017-01-23
Fix #5757 Error message is switched when trying to delete ref entity

### LINKED ISSUE #5757: Error message is switched when trying to delete ref entity
### How to Reproduce
Upload the demo dataset
Try to delete gender
You receive this message: 
screen shot 2017-01-17 at 15 16 04
But what it should say is that it could not delete gender because another entity depends on it. 
### Expected behavior
cannot delete gender because another entity depends on it
### Observed behavior
screen shot 2017-01-17 at 15 16 04

dennishendriksen:
#### Checklist
- [x] Unit tests

### DIFF
--- a/molgenis-data-postgresql/src/test/java/org/molgenis/data/postgresql/PostgreSqlExceptionTranslatorTest.java
+++ b/molgenis-data-postgresql/src/test/java/org/molgenis/data/postgresql/PostgreSqlExceptionTranslatorTest.java
@@ -84,7 +84,7 @@ public class PostgreSqlExceptionTranslatorTest
 		//noinspection ThrowableResultOfMethodCallIgnored
 		MolgenisValidationException e = postgreSqlExceptionTranslator
 				.translateDependentObjectsStillExist(new PSQLException(serverErrorMessage));
-		assertEquals(e.getMessage(), "Cannot delete entity 'myEntity' because entity 'myRefEntity' depends on it.");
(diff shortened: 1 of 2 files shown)


molgenis-data-postgresql/src/test/java/org/molgenis/data/postgresql/PostgreSqlExceptionTranslatorTest.java
```diff
@@ -111,7 +111,7 @@ public void translateDependentObjectsStillExistMultipleDependentTables()
 		MolgenisValidationException e = postgreSqlExceptionTranslator
 				.translateDependentObjectsStillExist(new PSQLException(serverErrorMessage));
 		assertEquals(e.getMessage(),
-				"Cannot delete entity 'myEntity' because entities 'myRefEntity, myOtherRefEntity' depend on it.");
+				"Cannot delete entity 'myRefEntity' because entity 'myEntity' depends on it..Cannot delete entity 'myOtherRefEntity' because entity 'myEntity' depends on it.");
```
bartcharbon: the 2 dots in the middle should be a dot and a space or a comma I think.

dennishendriksen: This is how the message of a MolgenisValidationException with two constraint violations with messages `xxxxxxx.` and `yyyyyyy.` is constructed.

PR #5796, opened 2017-02-03
Fix #4827

### LINKED ISSUE #4827: Upper case VCF extension fails
#### Reproduce
- try to upload a vcf with an upper case extension
#### Expected
- this works
#### Observed
- an illegal argument exception is thrown.

marikaris:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


fdlk:
When I select file test.vcf, the proposed entity name is test.
When I select file test.VCF, the proposed entity name is test.VCF.
Unsure what code path actually does this but seems like we missed a place where this transformation is done.

Issue #5814, opened 2017-02-10
Genome browser produces 500 error when processing an abstract entity


marikaris:
### How to Reproduce
Load genomic data with an abstract entity (like this one: 
[infever_model_test.xlsx](https://github.com/molgenis/molgenis/files/844409/infever_model_test.xlsx)

Go to data explorer and select lab1

### Expected behavior
Lab_model is an abstract entity so it is not shown

### Observed behavior
<img width="941" alt="screen shot 2017-02-10 at 09 51 49" src="https://cloud.githubusercontent.com/assets/10075439/22820369/acf1a482-ef76-11e6-9e3f-03d1dc9fcc03.png">

PR #5816, opened 2017-02-10
Fix/4952 DAS datasource selects first 1000? hits from a chromosome without looking at the position

marikaris:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

Still a nullpointer when having an abstract entity with genomic data, was already in there before fixing this issue, reported: #5814


molgenis-dataexplorer/src/main/resources/js/dataexplorer-directory.js
```diff
@@ -70,6 +70,12 @@ $.when($,
 
     $(function () {
         $('#directory-export-button').on('click', function () {
+            if( !molgenis.dataexplorer.getRSQL() ) {
+                // no filters selected yet
+                bootbox.alert(i18n.dataexplorer_directory_export_no_filters);
+                return;
+            }
```
bartcharbon: this doesn't seem to be related, please remove from PR

bartcharbon:
please remove the i18n.xlsx as well

PR #5831, opened 2017-02-17
Fix #5824 JavaScript date return value cannot be converted

### LINKED ISSUE #5824: JavaScript date return value cannot be converted
### How to Reproduce
load a dataset with a date as string in it, map it to a model which requires a date

### Expected behavior
I can map my string to an actual date using Date.parse() or Date.UTC() (of course using the right format, but even a copy of the ISO 8601 docs ("1994-11-05T08:15:30-05:00") in Date.parse does not result in a valid date. 

### Observed behavior
Error! (I think the function does not exist, it also gets slightly greyer than existing functions like Date.now())

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-js/src/main/java/org/molgenis/js/nashorn/NashornScriptEngine.java
+++ b/molgenis-js/src/main/java/org/molgenis/js/nashorn/NashornScriptEngine.java
@@ -9,8 +9,12 @@ import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 import javax.script.*;
+import javax.xml.bind.DatatypeConverter;
 import java.io.IOException;
+import java.util.Calendar;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static java.util.Arrays.asList;
 
@@ -20,6 +24,7 @@ public class NashornScriptEngine
 	private static final Logger LOG = LoggerFactory.getLogger(NashornScriptEngine.class);
 
 	private static final List<String> RESOURCE_NAMES;
+	private static Pattern PATTERN_NATIVE_DATE = Pattern.compile("\\[Date (.*?)\\]");
 
 	static
 	{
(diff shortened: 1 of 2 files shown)


molgenis-js/src/main/java/org/molgenis/js/nashorn/NashornScriptEngine.java
```diff
@@ -114,7 +120,21 @@ private static Object convertNashornValue(Object nashornValue)
 			}
 			else
 			{
-				throw new RuntimeException("Unable to convert [ScriptObjectMirror]");
+				// context: https://github.com/molgenis/molgenis/issues/5824
+				// returning new Date() in JavaScript results in a ScriptObjectMirror without values that wraps a
```
Mark-de-Haan: If you use a Date.now() in JS, it will work without this fix

marikaris:
This works for date.now, but there is no way to put any other date attribute in your code. Functions like Date.parse/Date.UTC do not work

fdlk:
They should work if you call getTime() on the date.

fdlk:
But I think this is a nice feature.

marikaris:
Then I also get the current time, I just want to put another date and time in a Date type. I cannot find a way to do that without this fix.

fdlk:
I am pretty certain calling getTime() on the JavaScript Date instance will give you the epoch value for that Date instance, not the value of Date.now()  
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime

marikaris:
Ah I see, yes it worked, I did it wrong. Then I found this link https://www.w3schools.com/jsref/jsref_gettime.asp which made me think it was only to get the current time. 

But besides all of this. I think it is not very intuitive to not be able to just call the new Date the way we were used to in the previous version of molgenis. And while fixing this bug, we saw that the date actually gets passed to the javascriptconverter of java, but returns nothing, while the date and time are in the toString method, which seems odd.

fdlk:
Yes this is a nice feature that will make working with dates easier.

dennishendriksen:
The existing behavior broke when moving switching JavaScript engines: thus this was a bug fix, not a feature.

PR #5851, opened 2017-02-23
Fix #5832 Import in update mode silently ignores non-existing entities

### LINKED ISSUE #5832: Import data in update mode silently ignores non-existing entities
### How to Reproduce
Upload a dataset, then alter the data slightly and add one or more id's. Upload again with the "Update" function. 

### Expected behavior
The upload fails because new IDs are present in your dataset and you only wanted to update (the way it worked in MOLGENIS 1.0), or the upload succeeds and I receive a message telling me which IDs did not update because they are not in the data yet. 

### Observed behavior
The data updates silently not telling me about my not matching IDs (this is needed when for instance updating variants for VKGL and E-rare, sometimes the same variants are specified by slightly different IDs and then I want to know which IDs I should alter or contact someone about)

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


bartcharbon:
test this please

PR #5852, opened 2017-02-24
Fix #5785 When mapping to new table, the new table is not added to the base package when no package is given

### LINKED ISSUE #5785: When mapping to new table, the new table is not added to the base package when no package is given
### How to Reproduce
Map to a table with the name test

### Expected behavior
Error! Specify the package or it succeeds and table base_test is created

### Observed behavior
table test outside of any package is created

marikaris:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
+++ b/molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
@@ -11,7 +11,6 @@ import org.molgenis.data.mapper.service.AlgorithmService;
 import org.molgenis.data.mapper.service.MappingService;
 import org.molgenis.data.meta.AttributeType;
 import org.molgenis.data.meta.DefaultPackage;
-import org.molgenis.data.meta.IdentifierLookupService;
 import org.molgenis.data.meta.model.Attribute;
 import org.molgenis.data.meta.model.AttributeFactory;
 import org.molgenis.data.meta.model.EntityType;
--- a/molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
+++ b/molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
@@ -185,19 +185,20 @@ public class MappingServiceImpl implements MappingService
 		}
 
 		Repository<Entity> targetRepo;
-		if (!dataService.hasRepository(entityName) && !dataService.hasRepository("base_"+entityName))
+		if (!dataService.hasRepository(entityName) && !dataService.hasRepository(defaultPackage.PACKAGE_DEFAULT+"_"+entityName))
 		{
 			targetMetaData.setPackage(defaultPackage);
 			// Create a new repository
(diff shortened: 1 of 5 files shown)


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
```diff
@@ -183,14 +185,22 @@ public String applyMappings(MappingTarget mappingTarget, String entityName, bool
...
+		}else
 		{
+			targetMetaData.setPackage(null);
+			//if the entity is in the base package, this should be added to the name
+			if(dataService.hasRepository("base_"+entityName))
+			{
+				entityName = "base_" + entityName;
+				targetMetaData.setName(entityName);
+				targetMetaData.setLabel(entityName);
+			}
```
Mark-de-Haan: This part is a bit weird.. So if I Have a repository base_test, I remove the base_ and then check if I have base_test. I already know I have base_test, why remove and add the package again?

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
```diff
@@ -173,7 +176,6 @@ public String applyMappings(MappingTarget mappingTarget, String entityName)
 	public String applyMappings(MappingTarget mappingTarget, String entityName, boolean addSourceAttribute)
 	{
 		EntityType targetMetaData = EntityType.newInstance(mappingTarget.getTarget(), DEEP_COPY_ATTRS, attrMetaFactory);
-		targetMetaData.setPackage(null);
```
Mark-de-Haan: So if we always want the targetMetadata to be in the base package,
Why not do `setPackage(DefaultPackage)` ?

marikaris: Because sometimes the given target entity is already existing and in another package. Without setting this setPackage to null, the package name is added twice for some reason.

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
```diff
@@ -183,14 +185,22 @@ public String applyMappings(MappingTarget mappingTarget, String entityName, bool
 		}
 
 		Repository<Entity> targetRepo;
-		if (!dataService.hasRepository(entityName))
+		if (!dataService.hasRepository(entityName) && !dataService.hasRepository("base_"+entityName))
```
marikaris: no because an entity can be in another package as well, it does not have to be in base

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
marikaris: if the entity was already created before in the base package, its entityname did not contain "base_", so I had to add it, but only if the entity was in base, otherwise it would put entities in another package in base as well.

marikaris:
Better fix will be here soon

PR #5897, opened 2017-03-08
M4597 App manager plugin

dennishendriksen:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


marikaris:
I had one failing row in the testplan and one with a logo not loading properly: 
[test-apps.xlsx](https://github.com/molgenis/molgenis/files/833248/test-apps.xlsx)
The feature is really cool!
Could you maybe add the plugin to the plugins menu by default? Otherwise people may not know of it's existence.

dennishendriksen:
The plugin is added to the plugins menu by default when you start the application with an empty database, see the changes in molgenis-app\src\main\resources\molgenis_ui.xml.

dennishendriksen:
> I had one failing row in the testplan

Error in the testplan, I updated the template

> with a logo not loading properly

Cannot reproduce, can you check that your icon URL is ```/img/_logo_molgenis_old.gif```?

> some comments on the "After" part

These tables can include rows that are not related to the apps test suite, so only relevant rows should be removed, so I propose to keep it like this.

PR #5903, opened 2017-03-08
Fix #5885 rest returns all data when 'num' is set to '0'

### LINKED ISSUE #5885: REST API v1 and v2 will return all data when specifying num=0
### How to Reproduce
curl https://directory.bbmri-eric.eu/api/v2/eu_bbmri_eric_collections?num=0

### Expected behavior
Only the metadata is returned and no items, nextHref is set to get the first batch of items.

### Observed behavior
All data is returned.

ConnorStroomberg:
Update the RestController to skip fetching the entities when page size is set to zero.

issue: https://github.com/molgenis/molgenis/issues/5885

Add test to check zero branch in createEntityCollectionResponse method of RestControllerV2 class. and add jsonpath dependency to to the rest pom for use in test. Jsonpath can be used to check the json responce body.

example: 

http://localhost:8080/api/v2/sys_md_Attribute?num=0

used to return all entities but now returns zero entities. result does still contain metadata en count result.

#### Checklist
- [x] Code reviewed
- [ ] Clean commits


ConnorStroomberg:
I now see this issue needs to be fixed for rest api v1 as well. In rest api v1 pagesize is set to default if num size is set to "0". Maybe v2 needs the save behaviour ? I will add a comment on the issue.

ConnorStroomberg:
After discussing this with @tommydeboer we decided the behaviour as proposed in the merge request is fine. This means v2-api will return zero entities and meta data on setting 'num=0'. V1-api still returns the default number of entities ( set to 100 at this moment).

PR #5929, opened 2017-03-14
M4307 As col7a1 I need a new solution for homozygous patients in the dataexplorer

bartcharbon:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


ConnorStroomberg:
deployed at: https://molgenis42.gcc.rug.nl -> Patients

ConnorStroomberg:
Functionality works & meets specifications:
- [x] duplicate mutation in table
- [x] only four columns of these mutations are shown (Mutation,Protein Change,Consequence,Exon)
- [x] sorting is disabled for the mutation column
- [x] pubmed is an xref, but it acts like a hyperlink to pubmed.
- [x] other x/mrefs in the table work the same as the standard molgenis table

ConnorStroomberg:
"COL7A1Table.js" shares most if its code with "Table.js' this makes it hard to maintain. A posible solution could be to Make Table.js more configurable or the Extend/Use Talbe.js in COL7A1Table.js.

bartcharbon:
I think it is already unfortunate that we have to add this table to the molgenis repo. 

I would not change the default table to support 1 very specific usecase. The current solution does not have any interactions with the rest of MOLGENIS, I think that is the less 'dirty' solution.

fdlk:
Can still filter on the mutations, but only the first mutation gets filtered.

bartcharbon:
@fdlk: I 'fixed'  this in the entity report. 
It is kinda dirty but I removed the "Mutations" from the tree, and the wizard button is not shown in anymore for the patients.

https://molgenis42.gcc.rug.nl/menu/main/dataexplorer?entity=col7a1_Patients&hideselect=true

PR #5932, opened 2017-03-15
Added GitBook documentation (without postgresql)

Mark-de-Haan:
#### Checklist
- [ ] Functionality works & meets specifications
- [ ] Code reviewed
- [ ] Code unit/integration/system tested
- [ ] User documentation updated
- [ ] Test plan template updated
- [ ] Clean commits

PR #5936, opened 2017-03-16
Add link to 1.21.x documentation

Mark-de-Haan:
Test this after merging #5932

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -13,4 +13,4 @@ If you are a researcher, a (bio)informatician, a biomedical practitioner, or any
 
 ### Documentation for previous versions
 
-[MOLGENIS v1.21.x](link TODO)
+[MOLGENIS v1.21.x](https://molgenis.gitbooks.io/molgenis/content/v/1.21/)


docs/README.md
```diff
@@ -13,4 +13,4 @@ If you are a researcher, a (bio)informatician, a biomedical practitioner, or any
 
 ### Documentation for previous versions
 
-[MOLGENIS v1.21.x](link TODO)
+[MOLGENIS v1.21.x](https://molgenis.gitbooks.io/molgenis/content/v/1.21/)
```
fdlk: Can the link be relative?

Mark-de-Haan: No because you are pointing to the content of another branch

fdlk: Eeew. Let's use https://plugins.gitbook.com/plugin/versions instead then?

Mark-de-Haan: URL will never change, why fancy if can be easy and straightforward?

fdlk: User experience is better
* clear on each documentation page what version of the documentation you're browsing
* ability to go back and forward between versions

Developer experience is better
* (once set up) is easier to maintain cause all versions show same list from single source of truth

I've put up pull request to use it, can you take a look?

PR #5963, opened 2017-03-23
Fix #5329 VCF invalid due to missing "." for multi-allelic missing values

### LINKED ISSUE #5329: VCF output invalid due to missing "." for multi-allelic missing values
Multi allelic values with annotations may be read in like this:

```
1   54728   .   G   T,C .   .   GoNL_AF=0.01,.;CADD_SCALED=34.0,.;
```

Where GoNL and CADD annotations for the 'C' allele are missing.

But the VCF writer outputs is as:

```
1   54728   .   G   T,C .   .   GoNL_AF=0.01,;CADD_SCALED=34.0,;
```

Which is NOT valid VCF, and loading this file with VCF reader will fail because only 1 value is seen.

Note that this only happens when values are typed correctly (as Float). Then they are parsed as "Object val = [34.0, null]" where String values are parsed as "Object val = [0.01, .]".

dennishendriksen:
Update Genotype-IO 1.0.1 to 1.0.3

#### Checklist
- [x] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/format/VcfToEntity.java
+++ b/molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/format/VcfToEntity.java
@@ -32,6 +32,7 @@ import static com.google.common.collect.Iterables.size;
 import static com.google.common.collect.Maps.newHashMapWithExpectedSize;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toSet;
 import static java.util.stream.StreamSupport.stream;
 import static org.molgenis.data.meta.AttributeType.*;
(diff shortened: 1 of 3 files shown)


molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/format/VcfToEntity.java
```diff
@@ -363,8 +363,11 @@ private void writeInfoFieldsToEntity(VcfRecord vcfRecord, Entity entity)
 				}
 				else if (vcfInfoVal instanceof List<?>)
 				{
+					List<?> vcfInfoValTokens = (List<?>) vcfInfoVal;
 					// TODO Use list data type once available (see http://www.molgenis.org/ticket/2681)
```
ConnorStroomberg: is this todo still relevant ?

dennishendriksen: Yes

PR #6008, opened 2017-04-06
Fix #5785 When mapping to new table, the new table is not added to the base package when no package is given

### LINKED ISSUE #5785: When mapping to new table, the new table is not added to the base package when no package is given

marikaris:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
```diff
@@ -205,6 +216,7 @@ public String applyMappings(MappingTarget mappingTarget, String entityName, bool
 			}
 		}
 
+
```
Mark-de-Haan: Run formatting on your file please

PR #6017, opened 2017-04-10
M4934 Metadata model changes

Mark-de-Haan:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


fdlk:
Acceptance criteria have been added to the story:

> FOR THIS TO HAVE VALUE WE ALSO NEED:
upload using ID
download including ID

ConnorStroomberg:
When id try to use the download tool to download a entity with meta data, if fails

`  java -jar downloader-1.1.jar -f downloaded.xlsx -u http://localhost:8080/ -a admin -p admin -m it_emx_datatypes_TypeTest`

This could be because there is no more 'name' field, is this acceptable ?

fdlk:
Moving entities or packages will result in breaking changes because the entity will still have the ID it was originally uploaded with. How are we going to fix this in the future and won't it be better to disable moving entities until that time?

Mark-de-Haan:
test this please

PR #6020, opened 2017-04-11
Fix #5988 No error when validating mapping of 'null' values to not nillable mref

### LINKED ISSUE #5988: No error when validating mapping of 'null' values to not nillable mref
### How to Reproduce
Use 3.0 RD connect
Load RD connect data in drive folder: rd connect marieke and mariska. In order of read me in that directory (at molgenis80 28-03-2017)
Alter mapping of UCL disease to: 
`var prefix = 'urn:miriam:orphanet:Orphanet_'
var code = $('Diagnosis_Orphanet_Code').value()
[prefix+code]`
Click validate expression

### Expected behavior
Validation fails because there are empty cells in source

### Observed behavior
Validation succeeds but mapping fails with: 
ClassCastException: null
No stacktrace

Mark-de-Haan:
#### Checklist
- [x] Code reviewed
- [ ] Clean commits

[test_5988.xlsx](https://github.com/molgenis/molgenis/files/915781/test_5988.xlsx)

### DIFF
--- a/molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/AlgorithmServiceImpl.java
+++ b/molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/AlgorithmServiceImpl.java
@@ -111,13 +111,13 @@ public class AlgorithmServiceImpl implements AlgorithmService
 				Object result = jsMagmaScriptEvaluator.eval(algorithm, entity);
 				derivedValue = convert(result, targetAttribute);
 			}
-			catch (NullPointerException e)
-			{
-				return algorithmResult.errorMessage(
-						"Applying an algorithm on a null source value caused an exception. Is the target attribute required?");
-			}
 			catch (RuntimeException e)
 			{
+				if (e.getMessage() == null)
+				{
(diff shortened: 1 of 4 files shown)


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/AlgorithmServiceImpl.java
```diff
@@ -101,23 +101,27 @@ public void autoGenerateAlgorithm(EntityType sourceEntityType, EntityType target
...
-		return Iterables.transform(sourceEntities, entity ->
+		return stream(sourceEntities.spliterator(), false).map(entity ->
 		{
 			AlgorithmEvaluation algorithmResult = new AlgorithmEvaluation(entity);
-
 			Object derivedValue;
+
 			try
 			{
 				Object result = jsMagmaScriptEvaluator.eval(algorithm, entity);
```
ConnorStroomberg: could you not check for null / empty instead of catching the exception ? If we expect null it does not seem like a exceptional state.

ConnorStroomberg:
look good, will merge when tests have run succesfully

PR #6022, opened 2017-04-11
Fix #5864 If an attribute has not been automatically suggested, there are no attributes to select from when manually editing the AttributeMapping

### LINKED ISSUE #5864: If an attribute has not been automatically suggested, there are no attributes to select from when manually editing the AttributeMapping
### How to Reproduce
- Import [mappingservice-test.xlsx](https://github.com/molgenis/molgenis/files/807228/mappingservice-test.xlsx)
- Create a mapping project with HOP
- Add Prevent as a source
- ID is not automatically mapped
- Go into manual mapping mode

### Expected behavior
There are attributes for me to select from

### Observed behavior
No attributes are shown, and searching for id gives no results.
I have to use a bug to see all attributes to be able to select an attribute for mapping
Search needs to be super specific
Searching for id will NOT find id

Mark-de-Haan:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


tommydeboer:
As discussed with @Mark-de-Haan, the underlying problem is issue #6027. This PR is a quick fix to make it possible to see and select the attributes again.

PR #6026, opened 2017-04-12
Fix #5865 When uploading a table with a type decimal column but not all values are decimals, the importer succeeds, but the table cannot be shown

### LINKED ISSUE #5865: When uploading a table with a type decimal column but not all values are decimals, the importer succeeds, but the table cannot be shown
### How to Reproduce
Upload:
[test.xlsx](https://github.com/molgenis/molgenis/files/809962/test.xlsx)
Go to test

### Expected behavior
Importer fails telling me my data does not match my model

### Observed behavior
screen shot 2017-03-01 at 08 41 25

Which leads to:
screen shot 2017-03-01 at 08 41 17

tommydeboer:
Disallow ```NaN``` as decimal values in entities.

#### Checklist
- [x] Code reviewed
- [ ] Clean commits


tommydeboer:
As discussed with @Mark-de-Haan, the underlying problem is #6027. This PR is a quick fix to make it possible to see and select the attributes again.

PR #6029, opened 2017-04-12
M4765 As MOLGENIS user I want to have a persistent URL for each instance of an entity

Mark-de-Haan:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

## Test
- In the dataexplorer settings set standalone reports to true
- Go to Entity table and view the entityreport for any row
- Copy the URL
- Paste the URL in your browser and see default report
- In the Freemarker template, create a view-standalone-report-specific-sys_md_EntityType.ftl with some nice HTML
- Go to Entity table and view the entityreport for any row
- Copy the URL
- Paste the URL in your browser and see nice HTML


molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
```diff
@@ -415,11 +408,41 @@ public String viewEntityDetails(@RequestParam(value = "entityTypeId") String ent
...
+	 * Builds a model containing one entity and returns standalone report ftl view
+	 *
+	 * @param entityTypeId
+	 * @param entityId
+	 * @param model
+	 * @return standalone report view
+	 * @throws Exception if an entity name or id is not found
+	 */
+	@RequestMapping(value = "/details/{entityTypeId}/{entityId}", method = GET)
+	public String viewEntityDetailsById(@PathVariable(value = "entityTypeId") String entityTypeId,
```
ConnorStroomberg: A 'Unit' Test for this method would be good

molgenis-dataexplorer/src/main/java/org/molgenis/dataexplorer/controller/DataExplorerController.java
```diff
@@ -415,11 +408,41 @@ public String viewEntityDetails(@RequestParam(value = "entityTypeId") String ent
...
+			@PathVariable(value = "entityId") String entityId, Model model) throws Exception
+	{
+		EntityType entityType = dataService.getEntityType(entityTypeId);
+		Object id = getTypedValue(entityId, entityType.getIdAttribute());
+
+		model.addAttribute("entity", dataService.getRepository(entityTypeId).findOneById(id));
```
ConnorStroomberg: what happens if the find returns zero/null ?

Mark-de-Haan: Does not happen because the ID originates from a on row click in the UI.
The only way this fails, if somehow magically that row is deleted in the same millisecond it is clicked

PR #6069, opened 2017-04-19
Fix #6052 Apps create/edit buttons malfunction if modal was closed by…

### LINKED ISSUE #6052: Apps create/edit buttons malfunction if modal was closed by clicking next to it
### How to Reproduce
Start creating an app but do not finish, instead close the dialog by clicking next to it.
Than click the create button again.

### Expected behavior
- I get an empty form

### Observed behavior
- the values I filled out the first time are still there

The same occurs when clicking the edit button of a different app, so:
- click create, call the app "pietje", click next to the modal
- click the edit button of a different app -> this form now shows the dialog with "name:pietje"

dennishendriksen:
… clicking next to it

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-apps/src/main/resources/js/apps.js
+++ b/molgenis-apps/src/main/resources/js/apps.js
@@ -10,7 +10,7 @@
                 onSubmitSuccess: function () {
                     window.location = molgenis.contextUrl;
                 }
-            }), $('#form-modal-container')[0]).setState({showModal: true});
+            }), $('<div>')[0]);
         });
 
         $('.edit-app-btn').on('click', function () {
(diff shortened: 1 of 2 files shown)


molgenis-apps/src/main/resources/js/apps.js
```diff
@@ -10,7 +10,7 @@
                 onSubmitSuccess: function () {
                     window.location = molgenis.contextUrl;
                 }
-            }), $('#form-modal-container')[0]).setState({showModal: true});
+            }), $('<div>')[0]);
```
ConnorStroomberg: this seems very brittle to me, could you explain why this fixes the issue ?

dennishendriksen: The React component is rendered in a detached DOM element which causes the component to be recreated on every button select. The same strategy is used in the user manager and scripts plugin.

PR #6119, opened 2017-04-25
M4915: Mapping Job reports progress

tommydeboer:


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/job/MappingJob.java
```diff
@@ -1,83 +1,64 @@
...
-	List<String> successfulAnnotators = Lists.newArrayList();
-	List<String> failedAnnotators = Lists.newArrayList();
-	Exception firstException = null;
+	private final MappingService mappingService;
+	private final MappingJobExecution mappingJobExecution;
+	private final DataService dataService;
 
-	public MappingJob(String username, Progress progress, Authentication userAuthentication,
-			TransactionTemplate transactionTemplate)
+	MappingJob(MappingJobExecution mappingJobExecution, Progress progress, Authentication userAuthentication,
```
fdlk: Probably this one is easier to unit test if you write out the separate arguments.

Issue #5983, opened 2017-03-28
Mapping to existing entity is broken when existing entity has more then 20.000 rows


marikaris:
### How to Reproduce
Upload the content of the mappingservicebug.zip in the projects/Leiden folder on drive. Dont forget to alter the id of the admin user. 
Go to mapping service. Map to leiden_RP
Same thing happens when first mapping to RP (which succeeds) and then mapping again to RP

### Expected behavior
Mapping succeeds, leiden_RP is updated with age categories

### Observed behavior
On 3.1 (SNAPSHOT, master)
Failed to create integrated entity.

; nested exception is org.springframework.dao.DataAccessResourceFailureException: PreparedStatementCallback; SQL [UPDATE "RP#268f9e95" SET "BBMRI_id" = ?, "rp2_id" = ?, "rp3_id" = ?, "rp4_id" = ?, "biobank" = ?, "wgs" = ?, "genotypes" = ?, "methylome" = ?, "transcriptome" = ?, "metabolome" = ?, "age" = ?, "sex" = ?, "smoking" = ?, "wbcc" = ?, "ageGroup" = ? WHERE "BBMRI_id"= ?]; ERROR: out of shared memory Hint: You might need to increase max_locks_per_transaction.; nested exception is org.postgresql.util.PSQLException: ERROR: out of shared memory Hint: You might need to increase max_locks_per_transaction.

On 3.0, 500 error with in log: 
`Caused by: org.postgresql.util.PSQLException: ERROR: out of shared memory
  Hint: You might need to increase max_locks_per_transaction.
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2458) ~[postgresql-9.4.1211.jar:9.4.1211]
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2158) ~[postgresql-9.4.1211.jar:9.4.1211]
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:463) ~[postgresql-9.4.1211.jar:9.4.1211]
	at org.postgresql.jdbc.PgStatement.executeBatch(PgStatement.java:794) ~[postgresql-9.4.1211.jar:9.4.1211]
	at org.postgresql.jdbc.PgPreparedStatement.executeBatch(PgPreparedStatement.java:1662) ~[postgresql-9.4.1211.jar:9.4.1211]
[...]

PR #6123, opened 2017-04-26
M4915: As a user I want to be able to programmatically trigger the mapping service from a script 

fdlk:
See [M4915](http://wiki.gcc.rug.nl/ticket/4915)

# Open functional questions

When you click the green button of a mapping project, you get the following dialog:
screen shot 2017-04-26 at 13 53 39
- [x] How should the dialog look? And how should it distinguish between both cases? See also [M4865](http://wiki.gcc.rug.nl/ticket/4865) 

# Bugs encountered while building this
#6124 Permissions lost when doing something transactional in the afterCommit handler of a Job **(fixed)**
#5983 Mapping to existing entity is broken when existing entity has more then 20.000 rows **(fix in progress)**

# TODO
- [x] raise polling frequency of mapping job
- [x] improve logging (nr. of rows mapped for every source)
- [x] add issue: mapping to same entity with many rows - job doesn't finish
- [x] extend 'make integrated dataset' form
- [x] extend MappingJobExecution with name+label parameters
- [x] add id validation in web service
- [x] let MappingServiceImpl log to Progress.status instead of LOG.info
- [x] log total amount of rows mapped
- [x] update batches *2 for self-referencing entities
- [x] test with self-referencing entities
- [x] user entityMapping.getLabel() instead of entityMapping.getSourceEntityType.getLabel() 

#### Checklist
- [x] Code reviewed
- [ ] Clean commits


molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/job/MappingJobFactory.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.security.core.userdetails.UserDetailsService;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.support.TransactionTemplate;
+
+import static java.util.Objects.requireNonNull;
+import static org.molgenis.data.mapper.job.MappingJobExecutionMetadata.MAPPING_JOB_EXECUTION;
+
+/**
+ * Creates AnnotationJob based on its {@link AnnotationJobExecution}. Is a bean so that it can use {@link Autowired}
```
tommydeboer: Copied documentation

mswertz:
I would expect the mapping service endpoint user can only define the identifier of the entity to map to. Either an existing entity or a new entity (I would use seperate parameters for this). A new entity will be in the default folder for the user (e.g. the root). If user want to add more metadata (like label, package) they can use the existing metadata services to update. Do not duplicate this functionality

For the UI I would expect indeed two form inputs. Either a select box to choose an existing entity or a new input to create a new entity using an identifier.

For the UI we could consider to provide the user immediate options to edit the metadata of this entity (i.e. label and package) but ideally this would happen after the data has been mapped.

fdlk:
@mswertz Thanks for the prompt response!
Question: Why two parameters then?
We now have one single parameter `targetEntityTypeId` which is in itself good enough to distinguish between both scenarios. If an entityType exists for that ID, map to that existing entity, otherwise create a new one in the default package with label equal to the ID and map to that new entity.

tommydeboer:
As discussed with @mswertz and @fdlk:
- The ```MappingService``` can be called with two optional fields: ```package``` and ```label```
- These fields are used to set the package and label of a new entity
- These fields are ignored when mapping to an existing entity
- The fields are added to the ```MappingJobExecution```
- The web API is also extended with these optional parameters
- The Mapping Service's  "create integrated dataset"-dialog is changed so that:
  - You can choose to add a new entity type (and pick the package and label)
  - You can choose to select an existing entity type (using the entity drop down menu)
  - Source attribute can still be added

fdlk:
I'm building the dialog but am a little confused by the "Add source attribute" checkbox. Does it even make sense to set this to true when mapping to an existing entity?

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
```diff
@@ -289,59 +326,76 @@ private void compareTargetMetaDatas(EntityType targetRepositoryMetaData, EntityT
...
+			}, MAPPING_BATCH_SIZE);
 		}
+	}
 
+	private void mapAndUpsertEntities(EntityMapping sourceMapping, Repository<Entity> targetRepo,
+			EntityType targetMetaData, List<Entity> entities)
+	{
+		entities.forEach(sourceEntity ->
+		{
+			// FIXME adding/updating row-by-row is a performance bottleneck, this code could do streaming upsert
```
fdlk: We can simply split them into two batches here and batch insert/update those, no?

tommydeboer:
Ready for the final review!

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/service/impl/MappingServiceImpl.java
tommydeboer: This will be fixed in a future PR.

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
```diff
@@ -499,32 +525,109 @@ public String getSuggestedAlgorithm(@RequestBody GenerateAlgorithmRequest genera
...
 
 	/**
+	 * Checks if an entityTypeID already exists.
+	 * Used to validate ID field in the "New dataset" tab when creating a new integrated dataset.
+	 */
+	@RequestMapping("/isNewEntity")
+	@ResponseBody
+	public boolean isNewEntity(@RequestParam String targetEntityTypeId)
+	{
+		return dataService.getEntityType(targetEntityTypeId) == null;
```
tommydeboer: there's a ```dataService.hasRepository()``` for that!

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
```diff
@@ -499,32 +525,109 @@ public String getSuggestedAlgorithm(@RequestBody GenerateAlgorithmRequest genera
...
-			return "redirect:" + menuReaderService.getMenu().findMenuItemPath(DataExplorerController.ID) + "?entity="
-					+ name;
+			validateEntityName(targetEntityTypeId);
+			if (mappingService.getMappingProject(mappingProjectId) == null)
+			{
+				throw new MolgenisDataException("No mapping project found with ID " + mappingProjectId);
+			}
+			if (packageId != null)
+			{
+				if (dataService.getMeta().getPackage(packageId) == null)
```
fdlk: Should add check that it's not a system package.

molgenis-jobs/src/main/java/org/molgenis/data/jobs/ProgressImpl.java
```diff
@@ -83,7 +92,7 @@ public void failed(Exception ex)
 		JOB_EXECUTION_LOG.error("Failed. " + ex.getMessage(), ex);
 		jobExecution.setEndDate(Instant.now());
 		jobExecution.setStatus(FAILED);
-		jobExecution.setProgressMessage(ex.getMessage());
+		jobExecution.setProgressMessage(abbreviate(ex.getMessage(), 255));
```
fdlk: No longer needed, fixed in JobExecution class

molgenis-jobs/src/main/java/org/molgenis/data/jobs/ProgressImpl.java
```diff
@@ -47,6 +48,7 @@ public void start()
 		JobExecutionContext.set(jobExecution);
 		JOB_EXECUTION_LOG.info("Execution started.");
 		jobExecution.setStartDate(Instant.now());
+		jobExecution.setProgressInt(0);
```
fdlk: Not entirely sure if this won't break the indeterminate progress bars for jobs that don't set progress info.

molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
fdlk: That one returns false if the entity type exists but is abstract.

PR #6127, opened 2017-04-26
Fix #6124 Permissions lost when doing something transactional in the …

### LINKED ISSUE #6124: Permissions lost when doing something transactional in the afterCommit handler of a Job
### How to Reproduce
This only happens in a weird constellation of circumstances. Unfortunately it happens when we run the mapping service as a job.
Check out the code for #6123, create a mapping project, with target TypeTestRef, sources TypeTestRef and ScriptType. Fill in the missing values and start a mapping job mapping the data to a non-existing entity.

### Expected behavior
It works.

### Observed behavior
The mapping job fails, stating that you have no read permissions on EntityType.

```15:33:15.065 - Execution started.
15:33:15.840 - Mapping source TypeTestRef
15:33:15.884 - Mapping source Script type
15:33:16.082 - Failed. No [READ] permission on entity type [Entity type] with id [sys_md_EntityType]
```

fdlk:
…afterCommit handler of a Job

The idea is to turn the order around, do the permission switch outside the transaction instead of the other way around.

Before:
Before

After:
After

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-jobs/src/main/java/org/molgenis/data/jobs/Job.java
+++ b/molgenis-jobs/src/main/java/org/molgenis/data/jobs/Job.java
@@ -6,91 +6,44 @@ import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContext;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.transaction.TransactionException;
-import org.springframework.transaction.support.TransactionTemplate;
+import org.springframework.transaction.support.TransactionOperations;
 
 import java.util.Objects;
 import java.util.concurrent.Callable;
 
 /**
- * Superclass for molgenis jobs that keeps track of their progress.
+ * Superclass for molgenis jobs.
  */
 public abstract class Job<Result> implements Callable<Result>
 {
 	private static final Logger LOG = LoggerFactory.getLogger(Job.class);
 	private final Progress progress;
-	private TransactionTemplate transactionTemplate;
+	private TransactionOperations transactionOperations;
 	private Authentication authentication;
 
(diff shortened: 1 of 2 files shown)


molgenis-jobs/src/main/java/org/molgenis/data/jobs/Job.java
```diff
@@ -31,18 +31,22 @@ public Job(Progress progress, TransactionTemplate transactionTemplate, Authentic
 	@Override
 	public Result call()
 	{
+		if (transactionTemplate != null)
+		{
+			return transactionTemplate.execute((status) -> doCallInTransaction());
+		}
+		else
+		{
+			return doCallInTransaction();
```
tommydeboer: Are we actually doing the call in a transaction here? I'm a bit confused

fdlk: The Jobs run running on a separate executor thread.
If we do nothing, it will run without permissions and not in a transaction.
It is no bean so annotating the call() method with @Transactional won't do anything.
So we need to manually set the permissions and manually call the TransactionTemplate.

This PR turns the order inside out.
Call the transaction template within the user switch.
Used to be switch the user within the transaction.
The bug was that the afterCommit then runs after the permissions get restored.

fdlk: Or do you mean the case where no transaction template is specified? There we don't run in a transaction.

dennishendriksen:
@fdlk How does this relate to #6127? Is #6427 the proper solution and can we undo the previous solution?

PR #6147, opened 2017-05-09
M4504 Schedule scripts, run scripts as a job

fdlk:
Building on the feature/schedule branch, add scheduling of scripts.
Leaves `/scripts/**/run` in place, you can use it if you want to wait for the result to become available.
Adds `/scripts/**/start`, you can use it to watch the job as soon as the script has started. 

#### Checklist
- [x] Functionality works & meets specifications
- [ ] Code reviewed
- [ ] Code unit/integration/system tested
- [ ] User documentation updated
- [ ] (If you have changed REST API interface) view-swagger.ftl updated
- [ ] Test plan template updated
- [ ] Clean commits

PR #6149, opened 2017-05-10
M4504 Job scheduling framework

tommydeboer:
### TODO
- [x] File Ingest Job
- [x] Mapping Job (Tommy) #6154 
- [x] Script Job (Fleur) #6147 
- [x] Bucket Job (Bart) #6157
- [x] Fix the frontend (Fleur + Tommy) #6163
- [x] Validate job parameters (Tommy) #6156 
- [x] Set username in ScheduledJob decorator (Tommy) #6156 
- [x] Possibility to run job async via JobExecutor #6155 
- [x] Separate ScheduledJobType from the Factory so that non-schedulable jobs can also use the new JobFactory and Job interface and JobExecutor (Tommy) #6160 
- [x] (Nice to have) Figure out how the JobExecutor can find the JobFactory<SpecificJobExecution> from a List<JobFactory<? extends JobExecution>> given an instance of SpecificJobExecution. (Tommy, Fleur) #6160 #6161 
- [x] Update Job documentation (Fleur)
- [x] JobExecutionTemplate transactionality (Fleur) #6159 
- [x] Code cleanup (Tommy) #6162
- [x] Update test plan (Tommy)

**Review:**
- [x] Add *JobExecutions to ```sys_job``` package (Tommy) #6166 
- [x] Put json validation stuff in ```molgenis-data-validation``` module (Tommy, Fleur) #6165 
- [x] JSON schema as map in configs (Tommy) #6165 
- [x] Remove scripts 'any' field from Swagger docs (Tommy) #6168 
 
#### Frontend
- [x] how should it look? (**answer: update file ingest plugin**)

#### API
Do we want to move to a single api for starting any type of scheduled job in 
`/jobs/{scheduledJobType}/run` instead?
So have a POST to `/jobs/mappingService/run` with body`{ "targetEntityId": "foo", "mappingProjectId": "bar" }` start a mapping job
And have a POST to `/jobs/script/run` with `{"name":"add", "parameters": "{ a:1, b:2 }" }` start a job that runs the add script.

#### Script file results
A script can have an output file as result. This output file is stored in the filestore when the script is executed. The ScriptController currently waits for Script execution to finish, writes the output file to the response, and deletes it.
When the script runs asynchronously, the HTTPResponse is already gone when the job finishes. We need to set the result file URL in the job's resultURL.
We already have a way to serve files from filestore, you create a FileMeta for it. Shall we switch to using that? As an extra benefit, a script running on a remote server can then upload the result to FileMeta.
Is it okay if the result files remain on the server or should we add a "Delete once served" kind of feature to the files controller?

### Other issues
- [ ] The decorator that registers and unregisters jobs with quartz is not transactional. If a stream insert or update fails, it won't roll back the changes it made to the administration. What to do?
- [x] You need to fill in a user name when scheduling a job (which can be someone else's username). What to do about that? **Resolved by auto-filling the username**

### Future work
- Return job location url in api
- JSON validation of the schema
- JSON datatype
- Dynamic forms based on JSON parameters
- Make deleting quartz jobs transactional
- Move  old jobs to new job framework (+ remove old-job-new-job glue)
- Nest script parameters JSON in job parameters (instead of escaping it in a string)
- Generate JSON schema based on attributes of JobExecution
- Dynamically list swagger endpoints with correct parameters for each script

#### Checklist
- [x] Code reviewed


fdlk:
Test this please

tommydeboer:
@mswertz Scheduled jobs will be executed as the user that added the job. In absence of the ownership role, the username is saved with the scheduled job. This will mean that 'ownership' will change when another user updates that job. Is that ok?

fdlk:
@mswertz Can you also take a look at the other functional questions listed in the description above?

fdlk:
test this please

tommydeboer:
When I run a job in the plugin I get an error:

PR #6161, opened 2017-05-12
M4504 Make JobFactory abstract so we can reliably determine which factory m…

fdlk:
…atches which JobExecution.

#### Checklist
- [x] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-jobs/src/main/java/org/molgenis/data/jobs/JobFactory.java
+++ b/molgenis-jobs/src/main/java/org/molgenis/data/jobs/JobFactory.java
@@ -5,8 +5,7 @@ import org.molgenis.data.jobs.model.JobExecution;
 /**
  * @param <T> Type of the JobExecutions that this factory understands.
  */
-@FunctionalInterface
-public interface JobFactory<T extends JobExecution>
+public abstract class JobFactory<T extends JobExecution>
 {
 	/**
 	 * Creates a Job instance.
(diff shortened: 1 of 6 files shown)


molgenis-jobs/src/main/java/org/molgenis/data/jobs/JobFactoryRegistry.java
```diff
@@ -20,38 +18,23 @@
 public class JobFactoryRegistry
```
tommydeboer: Outdated docs!

molgenis-jobs/src/main/java/org/molgenis/data/jobs/JobFactory.java
```diff
@@ -5,14 +5,13 @@
...
+public abstract class JobFactory<T extends JobExecution>
 {
 	/**
 	 * Creates a Job instance.
 	 *
 	 * @param jobExecution {@link JobExecution} with job parameters
 	 * @return the job
 	 */
-	Job createJob(T jobExecution);
+	public abstract Job createJob(T jobExecution);
```
tommydeboer: Cool!

molgenis-jobs/src/main/java/org/molgenis/data/jobs/JobFactoryRegistry.java
fdlk: Fixed

PR #6163, opened 2017-05-16
M4504 Frontend for Scheduled Jobs

fdlk:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


tommydeboer:
When opening the plugin, the ```type``` attribute is expanded in the jobs table. When I collapse this attribute and then try to select another Scheduled Job, I get an error:

```Uncaught TypeError: Cannot read property 'id' of undefined```

Also: re-expanding the attribute gives a slightly different view than the original.

fdlk:
The <Table> is being fed with props that tell it what attributes to show. But it happily overrides those props in the state. Fixing this by hiding the offensive collapse button on the scheduled jobs plugin.

PR #6196, opened 2017-05-23
Fix #6184 Creating new entity with Mapping Service allows empty label

### LINKED ISSUE #6184: Creating new entity with Mapping Service allows empty label
### How to Reproduce
1. Add dummy mapping project
2. Create integrated data set > New entity
3. Only fill in identifier
4. Submit

### Expected behavior
Creating new entity is not allowed without label

### Observed behavior
New entity is created with empty string as label, making it invisible in the Data Explorer entity dropdown

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
+++ b/molgenis-data-mapper/src/main/java/org/molgenis/data/mapper/controller/MappingServiceController.java
@@ -549,6 +549,11 @@ public class MappingServiceController extends MolgenisPluginController
 			@RequestParam(required = false, name = "package") String packageId,
 			@RequestParam(required = false) Boolean addSourceAttribute)
 	{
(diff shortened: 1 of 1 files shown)


molgenis-data-mapper/src/main/resources/templates/view-single-mapping-project.ftl
```diff
@@ -246,7 +246,8 @@
                             <div>
                                 <label for="newTargetEntityLabel">Label</label>
                                 <input type="text" value="" name="label" id="newTargetEntityLabel"
-                                       class="form-control">
+                                       class="form-control"
+                                       required minlength="1"/>
```
fdlk: But the help block says "Optionally"! I think we should instead interpret "" as null.

Issue #6226, opened 2017-06-06
Can't delete packages


tommydeboer:
### How to Reproduce
1. Import _all_datatypes_
2. Go to Package entity in Data Explorer
3. Delete the _it_emx_ package

### Expected behavior
Package + sub packages + entities are removed

### Observed behavior
```Error! Value 'it_emx_datatypes' for attribute 'id' is referenced by entity 'sys_md_EntityType'.```

**(Note: this bug hides #6109)**

PR #6228, opened 2017-06-06
Fix #6109 & #6226 Delete package issues

### LINKED ISSUE #6109: Error deleting somewhat complicated package structure
### How to Reproduce
As admin
* Upload 
[very_advanced_data_example_v20170126.xlsx](https://github.com/molgenis/molgenis/files/951252/very_advanced_data_example_v20170126-2.xlsx) in a clean molgenis, use mode ADD and default options.
* Go to Data Explorer, select Package entity
* Delete root package

### Expected behavior
Everything is gone

Everything is still there.

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data-security/src/main/java/org/molgenis/data/security/meta/EntityTypeRepositorySecurityDecorator.java
+++ b/molgenis-data-security/src/main/java/org/molgenis/data/security/meta/EntityTypeRepositorySecurityDecorator.java
@@ -251,7 +251,7 @@ public class EntityTypeRepositorySecurityDecorator extends AbstractRepositoryDec
 	@Override
 	public void delete(Stream<EntityType> entities)
 	{
-		super.update(entities.filter(entityType ->
+		super.delete(entities.filter(entityType ->
 		{
 			validateDeleteAllowed(entityType);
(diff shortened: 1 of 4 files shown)


molgenis-data-security/src/main/java/org/molgenis/data/security/meta/EntityTypeRepositorySecurityDecorator.java
```diff
@@ -251,7 +251,7 @@ public void delete(EntityType entity)
 	@Override
 	public void delete(Stream<EntityType> entities)
 	{
-		super.update(entities.filter(entityType ->
+		super.delete(entities.filter(entityType ->
```
fdlk: whoops! :)

tommydeboer:
You're still getting the error because you're removing a package which has entities in it that are referenced from outside the package. Deleting the 'root' package should be possible now (where it wasn't before).

I agree that the error message isn't all too clear but it's coming straight from the ```PostgresqlExceptionTranslator```. Do you have any suggestions on how to create a better message?

PR #6235, opened 2017-06-12
M5085 Make integration test resources more configurable. Re-enable concurrent integration test runs. Build and test pull requests on travis-ci.org

fdlk:
The goal is to re-enable concurrent integration test runs for this repository.

# Running the integration tests in maven
To run integration tests in maven, call `mvn clean verify`. By default:
* A new postgres database will be created with timestamped name `molgenis_test_yyMMddHHmmssSSS`
* A new empty elasticsearch cluster will be launched on random ports with timestamped clustername `molgenisyyMMddHHmmssSSS`.
* The elasticsearch instance is configured from config files specified in `src/test/resources/conf/elasticsearch`. Since the tests use relatively small amounts of data, the configuration uses only 500 MB memory.
* Both database and cluster will be removed by maven after the integration tests have finished

## To use a pre-existing database
To run the integration tests on a pre-existing database, disable the `create-it-db` maven profile: `mvn clean verify -P'!create-it-db'`.

By default the build will connect to database `molgenis_test` on a local postgres installation, the username and password will both be `molgenis`.
You can override these settings in the molgenis-platform-integration-tests pom.xml file or in command line parameters:
```
mvn clean verify -P'!create-it-db' -Dit_db_name=myname -Dit_db_user=postgres -Dit_db_password` 
```
(No value means a blank password.)

## To use pre-existing elasticsearch cluster
To run the integration tests on a pre-existing elasticsearch cluster, disable the `create-it-es` maven profile: `mvn clean verify -P'!create-it-es'`

By default the index name will be `molgenis`, and the transport address will be `localhost:9300`.
You can override these settings in the molgenis-platform-integration-tests pom.xml file or in  command line parameters:
```
mvn clean verify -P'!create-it-es' -Delasticsearch.transport.addresses=localhost:12345 -Delasticsearch.cluster.name=testcluster`
```

# Running one or more Integration Tests outside maven
When developing, you can run the integration tests as usual.
The test code expects a postgres database to be running.
You can configure the database credentials and jdbc url in `src/test/resources/conf/molgenis.properties`

The code will not make any attempts to delete the integration test database and index.
(But you can easily create and call a cleanup script yourself to use when needed.)

# Nightly builds / Release builds
The standard build works fine on jenkins and can handle concurrent builds. It'll launch a new elasticsearch instance for each run. However, I found that the additional elasticsearch memory requirements make concurrent builds quite slow so I'd suggest that for pull requests we turn off the jenkins pull request builder and switch to https://travis-ci.org/molgenis/molgenis who provide a clean vm for each concurrent build.

# Building pull requests
A travis config file is added that shows travis how to build molgenis and execute the integration tests.
The integration tests will run in the postgres 9.6 database provided by travis-ci.org.
user is travis, database name is travis, password is blank
The build script will use the default options to create a elasticsearch cluster.
Travis is stricter than Jenkins about the length of build logfiles and will cancel a build that outputs over 4MB of logging. I've added configuration options that tell the test runner to mute test output sent to the console and streamlined the test logging configuration a bit.

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


fdlk:
test this please

fdlk:
Ready for review and feedback!
Have put the documentation here in the PR for now, but I'll add it to developer documentation or create a separate readme in the platform integration test module.

molgenis-jobs/src/test/java/org/molgenis/data/jobs/ProgressImplTest.java
```diff
@@ -143,6 +147,18 @@ public void jobCanceledAndMailFails()
 	@Import(JobTestConfig.class)
 	public static class Config
 	{
+		public Config()
+		{
+			ch.qos.logback.classic.Logger jobExecutionLog = (ch.qos.logback.classic.Logger) LoggerFactory
```
dennishendriksen:
```
if(... instanceof ...) {
  throw new RuntimeException("Expected Logback as SLF4J implementation")
}
```

molgenis-platform-integration-tests/pom.xml
```diff
@@ -9,12 +9,162 @@
...
+                <elasticsearch.transport.addresses>localhost:${elasticsearch.tcp.port}
+                </elasticsearch.transport.addresses>
+            </properties>
+            <build>
+                <plugins>
+                    <!-- Reserve tcp and http ports for elasticsearch -->
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>build-helper-maven-plugin</artifactId>
+                        <version>3.0.0</version>
```
dennishendriksen: Master pom is using different build-helper-maven-plugin version. Can we use the same version?

pom.xml
```diff
@@ -189,18 +191,18 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-jar-plugin</artifactId>
                     <version>2.4</version>
-
                     <configuration>
-                        <archive>
+						<archive>
```
dennishendriksen: Indentation error?

molgenis-platform-integration-tests/src/test/resources/conf/molgenis.properties
```diff
@@ -0,0 +1,6 @@
+#Molgenis server properties used at dev time
+db_user=molgenis
+db_password=molgenis
+db_uri=jdbc\:postgresql\://localhost/molgenis_test?reWriteBatchedInserts=true&autosave=CONSERVATIVE
```
dennishendriksen: ```reWriteBatchedInserts=true&autosave=CONSERVATIVE``` repeated in multiple places. Append them in Java or can we figure out another smart way to do this?

molgenis-platform-integration-tests/pom.xml
```diff
@@ -9,12 +9,162 @@
...
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
     <build>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-failsafe-plugin</artifactId>
```
dennishendriksen: Maven build reports the following warning:
```
[INFO] --- maven-failsafe-plugin:2.8.1:verify (verify) @ molgenis-platform-integration-tests ---
[INFO] Failsafe report directory: C:\Users\Dennis\Dev\molgenis\molgenis-platform-integration-tests\target\failsafe-reports
[WARNING] File encoding has not been set, using platform encoding UTF-8, i.e. build is platform dependent!
```

dennishendriksen:
Todo: switch pull request builder to Travis and update docs above.

molgenis-platform-integration-tests/pom.xml
fdlk: Is existing warning, I think. I always get it when building molgenis.

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+    - node_modules
+    - $HOME/.m2
+services:
+  - postgresql
+addons:
+  postgresql: '9.6'
+# Set the max_map_count for elasticsearch, as instructed by maven Elasticsearch plugin docs
+sudo: true
+before_script:
+  - sudo sysctl -w vm.max_map_count=262144
```
fdlk: I don't really understand what the setting does tbh, only saw in the travis docs that you cannot expect elasticsearch to start without it.
I think that means that when you follow the install SOP on the golden VM, the value also gets set to a high enough value.
I can think of two possible reasons why:
* It's a different OS, so perhaps the setting is already set by default on CentOS 
* Perhaps it gets set in the yum package.

dennishendriksen: I think you are right about the yum package: https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html

molgenis-jobs/src/test/java/org/molgenis/data/jobs/ProgressImplTest.java
fdlk: Updated!

molgenis-platform-integration-tests/pom.xml
fdlk: Updated!

pom.xml
fdlk: Something with tabs and spaces I think..
Updated.

molgenis-platform-integration-tests/src/test/resources/conf/molgenis.properties
fdlk: Makes sense to set them programmatically in the PostgresRepository?
https://jdbc.postgresql.org/documentation/publicapi/org/postgresql/ds/common/BaseDataSource.html#setReWriteBatchedInserts-boolean-

PR #6260, opened 2017-06-21
M5085 Decide EntityTypes to index based on indexing depth per EntityType

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data/src/main/java/org/molgenis/data/meta/model/EntityTypeMetadata.java
```diff
@@ -38,6 +38,8 @@
 	public static final String BACKEND = "backend";
 	public static final String INDEXING_DEPTH = "indexingDepth";
 
+	private static final long MAX_INDEXING_DEPTH = 20L;
```
fdlk: Is this needed?

PR #6263, opened 2017-06-21
M5085 Trigger backend actions when updating EntityTypes

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data/src/main/java/org/molgenis/data/RepositoryCollection.java
```diff
@@ -75,6 +75,8 @@
 	 */
 	void deleteRepository(EntityType entityType);
 
+	void updateRepository(EntityType entityType, EntityType updatedEntityType);
```
fdlk: Perhaps make a bit clearer that this is only called for updates to EntityType's own properties, and not changes to one or more of its Attributes

molgenis-data/src/main/java/org/molgenis/data/support/AbstractRepositoryCollection.java
```diff
@@ -37,6 +37,12 @@ public void deleteRepository(EntityType entityType)
 	}
 
 	@Override
+	public void updateRepository(EntityType entityType, EntityType updatedEntityType)
+	{
+		throw new UnsupportedOperationException();
```
fdlk: Is a bit weird construction, I'd expect an abstract method or a do-nothing implementation. But I see you're simply doing what we're already doing :)

PR #6265, opened 2017-06-22
Feature/appstore index

marikaris:
Functionality works, do not merge before #5532 is fixed and merged, then test. Now changing "No, I don't want to use a freemarker template" to "Yes, I want to use a freemarker template" (because you have not index.html) is very very broken (infinite spinner, cryptic message). 

#### Checklist
- [ ] Functionality works & meets specifications
- [ ] Code reviewed
- [ ] Code unit/integration/system tested
- [ ] User documentation updated
- [ ] (If you have changed REST API interface) view-swagger.ftl updated
- [ ] Test plan template updated
- [ ] Clean commits

Issue #5532, opened 2016-11-15
Error editing entity with attribute of type FILE


dennishendriksen:
#### Reproduce
- Import [file_datatype-test.xlsx](https://github.com/molgenis/molgenis/files/591337/file_datatype-test.xlsx)
- Select entity 'File' in 'Data explorer' plugin
- Select '+' button and create a new row with Description 'test' and Attachment 'file_datatype-test.xlsx'
- Select 'Save changes'
- Select 'Edit' button
- Update Description to 'test2'

#### Expected
- Value is updated to 'test2'

#### Observed
- Value is updated to 'test2' and the following error occurs: ```Error! An error occurred. Please contact the administrator. Uncaught InvalidStateError: Failed to set the 'value' property on 'HTMLInputElement': This input element accepts a filename, which may only be programmatically set to the empty string.```

PR #6271, opened 2017-06-23
Feature/3983 As app developer I want to create an frontend locally and then upload from github or zip

marikaris:
New pull request, old branch was dirty. Comments from old pull request (#6265) fixed. 

Functionality works, do not merge before #5532 is fixed and merged, then test. Now changing "No, I don't want to use a freemarker template" to "Yes, I want to use a freemarker template" (because you have not index.html) is very very broken (infinite spinner, cryptic message).

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-apps/src/main/java/org/molgenis/apps/model/App.java
```diff
@@ -86,14 +86,27 @@ public void setIconHref(String iconHref)
...
+
 	public FreemarkerTemplate getHtmlTemplate()
 	{
 		return getEntity(LANDING_PAGE_HTML_TEMPLATE, FreemarkerTemplate.class);
 	}
 
 	public void setHtmlTemplate(FreemarkerTemplate htmlTemplate)
 	{
-		set(LANDING_PAGE_HTML_TEMPLATE, htmlTemplate);
+		if (this.getUseFreemarkerTemplate())
```
ConnorStroomberg: better to remove if if statement, just set the value, the user of the entity can than use the getUseFreemarkerTemplate function to choose to use the template of not.

PR #6272, opened 2017-06-23
Fix #6270: When determining which entities need to be reindexed, abstract entities are not taken into account.

### LINKED ISSUE #6270: When determining which entities need to be reindexed, abstract entities are not taken into account.
### How to Reproduce
Create an abstract entity with a reference attribute.
Extend the abstract entity with a concrete entity.
Update the value of a referenced entity.

### Expected behavior
concrete entity rows get reindexed

### Observed behavior
the entity rows are not reindexed

fdlk:
Depends on #6257
The idea is to delegate the computation of which entities reference an entity to a new class IndexDependencyModel.

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-index/src/main/java/org/molgenis/data/index/IndexActionRegisterServiceImpl.java
```diff
@@ -109,85 +110,127 @@ public synchronized void register(EntityType entityType, String entityId)
...
+
+		Fetch extendsFetch = new Fetch();
+		extendsFetch.field(ID);
+		fetch.field(EXTENDS, extendsFetch);
+
+		Fetch attributesFetch = new Fetch();
+		Fetch refEntityFetch = new Fetch();
+		refEntityFetch.field(ID);
+		attributesFetch.field(AttributeMetadata.REF_ENTITY_TYPE, refEntityFetch);
+		fetch.field(ATTRIBUTES, attributesFetch);
```
tommydeboer: We could put the creation of the Fetch in a static initializer block so we don't need to create it each time. This will also clean up the method a bit

molgenis-data-index/src/main/java/org/molgenis/data/index/IndexDependencyModel.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.stream.Stream;
+
+import static com.google.common.collect.Maps.uniqueIndex;
+import static java.util.Collections.emptySet;
+import static java.util.stream.StreamSupport.stream;
+
+/**
+ * Models the dependencies between Entities.
+ */
+public class IndexDependencyModel
```
tommydeboer: Very nice! Although it's not really index specific, is it? I have a feeling this might come in handy in other places as well.

tommydeboer: As future work, we could consider generifying.

.travis.yml
```diff
@@ -18,7 +18,7 @@ before_script:
 script:
   # Travis aborts a job if it doesn't write to console for more than 10 minutes, use travis_wait to
   # Travis provides us with a travis user and travis database
-  - travis_wait 30 mvn verify --batch-mode --quiet -Dmaven.test.redirectTestOutputToFile=true -Ddb_user=travis -Ddb_password -Dit_db_name=travis -P!create-it-db
+  - travis_wait 50 mvn verify --batch-mode --quiet -Dmaven.test.redirectTestOutputToFile=true -Ddb_user=travis -Ddb_password -Dit_db_name=travis -P!create-it-db
```
fdlk: This is awkward, why do we suddenly need significantly more time to run the integration tests?

molgenis-data-index/src/main/java/org/molgenis/data/index/IndexActionRegisterServiceImpl.java
fdlk: I'd like to create a constructor method or builder for Fetch, cause this is way too convoluted.

molgenis-data-index/src/main/java/org/molgenis/data/index/IndexDependencyModel.java
fdlk: Good point! Perhaps the name should be DependencyModel

PR #6275, opened 2017-06-26
Fix/explain

fdlk:
There was a typo in the refactoring of the SemanticSearchService where it calls the ExplainApi.
It no longer handed in the attribute ID, but the entire Attribute instead.

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


tommydeboer:
Not sure if it's related to this fix, but when I map TypeTest to TypeTest, the Mapping Service doesn't auto-map any of the attributes (while in the past it auto-mapped almost all attributes).

fdlk:
@tommydeboer Shall we compare to the behavior of the master?

PR #6276, opened 2017-06-26
Proof of concept: security model using Spring Security ACL

dennishendriksen:
#### Features
- Package-level, entity-level, column-level security using Spring Security ACL
- Hierarchical permissions + applied to entity types Package, EntityType and Attribute

##### Must
- UI build on top of API
- Require one or more attributes in queries with operator SEARCH
- Validate whether attributes used in queries can be read by the current user
- Aggregations / semantic search should only return rows that a user can read
- Delete ACL when deleting entity including children
- Delete all entity ACLs when deleting entity type
- Count should require count instead of read permissions
- Assign entity-level permissions to groups
- Replace Permission with class based on BasePermission with our desired permissions
- EntitySecurityRepositoryDecorator should work with complete data and metadata instead of the data and metadata that the user supplied so that caching etc. won't break
- Fix ImportWizardControllerTest, AttributeRepositorySecurityDecoratorTest, AttributeRepositorySecurityDecoratorTest, PermissionManagerServiceImplTest
- Resolve FIXME in SecurityDecoratorUtils.java
- Resolve MetaDataServiceImpl.java system entity type fixme: get rid of SystemEntityRegistry
##### Should
- Investigate bootstrap warning: ``` WARN  o.m.d.i.IndexActionRegisterServiceImpl - Transaction aaaacxhn36a74gg36jrg2zaaae has caused 1000 IndexActions to be created. Consider streaming your data manipulations.```
- Fixme EntityUtils.equals methods should check newly introduced attributes
- Decide whether we still need system entity types: Authority, UserAuthority, GroupAuthority
- Write unit and integration tests
- Introduce EntityType setReadOnly/isReadOnly and use to determine EntityTypeResponse.writable and EntityTypeResponseV2.writable
- Create MutableAclService/AclService facade that performs Entity --> ObjectEntity mapping and use everywhere
- Instead of querying ACEs with expanded permissions, create ACEs with expanded permissions and query for the desired permissions
- Remove AclTest, AclTestMetadata, AclTestFactory
- EntityType.entityLevelSecurityInheritance should be of type XREF instead of STRING
- ACL config should cache ACLs

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-core-ui/src/test/java/org/molgenis/data/importer/wizard/ImportWizardControllerTest.java
+++ b/molgenis-core-ui/src/test/java/org/molgenis/data/importer/wizard/ImportWizardControllerTest.java
@@ -418,7 +418,7 @@ public class ImportWizardControllerTest extends AbstractMolgenisSpringTest
 		@Bean
 		public PermissionManagerServiceImpl pluginPermissionManagerServiceImpl()
 		{
-			return new PermissionManagerServiceImpl(dataService, molgenisPluginRegistry(), grantedAuthoritiesMapper());
+			return new PermissionManagerServiceImpl(dataService, molgenisPluginRegistry(), grantedAuthoritiesMapper(), null); // FIXME replace null
 		}
 
 		@Bean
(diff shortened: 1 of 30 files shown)


LuukDijkhuis:
It might be informative to add a MoSCoW indicator to each ToDo, aimed at a first production-ready release. Or is it all "Must"?

dennishendriksen:
@LuukDijkhuis Good suggestion, I've split up the todo list.

Issue #3983, opened 2015-10-21
filter wizard does not show description


bartcharbon:
Test plan states for the filter wizard:
"description displayed"

This is not the case. (assuming the description of the entity is meant?)

Mark-de-Haan:
Test plan has been updated.
It now mentions that on hover of an attribute it should show the description
https://docs.google.com/spreadsheets/d/17KTnwRdwwtTUBuqzUSXUSPFH1IfQB8YovbFptGrRkBU/edit#gid=0

PR #6293, opened 2017-06-28
#3983 Appstore index fixes based on comments

marikaris:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-apps/src/main/java/org/molgenis/apps/model/AppMetaData.java
+++ b/molgenis-apps/src/main/java/org/molgenis/apps/model/AppMetaData.java
@@ -13,52 +13,54 @@ import static org.molgenis.data.meta.model.Package.PACKAGE_SEPARATOR;
 import static org.molgenis.data.system.model.RootSystemPackage.PACKAGE_SYSTEM;
 
 @Component
-public class AppMetaData extends SystemEntityType {
-    private static final String SIMPLE_NAME = "App";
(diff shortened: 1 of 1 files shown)


molgenis-apps/src/main/java/org/molgenis/apps/model/AppMetaData.java
```diff
@@ -53,11 +53,13 @@ protected void init()
 		addAttribute(IS_ACTIVE).setDataType(BOOL).setLabel("Active").setNillable(false).setVisible(false)
 				.setDefaultValue(Boolean.FALSE.toString());
 		addAttribute(USE_FREEMARKER_TEMPLATE).setDataType(BOOL).setLabel("Use freemarker template as index.html")
-				.setNillable(false).setDefaultValue(Boolean.TRUE.toString());
+				.setNillable(false).setDefaultValue(Boolean.FALSE.toString()).setDescription(
+				"When using a freemarker template, the index page of your app should be placed in "
```
ConnorStroomberg: Would be nice if we could only show the first part of this message in case on the template. And just show the part about the index.html in case of no template.

ConnorStroomberg:
code looks fine , i'll wait till we fix the CI build before merging

molgenis-apps/src/main/java/org/molgenis/apps/model/AppMetaData.java
marikaris: That would be nice yes, however I don't see any way to implement that given the current functionality of the app meta data. I cannot set a description conditionally.

PR #6313, opened 2017-06-30
Feature/link to manual

ConnorStroomberg:
Add a link to the docs, so the user can find the docs from the app

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
+++ b/molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
@@ -205,8 +205,11 @@
                     </#list>
                 </ul>
 
+
+
                 <#if authenticated?? && authenticated>
                     <form id="logout-form" class="navbar-form navbar-right" method="post" action="/logout">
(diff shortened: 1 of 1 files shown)


molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
```diff
@@ -207,6 +207,7 @@
 
                 <#if authenticated?? && authenticated>
                     <form id="logout-form" class="navbar-form navbar-right" method="post" action="/logout">
+                        <a id="manual" href="https://www.gitbook.com/book/molgenis/molgenis/details" target="_blank" class="btn btn-secondary">Manual</a>
```
Mark-de-Haan: - Link should be https://molgenis.gitbooks.io/molgenis/content/
- Maybe 'documentation' is nicer as a link text?

ConnorStroomberg: - i have updated the link
- personally i like 'Manual' as it is shorter, i suggest we keep it for now until we haven feedback from users.

PR #6405, opened 2017-07-18
Type enrichment for dates and text + code refactoring

Mark-de-Haan:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-one-click-importer/src/main/java/org/molgenis/oneclickimporter/service/EntityService.java
```diff
@@ -5,5 +5,11 @@
...
 public interface EntityService
 {
-	EntityType createEntity(DataCollection dataCollection);
+	/**
+	 * Create an {@link EntityType} from a {@link DataCollection}
+	 *
+	 * @param dataCollection
+	 * @return the ID of the newly created {@link EntityType}
+	 */
+	String createEntityType(DataCollection dataCollection);
```
ConnorStroomberg: i would expect to get the entity, instead of just the id, the entity is nice and typed, the string is not :(, this sort of thing leads to spaghetti in my experience

molgenis-one-click-importer/src/main/java/org/molgenis/oneclickimporter/service/Impl/AttributeTypeServiceImpl.java
```diff
@@ -0,0 +1,166 @@
...
+		int currentRowIndex = 0;
+
+		AttributeType guess = getBasicAttributeType(dataValues.get(0));
+		while (currentRowIndex < rowCount && !guessCompleted)
+		{
+			Object value = dataValues.get(currentRowIndex);
+			AttributeType basicType = getBasicAttributeType(value);
+
+			guess = getCommonType(guess, basicType);
+			guess = getEnrichedType(guess, value);
```
ConnorStroomberg: why not create a new var enrichedGuess reusing the var can lead to problems when moving the lines around

molgenis-one-click-importer/src/main/java/org/molgenis/oneclickimporter/service/Impl/AttributeTypeServiceImpl.java
```diff
@@ -0,0 +1,166 @@
...
+
+			if (guess.equals(STRING) || guess.equals(TEXT))
```
ConnorStroomberg: would be nice if the could be part for the enrichment step

molgenis-one-click-importer/src/main/java/org/molgenis/oneclickimporter/service/Impl/EntityServiceImpl.java
```diff
@@ -39,78 +39,78 @@
...
 		this.dataService = requireNonNull(dataService);
+		this.metaDataService = requireNonNull(metaDataService);
 		this.entityManager = requireNonNull(entityManager);
-		this.oneClickImporterService = requireNonNull(oneClickImporterService);
+		this.attributeTypeService = requireNonNull(attributeTypeService);
 	}
 
 	@Override
-	public EntityType createEntity(DataCollection dataCollection)
+	public String createEntityType(DataCollection dataCollection)
```
ConnorStroomberg: better to use Type than String , make the service less tightly coupled.

molgenis-one-click-importer/src/test/java/org.molgenis.oneclickimporter/controller/OneClickImporterControllerTest.java
```diff
@@ -106,17 +103,16 @@ public void testXLSXFileImport() throws Exception
...
 		when(excelService.buildExcelSheetFromFile(file)).thenReturn(sheet);
 		when(oneClickImporterService.buildDataCollection("simple-valid", sheet)).thenReturn(dataCollection);
-		when(entityService.createEntity(dataCollection)).thenReturn(table);
+		when(entityService.createEntityType(dataCollection)).thenReturn(tableId);
 		when(table.getId()).thenReturn(tableId);
 
 		mockMvc.perform(fileUpload(OneClickImporterController.URI + "/upload").file(multipartFile))
-				.andExpect(status().isCreated())
-				.andExpect(header().string("Location", endsWith(tableId)));
+			   .andExpect(status().isOk());
```
ConnorStroomberg: it is creating a entity and should really return created, and setting the Location is a nice way to make a usable web api

ConnorStroomberg:
i would like it if we could concentrate on finishing the user story instead of refactoring working code.

molgenis-one-click-importer/src/test/java/org.molgenis.oneclickimporter/controller/OneClickImporterControllerTest.java
Mark-de-Haan: Problem was that the response I was getting was: 'response', it wasn't something I could work with on the client side

PR #6520, opened 2017-08-16
Fix #6492 EMX importer fails on BOM preventing CSV's to be uploaded in some cases

### LINKED ISSUE #6492: EMX importer fails on BOM preventing CSV's to be uploaded in some cases
### How to Reproduce
upload the csv found in the attached zip file

### Expected behavior
this works, file is imported, an entity test with an attribute ID is created

### Observed behavior
error stating the "name" is missing.

"name" is the first column in this, the same behaviour is seen with other CSV files, in every case the first column is reported missing. Copy pasting the contents of the file in a simple text editor and trying again fixes the issue on my machine.
probable cause: we do not process the Byte Order Mark correctly. https://en.wikipedia.org/wiki/Byte_order_mark

[attributes.zip](https://github.com/molgenis/molgenis/files/1207464/attributes.zip)

sidohaakma:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


tommydeboer:
When importing the attached zip file in #6492 I get an (different) error:

```
Unsupported attribute metadata: attributes.e
```

tommydeboer:
Can you add a unit test for this edge case?

PR #6588, opened 2017-09-05
M5263 As developer I want a testplan and unit tests for the one-click-importer

Mark-de-Haan:
#### Checklist
- [ ] Code reviewed
- [ ] Test plan template updated -> [Import test plan, see bottom](https://docs.google.com/spreadsheets/d/17KTnwRdwwtTUBuqzUSXUSPFH1IfQB8YovbFptGrRkBU/edit#gid=1715344550)
- [ ] Clean commits


molgenis-one-click-importer/src/main/frontend/src/store/actions.js
```diff
@@ -1,42 +1,25 @@
...
-
-    fetch('/plugin/one-click-importer/upload', options).then(response => {
-      response.text().then(jobUrl => {
-        dispatch(POLL_JOB_URL, jobUrl)
-      })
+    api.postFile('/plugin/one-click-importer/upload', file).then(response => {
+      dispatch(POLL_JOB_URL, response)
     })
   },
   [POLL_JOB_URL] ({dispatch, commit}, jobUrl) {
```
ConnorStroomberg: The function poll_job_url is not called from outside the actions, and therefore should be private and then tested as part of import_file.

Mark-de-Haan: I tried moving it, but due to the polling mechanism inside the POLL_JOB_URL action, it won't work because the Promise never gets closed.

You could say that the second time the api.get(...) is called you want to return a job object which is Failed or Finished. But I haven't figured out how to to this with testdouble.

PR #6590, opened 2017-09-05
Fix #6589 Navigator async JS tests are very different from other async tests

### LINKED ISSUE #6589: Navigator async JS tests are very different from other async tests

Mark-de-Haan:
#### Checklist
Functionality works & meets specifications
- [x] Code reviewed
- Code unit/integration/system tested)
- User documentation updated) 
- If you have changed REST API interface) view-swagger.ftl updated)
- Test plan template updated
- [x] Clean commits


molgenis-navigator/src/main/frontend/src/store/actions.js
```diff
@@ -59,34 +59,21 @@ function toEntity (item: any) {
 }
 
 /**
- * Get all ( first 1000 ) packages from the server
- * @returns {Promise}, on success resolves to list of 'all' packages, on failure resolves to error message
+ * Get a MOLGENIS rest api encoded query for the Package EntityType
```
ConnorStroomberg: the fact that we 'only' get the first 1000 items seems important to me, i suggest to keep this information in the js docs

molgenis-navigator/src/main/frontend/src/store/actions.js
```diff
@@ -59,34 +59,21 @@ function toEntity (item: any) {
 }
 
 /**
- * Get all ( first 1000 ) packages from the server
```
ConnorStroomberg: the fact we only get the first 1000 items seems important to me, i suggest to keep this as part of the js doc, as this is not derivable from the function name or params

molgenis-navigator/src/main/frontend/src/store/actions.js
```diff
@@ -59,34 +59,21 @@ function toEntity (item: any) {
...
-    }).catch((error) => {
-      reject(error)
-    })
-  })
+function getPackageQuery (query: string) {
+  return '/api/v2/sys_md_Package?sort=label&num=1000&q=id=q="' + encodeURIComponent(query) + '",description=q="' + encodeURIComponent(query) + '",label=q="' + encodeURIComponent(query) + '"'
 }
 
 /**
- * Query's the server of ( first 1000 ) packages.
```
ConnorStroomberg: Same as above; the fact we only get the first 1000 items seems important to me, i suggest to keep this as part of the js doc, as this is not derivable from the function name or params

molgenis-navigator/src/main/frontend/test/unit/utils/action.utils.js
```diff
@@ -0,0 +1,48 @@
+import { expect } from 'chai'
```
ConnorStroomberg: The project contains multiple copies of the file, i suggest to put them in a single place

molgenis-navigator/src/main/frontend/test/unit/specs/store/actions.spec.js
```diff
@@ -0,0 +1,330 @@
...
+
+  describe('QUERY_PACKAGES', function () {
+    it('should fetch the packages and call the SET_PACKAGES mutation', done => {
+      const package1 = {id: 'pack1', label: 'packLabel1'}
+      const response = {
+        items: [package1]
+      }
+
+      const get = td.function('api.get')
+      td.when(get('/api/v2/sys_md_Package?sort=label&num=1000')).thenResolve(response)
```
ConnorStroomberg: I really like the fact that the exact query string is checked in this test solution.

molgenis-navigator/src/main/frontend/test/unit/specs/store/actions.spec.js
```diff
@@ -0,0 +1,330 @@
...
+      td.replace(api, 'get', get)
+
+      testAction(actions.__QUERY_PACKAGES__, null, {}, [{type: SET_PACKAGES, payload: response.items}], [], done)
```
ConnorStroomberg: Having a method call with 6 unnamed params seems like a bad idea to me, the fact that half of them are null or empty also raises some flags with me. I think it makes the test hard to read. calling the action in the same way as the component code does seem clearer to me. 

If you do want to test using the testAction setup is would suggest naming the params ( using a object or vars) and/or creating specific named functions that have sensible defaults set.

PR #6600, opened 2017-09-07
Fix #6583 Index queries fail when index doesn't exist (yet)

### LINKED ISSUE #6583: Index queries fail when index doesn't exist (yet)
### How to Reproduce
go to the dataexplorer
select index action
enter any "search" query in the "google search box" on the top left

### Expected behavior
search is preformed and matching rows, if any, are shown

### Observed behavior
Error! One or more indexes 'sysidxindexaction_43bbc99b' not found.

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-data-index/src/main/java/org/molgenis/data/index/IndexedRepositoryDecorator.java
+++ b/molgenis-data-index/src/main/java/org/molgenis/data/index/IndexedRepositoryDecorator.java
@@ -4,6 +4,8 @@ import org.molgenis.data.*;
 import org.molgenis.data.QueryRule.Operator;
 import org.molgenis.data.aggregation.AggregateQuery;
 import org.molgenis.data.aggregation.AggregateResult;
+import org.molgenis.data.index.exception.UnknownIndexException;
+import org.molgenis.data.index.job.IndexJobScheduler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


molgenis-data-index/src/main/java/org/molgenis/data/index/IndexedRepositoryDecorator.java
```diff
@@ -135,4 +158,89 @@ private boolean querySupported(Query<Entity> q)
...
+				return searchService.count(getEntityType(), q);
+			}
+			catch (UnknownIndexException e1)
+			{
+				throw translateUnknownIndexException();
+			}
+		}
+	}
+
+	private Object indexSearchOne(Query<Entity> q)
```
fdlk: The trying stuff twice can be put in single method (I've tried it out, will put up PR)

molgenis-data-index/src/main/java/org/molgenis/data/index/IndexedRepositoryDecorator.java
```diff
@@ -135,4 +158,89 @@ private boolean querySupported(Query<Entity> q)
...
+				throw translateUnknownIndexException();
+			}
+		}
+	}
+
+	private void waitForIndexToBeStable()
+	{
+		try
+		{
+			indexJobScheduler.waitForIndexToBeStableIncludingReferences(getEntityType());
```
fdlk: This one used to be used in tests only, if we're going to use in production I think it's best to add a timeout. What do you think?

dennishendriksen: Its up to `waitForIndexToBeStableIncludingReferences` to interrupt after a certain amount of time. I propose to keep it like this and evaluate at the end of this sprint if and which changes are desired.

PR #6617, opened 2017-09-11
Global exception handling

dennishendriksen:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


sidohaakma:
Maybe think about changing the order in which exceptions are handled. This way we can enforce that exceptions are handled by the global exception handling. Maybe we should determine for each scenario which exception-handling-order we should use.

molgenis-web/src/main/java/org/molgenis/web/exception/GlobalControllerExceptionHandler.java
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.web.servlet.NoHandlerFoundException;
+
+import javax.servlet.http.HttpServletRequest;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import static org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR;
+import static org.springframework.http.HttpStatus.NOT_FOUND;
+
+@ControllerAdvice
```
Mark-de-Haan: This is badass 👍

molgenis-web/src/main/java/org/molgenis/web/exception/GlobalControllerExceptionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+			return new ResponseEntity<>(errorMessageResponse, INTERNAL_SERVER_ERROR);
+		}
+	}
+
+	private boolean isHtmlRequest(HttpServletRequest httpServletRequest)
+	{
+		try
+		{
+			URL url = new URL(httpServletRequest.getRequestURL().toString());
+			return url.getPath().startsWith(PluginController.PLUGIN_URI_PREFIX);
```
fdlk: Surely there's a better way to determine this?

dennishendriksen: Any suggestions in case the URL isn't mapped to a method?

sidohaakma:
When we implement this global exception handling ALWAYS log the message in the MOLGENIS log. Sometimes this does not happen in some shady corners of MOLGENIS.

PR #6630, opened 2017-09-12
M5312 description per plugin

Mark-de-Haan:
How to test:
1. Go to the sys_Plugin table
2. Edit a row
3. Create a nice description, you can go crazy with the HTML and links
4. The plugin for which you added a description will now have the description just under the menu
5. Plugins without a description will show nothing

#### Checklist
- [ ] Code reviewed
- [ ] [Test plan template updated](https://docs.google.com/spreadsheets/d/17KTnwRdwwtTUBuqzUSXUSPFH1IfQB8YovbFptGrRkBU/edit#gid=507173900)
- [ ] Clean commits

### DIFF
--- a/molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
+++ b/molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
@@ -180,6 +180,10 @@
 <div class="row">
 <div class="col-md-12">
 <div id="plugin-container">
+    <#assign plugin_description_key = plugin_id + '_description_text'>
+    <#if i18n[plugin_description_key] != "#" + plugin_id + "_description_text#">
+        ${i18n[plugin_description_key]}
+    </#if>
 </#macro>
 
 
--- a/docs/user_documentation/guide-customize.md
+++ b/docs/user_documentation/guide-customize.md
@@ -211,3 +211,25 @@ from the dropdown.
 4. Provide a translation for each available language.
 
 Missing translations will show up in English in the user interface.
+
+# Adding plugin descriptions
+Sometimes you might want to add some text to explain an interface, or to add a link to some documentation.
+You can now go to the sys_L10nString table in the data explorer and add descriptions for every plugin in multiple languages.
+
+1. In the top menu, select the Data Explorer.
+2. In the pulldown, select the entity 'Localization'.
+3. Push the plus icon.
+4. Fill in the form according to which plugin you would like to add a description to.
+
+Example for the importwizard plugin
+
+<code>
+    msgid: importwizard_description_text
+    namespace: importwizard
+    description: the i18n entry for the importwizard description text
+    en: Import wizard. Use the EMX format to create excel files for import
+</code>
+
+After saving, the importwizard plugin will now have a description at the top when the english language is selected.
+
+This is an admin only feature and allows the admin to use HTML.
--- a/docs/user_documentation/guide-customize.md
+++ b/docs/user_documentation/guide-customize.md
@@ -213,9 +213,6 @@ from the dropdown.
 Missing translations will show up in English in the user interface.
 
 # Adding plugin descriptions
-Sometimes you might want to add some text to explain an interface, or to add a link to some documentation.
-You can now go to the sys_L10nString table in the data explorer and add descriptions for every plugin in multiple languages.
-
 1. In the top menu, select the Data Explorer.
 2. In the pulldown, select the entity 'Localization'.
 3. Push the plus icon.
@@ -231,5 +228,4 @@ Example for the importwizard plugin
 </code>
 
 After saving, the importwizard plugin will now have a description at the top when the english language is selected.
-
 This is an admin only feature and allows the admin to use HTML.
--- a/docs/user_documentation/guide-customize.md
+++ b/docs/user_documentation/guide-customize.md
@@ -221,12 +221,12 @@ Missing translations will show up in English in the user interface.
 Example for the importwizard plugin, the pluginId is 'importwizard'. 
 The plugin ID can be found in the URL, the sys_Plugin table, or in the menu manager.
 
-<code>
-    msgid: [pluginId]_description_text
-    namespace: [pluginId]
-    description: the i18n entry for the importwizard description text
-    en: Import wizard. Use the EMX format to create excel files for import
-</code>
+```
+	msgid: [pluginId]_description_text
+    	namespace: [pluginId]
+    	description: the i18n entry for the importwizard description text
+    	en: Import wizard. Use the EMX format to create excel files for import
+```
 
 After saving, the importwizard plugin will now have a description at the top when the english language is selected.
 This is an admin only feature and allows the admin to use HTML.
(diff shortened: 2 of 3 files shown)


molgenis-data-plugin/src/main/java/org/molgenis/data/plugin/model/PluginMetadata.java
```diff
@@ -29,6 +30,6 @@ public void init()
 
 		addAttribute(ID, ROLE_ID).setLabel("Identifier");
 		addAttribute(LABEL, ROLE_LABEL, ROLE_LOOKUP).setNillable(false).setUnique(true).setLabel("Label");
-		addAttribute(DESCRIPTION, ROLE_LOOKUP).setLabel("Description");
+		addAttribute(DESCRIPTION, ROLE_LOOKUP).setLabel("Description").setDataType(HTML);
```
dennishendriksen: TEXT datatype would be more appropriate.

molgenis-web/src/main/java/org/molgenis/web/PluginController.java
```diff
@@ -58,6 +60,16 @@ public Entity getPluginSettings()
 		return RunAsSystemAspect.runAsSystem(() -> getPluginSettings(entityTypeId));
 	}
 
+	/**
+	 * Returns the description of the plugin
+	 */
+	public String getDescription(String pluginId)
+	{
+		Plugin plugin = RunAsSystemAspect.runAsSystem(
```
dennishendriksen: If you are not allowed to read a Plugin entity, then you should not be allowed to read the description. Why runAsSystem?

molgenis-web/src/main/java/org/molgenis/web/PluginInterceptor.java
```diff
@@ -78,6 +78,7 @@ public void postHandle(HttpServletRequest request, HttpServletResponse response,
 			modelAndView.addObject(PluginAttributes.KEY_AUTHENTICATED, SecurityUtils.currentUserIsAuthenticated());
 			modelAndView.addObject(PluginAttributes.KEY_PLUGIN_ID_WITH_QUERY_STRING,
 					getPluginIdWithQueryString(request, pluginId));
+			modelAndView.addObject(PluginAttributes.KEY_PLUGIN_DESCRIPTION, molgenisPlugin.getDescription(pluginId));
```
dennishendriksen: Seeing plugin id and plugin description makes me wonder why there is no plugin label?

dennishendriksen: How about using one key - value that contains the plugin id and description (and possibly label, see previous remark) so we don't have to update the interceptor when we change Plugin attributes in the future.

molgenis-web/src/main/java/org/molgenis/web/bootstrap/PluginPopulator.java
```diff
@@ -31,28 +30,22 @@ public PluginPopulator(DataService dataService, PluginFactory pluginFactory)
 
 	public void populate(ApplicationContext ctx)
 	{
-		Map<String, Plugin> newOrChangedPluginMap = getPlugins(ctx);
-		List<Plugin> deletedPlugins = new ArrayList<>();
+		Map<String, Plugin> allPlugins = getPlugins(ctx); // Plugins from context
```
dennishendriksen: Are these changes related to the description update? If this a bug fix? In case of a bug fix create an issue and use a separate PR.

molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
```diff
@@ -180,6 +180,9 @@
 <div class="row">
 <div class="col-md-12">
 <div id="plugin-container">
+    <#if plugin_description??>
+        ${plugin_description}
```
dennishendriksen: Why display the plugin description above every plugin?
- It will push down relevant content even more.
- I probably know what this plugin does since I selected it
- When I open another application (e.g. Excel) it doesn't start with a line telling me what Excel is.
- Displaying 'This is a plugin that displays static content stored in the database' above the Home plugin is not something I would want to see as a user.

I would expect to see the Plugin description in the menu manager plugin or when I select it as a data set.

dennishendriksen: Inserting raw HTML is dangerous and could lead to vulnerabilities. I suggest using TEXT type and for Plugin.DESCRIPTION and render all descriptions as ${plugin_description?html} with a similar style regardless of content

molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
```diff
@@ -180,6 +180,9 @@
 <div class="row">
 <div class="col-md-12">
 <div id="plugin-container">
+    <#if plugin_description??>
```
dennishendriksen: How to deal with long descriptions? See for example the data explorer entity type description and how we deal with abbreviation and hovering.

dennishendriksen: How about internationalization?

molgenis-web/src/test/java/org/molgenis/web/PluginInterceptorTest.java
```diff
@@ -21,13 +23,21 @@
 	private Ui molgenisUi;
 	private PermissionService permissionService;
 	private Authentication authentication;
+	private DataService dataService;
 
 	@BeforeMethod
-	public void setUp()
+	public void beforeMethod()
```
dennishendriksen: In MOLGENIS this method is usually name setUpBeforeMethod, in suggest renaming it.

Mark-de-Haan: Actually:

setUpBeforeMethod: 88 occurrences in 88 files
beforeMethod: 100+ occurrences in 100+ files

But I agree we should be consistent in naming these @BeforeMethod and @BeforeClass annotated classes

molgenis-web/src/main/java/org/molgenis/web/bootstrap/PluginPopulator.java
Mark-de-Haan: Yeah this is a fix where the plugins from the context always overwrite the plugins in the database at startup.

In practice this means that the plugin description (which is filled in at runtime) is overwritten with NULL by the context when the server is rebooted.

Il make a separate PR

molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
Mark-de-Haan: > It will push down relevant content even more.


It will only push down content if there is a description, and if there is no description, nothing is pushed down.

> I probably know what this plugin does since I selected it

Practice says this is not true, users tend to discover what the interface has to offer.

> When I open another application (e.g. Excel) it doesn't start with a line telling me what Excel is.

But expecting a user to look at the import wizard and then know how to create an EMX file etc is a bit 2 much. 

> Displaying 'This is a plugin that displays static content stored in the database' above the Home plugin is not something I would want to see as a user.

Adding a description for the home plugin seems to be pointless, so no one will ever do it? And if they do, well its their instance of MOLGENIS.

Everything about this description is 100% configurable. If you do not want it, then you don't have it.

Mark-de-Haan:
@dennishendriksen completely changed the implementation.

You can now add a #[pluginId]_description_text msgid to the Localization table to enable a plugin description in multiple languages. 

100% configurable, and admin feature. So HTML (see morris) injection is not a concern

PR #6644, opened 2017-09-14
Feature/#5269 filter sytem entities in entity editor

ConnorStroomberg:
Hide system packages and entities for non super user.
+
Move duplicate controller code to abstract parent class

#### Checklist
- [x] Code reviewed
- [x] Clean commits


molgenis-metadata-manager/src/main/frontend/src/store/mutations.js
```diff
@@ -28,14 +30,44 @@ const swapArrayElements = (array: Array<EditorAttribute>, originalIndex: number,
   return array
 }
 
+/**
+ * Filter out all system entities unless user is superUser
+ * @param entities
+ * @returns {Array.<Object>}
+ */
+const filterNonVisibleEntities = (entities: Array<Object>) => {
+  if (INITIAL_STATE.isSuperUser) {
```
Mark-de-Haan: alternative lambda: 

```js
return INITIAL_STATE.isSuperUser ? entities : entities.filter(entity => !entity.id.startsWith(SYS_PACKAGE_ID + '_'))
```

PR #6701, opened 2017-09-21
Fix #6695 Search All breaks when there are Entity Types not in a package

### LINKED ISSUE #6695: Search All breaks when there are Entity Types not in a package
### How to Reproduce
1. Log in as super user
2. Create an Entity Type with the Metadata Manager without putting it in a package
3. Go to the Search All plugin and search for whatever you like 

### Expected behavior
Results.

### Observed behavior
No results.
There is an internal server error, which turns out to be an NPE @ line 62 of ```SearchAllService```.

Please discuss if it isn't a bug that it's possible to create a package-less Entity Type in the first place.

ConnorStroomberg:
Fix #6695 Search All breaks when there are Entity Types not in a package

#### Checklist
- [x] Code reviewed
- [x] User documentation updated (N.A.)
- [x] (If you have changed REST API interface) view-swagger.ftl updated (N.A.)
- [x] Test plan template updated (N.A. since it is unit tested)
- [x] Clean commits

### DIFF
--- a/molgenis-navigator/src/main/frontend/src/store/actions.js
+++ b/molgenis-navigator/src/main/frontend/src/store/actions.js
@@ -3,7 +3,7 @@ import type { Package, Entity } from '../flow.types'
 import { INITIAL_STATE } from './state'
 // $FlowFixMe
 import api from '@molgenis/molgenis-api-client'
-import { RESET_PATH, SET_ENTITIES, SET_ERROR, SET_PACKAGES, SET_PATH } from './mutations'
+import { RESET_PATH, SET_ENTITIES, SET_ERROR, SET_PACKAGES, SET_PATH, SET_QUERY } from './mutations'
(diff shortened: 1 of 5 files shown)


marikaris:
I found this bug: https://github.com/molgenis/molgenis/issues/6705
But it was already in RC1, but I think it should be fixed since this will be used? I don't know if this happens for other packages than sys as well, but I guess it does and then it is certainly relevant.

Besides that. When I reproduce the issue in the navigator (not the search all), it is partly fixed, since my entity called "Tommy" is not showing in the home of the navigator but it is in the root because it's not in a package

marikaris:
Created new issue: https://github.com/molgenis/molgenis/issues/6706
Code is better now than it was before the PR so, let's merge and file the issue to fix later (it's not an easy one)

PR #6707, opened 2017-09-22
Fix #6673 Altering ref entity results in broken table

### LINKED ISSUE #6673: Altering ref entity results in broken table
### How to Reproduce
Upload [vkgl-model.xlsx](https://github.com/molgenis/molgenis/files/1317795/vkgl-model.xlsx)
Open the 'Consensus' Entity in the metadata manager.
Change ref entity in 'OMIM disease' attribute to another entity

### Expected behavior
Another entity is referenced and when the column is empty or the references are correct this will go fine

### Observed behavior
Unknown error!
Spinner
Table not usable (is fixed by changing the refentity back)

Mark-de-Haan:
#### Checklist
- [x] Code reviewed
- [x] Test plan template updated; No but i did it for you, otherwise the test plan would fail now
- [x] Clean commits

### DIFF
--- a/molgenis-metadata-manager/src/main/frontend/src/components/MetadataManagerEntityEditForm.vue
+++ b/molgenis-metadata-manager/src/main/frontend/src/components/MetadataManagerEntityEditForm.vue
@@ -20,9 +20,9 @@
         </div>
 
         <div class="form-group-row float-right">
-            <save-button :onClick="saveEntityType" :disabled="!isEntityTypeEdited">
-              {{ 'save-changes-button' | i18n }}
-            </save-button>
+          <save-button :onClick="saveEntityType" :disabled="!isEntityTypeEdited">
+            {{ 'save-changes-button' | i18n }}
+          </save-button>
           <button @click="deleteEntityType(editorEntityType.id)" class="btn btn-danger"
                   :disabled="editorEntityType.isNew">
             {{ 'delete-entity-button' | i18n }}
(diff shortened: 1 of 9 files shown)


ConnorStroomberg:
when running `yarn test` i get the following output:

```
/karma-phantomjs-launcher" plugin:
  Path must be a string. Received null
```

Mark-de-Haan:
Should be a local problem, because yarn test works on my machine.
I use yarn v1.0.2

molgenis-metadata-manager/src/main/frontend/src/store/actions.js
```diff
@@ -170,6 +170,10 @@ export default {
    * Persist metadata changes to the database
    */
   [SAVE_EDITOR_ENTITY_TYPE] ({commit, state}: { commit: Function, state: State }) {
+    if (!state.editorEntityType.labelAttribute) {
+      commit(UPDATE_EDITOR_ENTITY_TYPE, {key: 'labelAttribute', value: state.editorEntityType.idAttribute})
```
ConnorStroomberg: Changing the form input on save seems confusing to me, If a label is needed for entity to be valid, i suggest making it a required field ( both in back- and frontend ) and notifying the user about the need for a label attribute.

molgenis-metadata-manager/src/main/frontend/src/components/MetadataManagerEntityEditForm.vue
```diff
@@ -123,7 +123,13 @@
     name: 'metadata-manager-entity-edit-form',
     methods: {
       saveEntityType () {
-        this.$store.dispatch(SAVE_EDITOR_ENTITY_TYPE)
+        if (this.editorEntityType.idAttribute === null || this.editorEntityType.idAttribute === undefined) {
+          this.$store.commit(CREATE_ALERT, {type: 'warning', message: 'ID attribute can not be empty'})
+        } else if (this.editorEntityType.labelAttribute === null || this.editorEntityType.labelAttribute === undefined) {
+          this.$store.commit(CREATE_ALERT, {type: 'warning', message: 'Label attribute can not be empty'})
```
ConnorStroomberg: This is good, but does not stop the entity being created on the server side. The code creating the entity should check the modal validity. The checking model validity on the client side could maybe be grouped into a dedicated named function.

PR #6710, opened 2017-09-22
Fix #6709 After using search in the navigator the 'home' button stops working

### LINKED ISSUE #6709: After using search in the navigator the 'home' button stops working
### How to Reproduce
Go to the 'Navigator' plugin ( if it is not in the menu, add it using the menu)
Use the search functionality to search for something ( b.v. 'Mail')
Click on the home icon, at the start of the breadcrumb path

### Expected behaviour
The 'home' view is dispalyed

### Observed behaviour
Nothing happens

ConnorStroomberg:
Reset state when clicking on the home breadcrumb, even when state is already set to home ( when showing search result)

#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-navigator/src/main/frontend/src/components/Navigator.vue
+++ b/molgenis-navigator/src/main/frontend/src/components/Navigator.vue
@@ -28,9 +28,9 @@
       <div class="col input-group">
         <ol class="breadcrumb">
           <li class="breadcrumb-item">
-            <router-link to="/">
+            <a :href="homeUrl" v-on:click="reset">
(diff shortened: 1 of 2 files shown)


Mark-de-Haan:
When clearing my search, sub packages under System are shown, instead of only the system package. Different issue I think, but in the same area

ConnorStroomberg:
@Mark-de-Haan i also spotted that and fixed it in a separate PR https://github.com/molgenis/molgenis/pull/6711

PR #6721, opened 2017-09-26
Fix #6618 '+' character is not encoded in the URL, causing the request to the negotiator to fail

### LINKED ISSUE #6618: Search queries with spaces do not get sent to negotiator
### How to Reproduce
- Go to the BBMRI-ERIC directory
- Search for OMEGA alpha
- Click on 'Go to sample/data negotiation'
- Click on 'Yes, send to negotiator'

### Expected behavior
I'm taken to the Negotiator

### Observed behavior
I receive an error: 
`Error! An error occurred. Please contact the administrator.`

Mark-de-Haan:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
+++ b/molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
@@ -256,7 +256,7 @@ $.when($,
                     } else {
                         entityCollectionRequest.q.push({
                             operator: 'SEARCH',
-                            value: searchQuery
+                            value: encodeURIComponent(searchQuery)
                         });
                     }
                 }
--- a/molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
+++ b/molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
@@ -256,7 +256,7 @@ $.when($,
                     } else {
                         entityCollectionRequest.q.push({
                             operator: 'SEARCH',
-                            value: encodeURIComponent(searchQuery)
+                            value: searchQuery
                         });
                     }
                 }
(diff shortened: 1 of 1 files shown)


molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
```diff
@@ -256,7 +256,7 @@ $.when($,
                     } else {
                         entityCollectionRequest.q.push({
                             operator: 'SEARCH',
-                            value: searchQuery
+                            value: encodeURIComponent(searchQuery)
```
ConnorStroomberg: Could the encoding be moved to a later point in de code to 'fix' 'all' encoding problems instead of this specific branch ?.

Mark-de-Haan: The reason why I placed it here, is because the **IF** and **ELSE IF** branches have to do with searching for genomic positions e.g. chromosome_number:position-position 12:19083232-987393292.

All other 'normal' queries are routed to this **ELSE** branch, so technically we do fix all encoding problems for the 'google search' field

marikaris:
When I search for: Bimetra Biobank, i don't retrieve any rows, as I did previously. I think the + gets encoded instead of the space and no results are found for Bimetra+Biobank

marikaris:
Something still goes wrong: I look for N/A and I assume to find a lot of NL tables as I do on molgenis01, but I get tables from Belgium for some reason while they did not even specify N/A in their data.

marikaris:
The search query that is send to the negotiator still contains the +

marikaris:
When I send to negotiator without filtering on something else than free text with space in free text, I receive error. When I have something else selected (for instance Country: belgium), my free text search is shown in free text search on negotiator but it is ignored in search results.

molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
```diff
@@ -425,7 +425,7 @@ $.when($,
             // update browser state
             if (filter) history.pushState(state, '', molgenis.getContextUrl() + '?' + $.param(cleanState)
```
fdlk: Create new parameter encoding method and use it instead of $.param

marikaris:
The + is ok, there is another problem

PR #6725, opened 2017-09-27
Fix #6724 Search all broken with abstract entities in database.

### LINKED ISSUE #6724: Search all broken with abstract entities in database
### How to Reproduce
Load typetest
Load VKGL model (model with abstract entities)
[vkgl-model.xlsx](https://github.com/molgenis/molgenis/files/1336326/vkgl-model.xlsx)

Go to Search all 
Search for test

### Expected behavior
I get typetest

### Observed behavior
Nothing and errors in network tab
screen shot 2017-09-27 at 10 36 05

ConnorStroomberg:
Getting the repository of a abstract entity results in a null pointer.
To fix this abstract entities are filtered from the search result.

It might be better if MetaDataService.getRepository(id) did not return null in case of a abstract repository.

#### Checklist
- [x] Code reviewed
- [x] Clean commits


marikaris:
The PR is ok, but we should not forget to close this one manually after merging with master because there is no # in the commit message

PR #6740, opened 2017-09-28
Fix #6713 One-click-importer permission issue

### LINKED ISSUE #6713: One-click-importer permission issue for non-superusers
### How to Reproduce
Create a user with:
- View permissions on one click importer plugin, OneClickImporterJobExecution
- View permissions on dataexplorer and dataexplorer-settings and genomebrowsersettings and attributes
- Write permissions on attribute, entityType, Package
- Login as user
- Import a simple csv or xlsx
- Click the link to the dataexplorer

### Expected behavior
I can see my data

### Observed behavior
Error
After looking at permissions on the EntityType, I can see the user has none

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-security/src/main/java/org/molgenis/security/permission/PermissionSystemServiceImpl.java
+++ b/molgenis-security/src/main/java/org/molgenis/security/permission/PermissionSystemServiceImpl.java
@@ -48,7 +48,7 @@ public class PermissionSystemServiceImpl implements PermissionSystemService
 		this.roleHierarchy = requireNonNull(roleHierarchy);
 		this.dataService = requireNonNull(dataService);
 		this.principalSecurityContextRegistry = requireNonNull(principalSecurityContextRegistry);
-		this.authenticationAuthoritiesUpdater = authenticationAuthoritiesUpdater;
+		this.authenticationAuthoritiesUpdater = requireNonNull(authenticationAuthoritiesUpdater);
 	}
 
 	@Override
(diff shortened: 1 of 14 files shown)


molgenis-security/src/main/java/org/molgenis/security/permission/PermissionSystemServiceImpl.java
```diff
@@ -61,15 +64,36 @@ public void giveUserWriteMetaPermissions(Collection<EntityType> entityTypes)
...
+
+	private void updateSecurityContextAuthorities(SecurityContext securityContext,
+			Collection<? extends GrantedAuthority> newAuthorities)
+	{
+		Collection<? extends GrantedAuthority> authorities = securityContext.getAuthentication().getAuthorities();
+		List<GrantedAuthority> updatedAuthorities = new ArrayList<>(authorities);
+		updatedAuthorities.addAll(newAuthorities);
+
+		Authentication authentication = securityContext.getAuthentication();
+		UsernamePasswordAuthenticationToken newToken = new UsernamePasswordAuthenticationToken(
```
fdlk: I think that means that they'll need to re-enter the 2FA code if 2FA is enabled.

fdlk:
Functional question: Do we want all of the user's sessions to be updated, or only the one that triggered the import job?

molgenis-security/src/main/java/org/molgenis/security/permission/PermissionSystemServiceImpl.java
dennishendriksen: Oops, I wasn't aware of the fact that tokens could be of a different type. This fix indeed doesn't work for 2FA.

sidohaakma:
In the PermissionSystemServiceImpl use requireNonNull on this line:
this.authenticationAuthoritiesUpdater = authenticationAuthoritiesUpdater;

PR #6749, opened 2017-09-29
Test frozen lock file fail, DO NOT MERGE

ConnorStroomberg:
Test frozen lock file fail, DO NOT MERGE


ConnorStroomberg:
it works, i see a the expected error in the CI output and the build fails

PR #6750, opened 2017-09-29
DO NOT MERGE test change yarn version 

ConnorStroomberg:


ConnorStroomberg:
other yarn version does not make build fail, need to run more tests to verify package deps are still the same ( i guess so)

PR #6785, opened 2017-10-05
Fix SONAR vulnerabilities

dennishendriksen:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-core-ui/src/main/java/org/molgenis/data/importer/wizard/OptionsWizardPage.java
```diff
@@ -92,7 +92,10 @@ public String handleRequest(HttpServletRequest request, BindingResult result, Wi
 					fileRepositoryCollectionFactory.createFileRepositoryCollection(tmpFile).getFileNameExtensions());
 
 			File file = new File(tmpFile.getParent(), userGivenName + "." + extension);
-			tmpFile.renameTo(file);
+			if (!tmpFile.renameTo(file))
+			{
+				LOG.error("Failed to rename '{}' to '{}'", tmpFile.getName(), file.getName());
```
fdlk: If you expect it can actually happen, add error to the result. Otherwise, don't just log but also throw exception.

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/core/entity/impl/GoNLAnnotator.java
```diff
@@ -110,7 +110,7 @@ public void init()
 				return createGoNlOutputAttributes();
 			}
 
-			public String postFixResource = "";
+			private String postFixResource = "";
```
fdlk: This should be a local variable, not a field

molgenis-core-ui/src/main/java/org/molgenis/data/importer/wizard/OptionsWizardPage.java
dennishendriksen: A failed rename attempt of a system temporary file doesn't have to impact the user import.

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/core/entity/impl/GoNLAnnotator.java
dennishendriksen: I've tried to keep the fixes as minimal as possible and suggest to implement your comment separately since this code is quite hairy.

molgenis-file-ingester/src/main/java/org/molgenis/file/ingest/meta/FileIngestJobExecutionMetaData.java
```diff
@@ -22,7 +22,7 @@
 
 	public static final String URL = "url";
 	public static final String LOADER = "loader";
-	public static final List<String> LOADERS = Collections.singletonList("CSV");
+	public static final List<String> LOADERS = ImmutableList.of("CSV");
```
fdlk: Doesn't do anything, singletonList is also immutable!

molgenis-data-annotators/src/main/java/org/molgenis/data/annotation/core/entity/impl/GoNLAnnotator.java
fdlk: @dennishendriksen is good. Bit worried that if we only fix the sonar warning and not the root cause, we'll not get the maximum benefit out of the warnings.

molgenis-file-ingester/src/main/java/org/molgenis/file/ingest/meta/FileIngestJobExecutionMetaData.java
dennishendriksen: Sonar now recognizes that the List is immutable. I prefer this solution above marking the Sonar warning as false positive, because it might prevent issue detection is this code is changed.

fdlk: The issue was that the field should be protected

fdlk: The security issue is that the field should be protected, I think?
https://sonarcloud.io/project/issues?id=org.molgenis%3Amolgenis&issues=AVzJsld5sY0y2VuVC9S0&open=AVzJsld5sY0y2VuVC9S0

dennishendriksen: Yes, because Sonar thinks it is a mutable field: see the '...' in Sonar for details.

fdlk: Ah, I see!

dennishendriksen: For clarity we could store Immutable* values in a Immutable* instead of *, but lets make that a different task.

molgenis-core-ui/src/main/java/org/molgenis/data/importer/wizard/OptionsWizardPage.java
fdlk: If it fails, that probably is because it already exists and someone else has already imported a file with that name. I think the whole renaming is very hairy here. Will file a separate issue for these things.

PR #6844, opened 2017-10-26
Fix #6800 Not able to enter multiple email addresses for scheduled jobs

### LINKED ISSUE #6800: Not able to enter multiple email addresses for scheduled jobs
### How to Reproduce

Enter a comma separated list of email addresses in the succes email or failure email field.

### Expected behavior

I can enter multiple email addresses, comma separated.
This is actually what the description of the field tells me: "Comma-separated list of emails. Leave blank if you don't want to receive emails if the jobs failed."

### Observed behavior

Only one email can be given. It gives an error message : "please enter a valid email address."

ConnorStroomberg:
change field type to String, the code processing the mail list expects a comma separated string

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


fdlk:
I want my generic value list type!

ConnorStroomberg:
@fdlk be careful what you wish for :)

But yes that would be nice, maybe ( maybe not). This PR fixes the issue for the user, the same data type String is used for other fields and the code handeling the data seems to expect a comma separated string list. The  'generic value list type' is a high impact story that is not 'ready to plan'.

Mark-de-Haan:
Breaking change: Can not alter system entity metadata, so drop your database!

fdlk:
@Mark-de-Haan We already have one of those for the null expressions so keep 'em coming! :)

dennishendriksen:
@ConnorStroomberg the single email address was validated, the multiple email addresses are not (typical use case, my value is 'you@molgenis.org;me@molgenis.org').

Validation can be restored by using a validation expression or better by introducing an entity type decorator for scheduled job that reuses existing email address validation code.

PR #6885, opened 2017-11-06
M5419 Add nullableExpression validation to forms

Mark-de-Haan:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


marikaris:
Works with simple questionnaire, but fails on the big one. I cannot find out why (it works if i extend the simple example from sys_Questionnaire and putting the expression item in a compound also doesn't reproduce it). Anyway, it's something in the backend, I will file the issue. The frondend is working correctly.

PR #6886, opened 2017-11-06
Fix Sonar bugs in data, data-validation and data-rest

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/EntityCollectionBatchRequestV2.java
```diff
@@ -17,6 +17,7 @@
 	public abstract List<Map<String, Object>> getEntities();
 
 	@Override
+	@SuppressWarnings("squid:S1206")
```
bartcharbon: To discuss with the team: would it be usefull to always accompany these suppressions with a comment stating what the identifier means?

Issue #2816, opened 2015-04-14
HTTP status 500 hiccups when Safari 8 or 9 surfs to molgenis.


fdlk:
When testing with multiple developers on a molgenis target vm, the server has more than once become unreachable.
When I send a kill -quit to the tomcat process, all threads look to be waiting for work.
However, Apache is sending out 500 errors with 1 minute delays in between.

Restarting both httpd and tomcat7 helps.

This has happened more than once.

fdlk:
In /etc/httpd/logs/error_log, repeatedly:
````
[Tue Apr 14 12:48:31 2015] [error] (70007)The timeout specified has expired: ajp_ilink_receive() can't receive header
````

In /etc/httpd/logs/ssl_error_log, repeatedly:
````
[Tue Apr 14 12:48:31 2015] [error] ajp_read_header: ajp_ilink_receive failed
[Tue Apr 14 12:48:31 2015] [error] (70007)The timeout specified has expired: proxy: read response failed from [::1]:8009 (localhost)
````

Threaddump was at 12:48:59.
Shows the full 200 ajp-bio-8009-exec threads all doing
````
"ajp-bio-8009-exec-200" #2776 daemon prio=5 os_prio=0 tid=0x00007f2a2c1a4800 nid=0x1b18 runnable [0x00007f298f1dd000]
   java.lang.Thread.State: RUNNABLE
        at java.net.SocketInputStream.socketRead0(Native Method)
        at java.net.SocketInputStream.read(SocketInputStream.java:150)
        at java.net.SocketInputStream.read(SocketInputStream.java:121)
        at org.apache.coyote.ajp.AjpProcessor.read(AjpProcessor.java:312)
        at org.apache.coyote.ajp.AjpProcessor.readMessage(AjpProcessor.java:367)
        at org.apache.coyote.ajp.AjpProcessor.process(AjpProcessor.java:118)
        at org.apache.coyote.AbstractProtocol$AbstractConnectionHandler.process(AbstractProtocol.java:611)
        at org.apache.tomcat.util.net.JIoEndpoint$SocketProcessor.run(JIoEndpoint.java:314)
        - locked <0x0000000085b4dc60> (a org.apache.tomcat.util.net.SocketWrapper)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[...]

PR #6948, opened 2017-11-23
Fix #6264 MOLGENIS documentation should warn that you should use the AJP protocol from proxies

### LINKED ISSUE #6264: MOLGENIS documentation should warn that you should use the AJP protocol from proxies
### How to Reproduce
- Go to the installation instruction for MOLGENIS in the manual

### Expected behavior
- The installation instructions should note special considerations such as requiring AJP if you use a proxy in front of Tomcat
- We should refer to AJP documentation such as http://httpd.apache.org/docs/2.4/mod/mod_proxy_ajp.html on how to configure things instead of writing extensive documentation

### Observed behavior
- The installation instructions do not tell this.

Mark-de-Haan:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/docs/quickstart/guide-tomcat.md
+++ b/docs/quickstart/guide-tomcat.md
@@ -79,5 +79,7 @@ If you open up a web browser and navigate to where your apache-tomcat applicatio
 
 Congratulations! You now have your MOLGENIS application up and running. Remember the admin.password you set in the molgenis-server.properties file? Use that password to login as the admin user. The next section will take you through the different modules MOLGENIS has to offer.
 
-**Proxying with apache httpd**
-When using a proxy in front of tomcat, make sure you use the [AJP protocols](http://httpd.apache.org/docs/2.4/mod/mod_proxy_ajp.html).
+**Deployment**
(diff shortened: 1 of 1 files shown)


marikaris:
There is now a sign in button, but I cannot click it haha.

fdlk:
In general, I think that a webapp shouldn't worry about how it's being deployed and only list the restrictions it makes on how it can be deployed. (In our case, things like that you must deploy it as the root app and that we include the jdbc postgres driver both of which are a bit of a nono.)

I don't have a clear overview of the advantages and benefits of using AJP when you proxy but I would expect that the benefits of using AJP are actually small. We've had an issue (#2816) with the protocol because it is more stateful than simple HTTP.

I would prefer to alert the sysops who deploy a molgenis to the option of using AJP than demand that they use it which would suggest that molgenis doesn't work properly without it. If they prefer to run it behind an existing nginx, for instance, that will work great as well and we shouldn't discourage them from it.

dennishendriksen:
I don't think this section should be added to our documentation since it is specific for the deployment environment.

dennishendriksen:
@fdlk AJP was introduced in our VM SOP late 2014 to resolve the issue where the web application was not able to determine the request URI (more specifically: the 'HTTP activation URL')

Mark-de-Haan:
Then we close this.

PR #6961, opened 2017-11-28
Feature/global exception handling

bartcharbon:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-cache/src/main/java/org/molgenis/data/cache/l2/L2Cache.java
```diff
@@ -115,11 +114,7 @@ public Entity get(Repository<Entity> repository, Object id)
 		catch (ExecutionException exception)
 		{
 			// rethrow unchecked
-			if (exception.getCause() != null && exception.getCause() instanceof RuntimeException)
-			{
-				throw (RuntimeException) exception.getCause();
-			}
-			throw new MolgenisDataException(exception);
+			throw new RuntimeException(exception);
```
fdlk: Why this change?

molgenis-data-i18n/src/main/java/org/molgenis/data/i18n/LocalizationService.java
```diff
@@ -163,15 +162,10 @@ public void addMissingMessageIDs(String namespace, Set<String> messageIDs)
...
-				LOG.debug("Added message IDs to namespace '{}' : {}.", namespace, messageIDs);
-			}
-			catch (MolgenisPermissionException ex)
-			{
-				LOG.info("No permission to add message IDs to namespace '{}' : {}.", namespace, messageIDs, ex);
-			}
+
+			dataService.add(L10N_STRING, entities);
+			LOG.debug("Added message IDs to namespace '{}' : {}.", namespace, messageIDs);
+
```
fdlk: If we remove the catch, the method will suddenly begin to throw runtime exceptions it previously nommed up. Is this intentional?

molgenis-data-cache/src/main/java/org/molgenis/data/cache/l2/L2Cache.java
bartcharbon: MolgenisDataException is deprecated, it's not something to show to the user, so RuntimeException would be fine I think

molgenis-data-i18n/src/main/java/org/molgenis/data/i18n/LocalizationService.java
bartcharbon: Ahum.... cannot remember removing this... will revert

PR #6964, opened 2017-11-28
Add FormatFactories for {0,label} and {1,id} MessageFormats and let L…

fdlk:
…anguageService create them.

#### Checklist
- [ ] Functionality works & meets specifications
- [ ] Code reviewed
- [ ] Code unit/integration/system tested
- [ ] User documentation updated
- [ ] (If you have changed REST API interface) view-swagger.ftl updated
- [ ] Test plan template updated
- [ ] Clean commits

PR #6965, opened 2017-11-28
Use getLocalizedMessageArguments to ditch boilerplate code in getLocalizedMessage method of exceptions.

fdlk:
Beware: Contains #6964 and #6961 

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-security/src/main/java/org/molgenis/data/security/exception/PluginPermissionDeniedException.java
```diff
@@ -0,0 +1,53 @@
...
+		return getLanguageService().map(languageService ->
+		{
+			String permissionName = getPermissionName(languageService, permission);
+			MessageFormat format = languageService.getMessageFormat(ERROR_CODE);
+			return format.format(new Object[] { permissionName, pluginId });
+		}).orElseGet(super::getLocalizedMessage);
+	}
+
+	@Override
+	protected Object[] getLocalizedMessageArguments()
```
tommydeboer: Why isn't it possible to construct the localized message arguments here by using getPermissionName()?

molgenis-data/src/main/java/org/molgenis/data/UnknownDataAccessException.java
```diff
@@ -1,11 +1,10 @@
 package org.molgenis.data;
 
-import static org.molgenis.data.i18n.LanguageServiceHolder.getLanguageService;
-
 /**
  * TODO discuss: extend from UncategorizedDataAccessException?
  */
-public class UnknownDataAccessException extends DataAccessException
+@SuppressWarnings("squid:S2166")
```
tommydeboer: Explain what is suppressed here

molgenis-data-security/src/main/java/org/molgenis/data/security/exception/PluginPermissionDeniedException.java
fdlk: Because getPermissionName needs the LanguageService to resolve the permission name.
The Spring way is to use MessageSourceResolvable as an argument, then the MessageSource will first format the argument and then the message.

molgenis-data/src/main/java/org/molgenis/data/UnknownDataAccessException.java
fdlk: will do!

PR #7019, opened 2017-12-15
Specify supported R-version in documentation

sidohaakma:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/docs/developer_documentation/ref-R.md
+++ b/docs/developer_documentation/ref-R.md
@@ -13,7 +13,7 @@ at the top of your script and you can connect to a MOLGENIS server. Typically th
 eval(expr = parse(text = getURL("https://molgenis.mydomain.example/molgenis.R?molgenis-token=${molgenisToken}")))
 ```  
 
-**NOTE:** We support R-versions earlier or equal to 3.2.x
+**NOTE:** The MOLGENIS R-api client supports up to R-version 3.2.x
 
 # Overview example
 
--- a/docs/developer_documentation/ref-R.md
+++ b/docs/developer_documentation/ref-R.md
@@ -8,10 +8,12 @@ source("http://molgenis.mydomain.example/molgenis.R")
 ``` 
 at the top of your script and you can connect to a MOLGENIS server. Typically the first thing you do is login and the last thing is logout.
 
-NOTE: For https connections use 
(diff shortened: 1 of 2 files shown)


dennishendriksen:
Does https://github.com/molgenis/docker meet this requirement?

sidohaakma:
No, opencpu/base (the latest runs on R 3.4.1). Maybe we should specify a version of the opencpu docker?

dennishendriksen:
How about requiring R 3.4.x instead of 3.2 for MOLGENIS?

sidohaakma:
That is a good idea. We need to perform a 2 steps then:
- require the bit (this package contains the method chunk in R 3.4.1) in the molgenis.R script
- check if the package is available on opencpu (golden vm and docker)

dennishendriksen:
FYI from http://www.molgenis.org/wiki/GccSopCreateMolgenisVM:
```
# install R libraries
Rscript -e 'install.packages(c("bitops", "RCurl", "rjson", "gdata", "corrplot", "qqman", "ggplot2", "ff", "httr"), repos="http://cran-mirror.cs.uu.nl/")'
```

sidohaakma:
Now concluded we need to add to the documentation that only the molgenis.R client is comptabli with R-versions up to 3.2.1.

docs/developer_documentation/ref-R.md
```diff
@@ -8,11 +8,13 @@ source("http://molgenis.mydomain.example/molgenis.R")
 ``` 
 at the top of your script and you can connect to a MOLGENIS server. Typically the first thing you do is login and the last thing is logout.
 
-NOTE: For https connections use 
+**NOTE:** For https connections use 
 ```
 eval(expr = parse(text = getURL("https://molgenis.mydomain.example/molgenis.R?molgenis-token=${molgenisToken}")))
 ```  
 
+**NOTE:** We support R-versions earlier or equal to 3.2.x
```
fdlk: Could you make it a little bit clearer that this specifically refers to the R api client?

fdlk:
@sidohaakma what's the status of this one?

sidohaakma:
@dennishendriksen I have specified the R-api supported R version in the documentation.

PR #7024, opened 2017-12-18
M5327 Improve controller tests and autowire metadata for exceptions

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-apps/src/test/java/org/molgenis/apps/AppsControllerTest.java
```diff
@@ -196,23 +191,21 @@ public void testActivateApp() throws Exception
 	}
 
 	@Test(expectedExceptions = AppsException.class, expectedExceptionsMessageRegExp = "Unknown app 'id'")
-	public void testActivateAppUnknownApp() throws Exception
```
fdlk: Let's downgrade the performer's signature to throw Exception instead of Throwable.

tommydeboer: After discussion with @fdlk we decided to keep it 'as is' because ```NestedServletException.getCause()``` returns a Throwable

PR #7025, opened 2017-12-18
 Fix typo's, missing curlies, improve comments, rename classes, add dots, etc.

bartcharbon:
#### Checklist
- (n.a.) Functionality works & meets specifications
- [x] Code reviewed
- (n.a.) Code unit/integration/system tested
- (n.a.) User documentation updated
- (n.a.) Clean commits


molgenis-data-rest/src/main/java/org/molgenis/data/rest/exception/IncompatibleValueTypeException.java
```diff
@@ -48,6 +48,6 @@ public String getMessage()
 	@Override
 	protected Object[] getLocalizedMessageArguments()
 	{
-		return new Object[] { attribute.getName(), type, String.join(",", expectedTypes) };
+		return new Object[] { attribute.getName(), type, String.join("or", expectedTypes) };
```
tommydeboer: `" or "` with the spaces me thinks

tommydeboer: Also, `or` isn't localized here. If we really want to do that I guess we'd have to add another message formatter?

fdlk: or a key for the delimiter

PR #7027, opened 2017-12-18
M4797 Dynamic Decorators

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-platform/src/main/java/org/molgenis/data/platform/decorators/MolgenisRepositoryDecoratorFactory.java
```diff
@@ -95,8 +103,15 @@ public MolgenisRepositoryDecoratorFactory(EntityManager entityManager,
 		// 10. Register the cud action needed to index indexed repositories
 		decoratedRepository = new IndexActionRepositoryDecorator(decoratedRepository, indexActionRegisterService);
 
-		// 9. Custom decorators
-		decoratedRepository = repositoryDecoratorRegistry.decorate(decoratedRepository);
+		// 9. Custom & dynamic decorators
+		if (systemEntityTypeRegistry.hasSystemEntityType(repository.getEntityType().getId()))
```
tommydeboer: Fix this workaround for a bootstrapping issue. It should be possible to wrap ```SystemEntityType``` with ```DynamicDecorator```.

molgenis-data-platform/src/main/java/org/molgenis/data/platform/decorators/MolgenisRepositoryDecoratorFactory.java
```diff
@@ -80,23 +84,26 @@ public MolgenisRepositoryDecoratorFactory(EntityManager entityManager,
...
 		decoratedRepository = new L3CacheRepositoryDecorator(decoratedRepository, l3Cache, transactionInformation);
 
-		// 10. Register the cud action needed to index indexed repositories
+		// 11. Register the cud action needed to index indexed repositories
 		decoratedRepository = new IndexActionRepositoryDecorator(decoratedRepository, indexActionRegisterService);
 
-		// 9. Custom decorators
-		decoratedRepository = repositoryDecoratorRegistry.decorate(decoratedRepository);
+		// 10. Dynamic decorators
+		decoratedRepository = dynamicRepositoryDecoratorRegistry.decorate(decoratedRepository);
```
dennishendriksen: Could you explain why the location of the dynamic decorators in the chain? Did you consider the security impact e.g. in case of an `evil` decorator?

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorFactory.java
```diff
@@ -0,0 +1,24 @@
+package org.molgenis.data.decorator;
+
+import org.molgenis.data.Entity;
+import org.molgenis.data.Repository;
+import org.molgenis.data.meta.model.EntityType;
+
+/**
+ * Repository decorator factory that creates decorated {@link Repository repositories} for specific entity types.
+ * Used to decorate (system) entity types dynamically.
```
dennishendriksen: Would it make sense to merge system repository decorators with dynamic decorators?

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorFactoryRegistrar.java
```diff
@@ -0,0 +1,28 @@
...
+public class DynamicRepositoryDecoratorFactoryRegistrar
+{
+	private final DynamicRepositoryDecoratorRegistry repositoryDecoratorRegistry;
+
+	public DynamicRepositoryDecoratorFactoryRegistrar(DynamicRepositoryDecoratorRegistry repositoryDecoratorRegistry)
+	{
+		this.repositoryDecoratorRegistry = requireNonNull(repositoryDecoratorRegistry);
+	}
+
+	public void register(ContextRefreshedEvent event)
```
dennishendriksen: Pass a ```ApplicationContext``` instead of ```ContextRefreshedEvent``` since you are only using the event to get the context.

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorRegistry.java
```diff
@@ -0,0 +1,15 @@
+package org.molgenis.data.decorator;
+
+import org.molgenis.data.Entity;
+import org.molgenis.data.Repository;
+
+import java.util.stream.Stream;
+
+public interface DynamicRepositoryDecoratorRegistry
+{
+	void addFactory(DynamicRepositoryDecoratorFactory factory);
```
dennishendriksen: Does ```dynamic``` imply that you should also be able to remove a factory?

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorRegistryImpl.java
```diff
@@ -0,0 +1,90 @@
...
+import static java.util.stream.Collectors.toList;
+import static org.molgenis.data.decorator.meta.DecoratorConfigurationMetadata.DECORATOR_CONFIGURATION;
+import static org.molgenis.data.decorator.meta.DecoratorConfigurationMetadata.ENTITY_TYPE_ID;
+
+@Component
+public class DynamicRepositoryDecoratorRegistryImpl implements DynamicRepositoryDecoratorRegistry
+{
+	private final Map<String, DynamicRepositoryDecoratorFactory> factories = new HashMap<>();
+	private final DataService dataService;
+	private RepositoryCollection repositoryCollection;
```
dennishendriksen: Can be final?

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorRegistryImpl.java
```diff
@@ -0,0 +1,90 @@
...
+	private DecoratorConfigurationMetadata decoratorConfigurationMetadata;
```
dennishendriksen: Can be final?

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorRegistryImpl.java
```diff
@@ -0,0 +1,90 @@
...
+	{
+		this.dataService = requireNonNull(dataService);
+		this.repositoryCollection = requireNonNull(repositoryCollection);
+		this.decoratorConfigurationMetadata = requireNonNull(decoratorConfigurationMetadata);
+	}
+
+	@Override
+	public synchronized void addFactory(DynamicRepositoryDecoratorFactory factory)
+	{
+		String factoryId = factory.getName();
```
dennishendriksen: Consistency: factory.getIdentifier()/getId() instead of getName()

molgenis-data/src/main/java/org/molgenis/data/decorator/example/SpyingRepositoryDecorator.java
```diff
@@ -0,0 +1,30 @@
+package org.molgenis.data.decorator.example;
+
+import org.molgenis.data.AbstractRepositoryDecorator;
+import org.molgenis.data.Entity;
+import org.molgenis.data.Repository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SpyingRepositoryDecorator extends AbstractRepositoryDecorator<Entity>
```
dennishendriksen: Move examples to test folder?

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DecoratorConfigurationMetadata.java
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	@Override
+	public void init()
+	{
+		setLabel("Decorator Configuration");
+		setPackage(decoratorPackage);
+
+		addAttribute(ID, ROLE_ID).setAuto(true).setVisible(false).setLabel("Identifier");
+		addAttribute(ENTITY_TYPE_ID).setNillable(false).setUnique(true).setLabel("Entity Type Identifier");
```
dennishendriksen: Shouldn't this be a XREF to EntityType to avoid dynamic decorators pointing to entity types that don't exist?

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DecoratorConfigurationMetadata.java
```diff
@@ -0,0 +1,45 @@
...
+		super(SIMPLE_NAME, PACKAGE_DECORATOR);
+		this.decoratorPackage = requireNonNull(decoratorPackage);
+		this.dynamicDecoratorMetadata = requireNonNull(dynamicDecoratorMetadata);
+	}
+
+	@Override
+	public void init()
+	{
+		setLabel("Decorator Configuration");
+		setPackage(decoratorPackage);
```
dennishendriksen: Could use a setDescription

molgenis-data-platform/src/main/java/org/molgenis/data/platform/decorators/MolgenisRepositoryDecoratorFactory.java
dennishendriksen: Security: as a user can a decorate a repository that I don't have permission to read and e.g. email all traffic to my mail address?

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DynamicDecoratorMetadata.java
```diff
@@ -0,0 +1,35 @@
...
+		this.decoratorPackage = requireNonNull(decoratorPackage);
+	}
+
+	@Override
+	public void init()
+	{
+		setLabel("Dynamic Decorator");
+		setPackage(decoratorPackage);
+
+		addAttribute(ID, ROLE_ID).setLabel("Identifier");
```
dennishendriksen: How about adding a label attribute?

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DynamicDecoratorPopulator.java
```diff
@@ -0,0 +1,88 @@
...
+	private void removeNonExistingDecorators()
+	{
+		Set<Object> nonExistingDecorators = getNonExistingDecorators();
+		updateDecoratorConfigurations(nonExistingDecorators);
+		dataService.deleteAll(DYNAMIC_DECORATOR, nonExistingDecorators.stream());
+	}
+
+	private Set<Object> getNonExistingDecorators()
+	{
+		Set<Object> decorators = dataService.findAll(DYNAMIC_DECORATOR).map(Entity::getIdValue).collect(toSet());
```
dennishendriksen: Better: dataService.findAll(DYNAMIC_DECORATOR, DynamicDecorator.class).map(DynamicDecorator::getId)

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DynamicDecoratorPopulator.java
```diff
@@ -0,0 +1,88 @@
...
+		decorators.removeAll(registry.getFactoryIds().collect(toSet()));
+		return decorators;
+	}
+
+	private void updateDecoratorConfigurations(Set<Object> nonExistingDecorators)
+	{
+		Stream<Entity> updatedEntities = dataService.findAll(DECORATOR_CONFIGURATION)
```
dennishendriksen: Better: use dataService.findAll(DECORATOR_CONFIGURATION, DecoratorConfiguration.class)

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DynamicDecoratorPopulator.java
```diff
@@ -0,0 +1,88 @@
...
+													.map(entity -> removeReferences(nonExistingDecorators, entity));
+		dataService.update(DECORATOR_CONFIGURATION, updatedEntities);
+	}
+
+	private Entity removeReferences(Set<Object> nonExistingDecorators, Entity entity)
+	{
+		List<Entity> decorators = StreamSupport.stream(entity.getEntities(DYNAMIC_DECORATORS).spliterator(), false)
```
dennishendriksen: Better: entity.getEntities(DYNAMIC_DECORATORS, DynamicDecorator.class)

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DynamicDecoratorPopulator.java
```diff
@@ -0,0 +1,88 @@
...
+	{
+		return dataService.findOneById(DYNAMIC_DECORATOR, id, DynamicDecorator.class) == null;
+	}
+
+	private DynamicDecorator createDecorator(String id)
+	{
+		return dynamicDecoratorFactory.create(id);
+	}
+}
+
```
dennishendriksen: Remove newline

molgenis-data-platform/src/main/java/org/molgenis/data/platform/decorators/MolgenisRepositoryDecoratorFactory.java
tommydeboer: In theory: yes, but that would imply two things:
1. The admin gave a user permission to edit the configuration entity, which he/she shouldn't do because it's a high-level admin feature.
2. The evil user managed to add a malicious decorator to the code base and deployed the new artifact to the server without the admins credentials.

tommydeboer: Still under consideration. At least the system entity type specific decorators should go first because of the row level security on metadata.

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorFactory.java
tommydeboer: I've thought about it but that would mean that it would be possible to remove system critical decorators from system entities, among other things. The way I see it, 'custom' decorators are specific to one SystemEntityType and 'dynamic' decorators should be generally useful for more than one EntityType.

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorRegistry.java
tommydeboer: Maybe the words are in the wrong order here 😄 . The decorators are dynamic, the factories aren't.

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DecoratorConfigurationMetadata.java
tommydeboer: We don't want to change EntityType and we don't have cascading deletes yet so we decided to do it like this in the meantime. I've added a story for cascading deletes:

http://www.molgenis.org/ticket/5565#ticket

tommydeboer: Having a dangling decorator won't break anything as it will never be added or looked up.

molgenis-data/src/main/java/org/molgenis/data/decorator/example/SpyingRepositoryDecorator.java
tommydeboer: Will be removed or moved to integration test suite.

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorFactory.java
dennishendriksen: > but that would mean that it would be possible to remove system critical decorators from system entities

Can be prevented using a check similar to EntityTypeRepositorySecurityDecorator.validateDeleteAllowed

molgenis-data-platform/src/main/java/org/molgenis/data/platform/decorators/MolgenisRepositoryDecoratorFactory.java
dennishendriksen: It gets trickier once we add a dynamic decorator that executes scripts that can be modified by the user (side note: script executing decorator, sounds like ultimate flexibility). Why not execute your decorator right after the repository call to prevent any security issues?

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorRegistry.java
dennishendriksen: :-D

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DecoratorConfigurationMetadata.java
dennishendriksen: But we don't want dangling stuff right? If we can't use an XREF then we should do the cleanup ourselves.

molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorFactory.java
tommydeboer: I'm not sure of the benefit of introducing extra logic whilst exposing the inner workings of system decorators to admins/users who won't be able to change or use them anyways. System decorators really are different from dynamic decorators if you ask me and it's best to keep them separated.

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DecoratorConfigurationMetadata.java
tommydeboer: I could add an extra system decorator to EntityType to manage this cleanup, or wait for the cascading delete feature to be implemented. Either's fine by me.

molgenis-data-platform/src/main/java/org/molgenis/data/platform/decorators/MolgenisRepositoryDecoratorFactory.java
tommydeboer: That would make decorators that change values impossible I think?

Letting users edit scripts that are triggered with system permissions by a decorator is something you shouldn't do if you ask me. But yes, it is possible to write 'dangerous' decorators and admins should know what they are doing.

dennishendriksen: > That would make decorators that change values impossible I think?

Why? You should have permissions to change the values of course.

> Letting users edit scripts that are triggered with system permissions by a decorator is something you shouldn't do if you ask me

Too dynamic? ;-D

molgenis-data/src/main/java/org/molgenis/data/decorator/meta/DecoratorPackage.java
```diff
@@ -0,0 +1,32 @@
...
+	public DecoratorPackage(PackageMetadata packageMetadata, RootSystemPackage rootSystemPackage)
+	{
+		super(PACKAGE_DECORATOR, packageMetadata);
+		this.rootSystemPackage = requireNonNull(rootSystemPackage);
+	}
+
+	@Override
+	protected void init()
+	{
+		setLabel("System");
```
bartcharbon: Should be decorator or something similar?

molgenis-data/src/main/java/org/molgenis/data/decorator/example/SpyingRepositoryDecorator.java
bartcharbon: When moved to the test folder we do not have a way to demo and test this since none of the planned dynamic decorators are build yet

dennishendriksen: Demo by executing a test or locally with a dummy dynamic decorator.

PR #7030, opened 2017-12-19
Fix #7029 Make RepositoryCollection.createRepository void

### LINKED ISSUE #7029: RepositoryCollection.createRepository shouldn't return the created repository
Open `RepositoryCollection` and inspect:

```java
/**
* Creates a new {@link Repository} within this repository collection for the given entity meta data.
*
* @param entityType entity meta data
* @return the created repository
* @throws UnsupportedOperationException if this repository collection is not {@link RepositoryCollectionCapability#WRITABLE}
*/
Repository<Entity> createRepository(EntityType entityType);
```

The result of this method is never used anywhere. Also, during bootstrapping when you are interested in creating a repository, instantly getting it (which involves decorating it) can result in issues. 

Change the method signature to void.

tommydeboer:
Fix for #7029 

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data-index/src/main/java/org/molgenis/data/index/IndexActionRepositoryCollectionDecorator.java
+++ b/molgenis-data-index/src/main/java/org/molgenis/data/index/IndexActionRepositoryCollectionDecorator.java
@@ -1,7 +1,6 @@
 package org.molgenis.data.index;
 
 import org.molgenis.data.AbstractRepositoryCollectionDecorator;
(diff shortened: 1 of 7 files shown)


dennishendriksen:
In my opinion a ```create``` method should always return the created instance: in this instance the decorated instance. I would expect the return value to be used in the future and thus think that suppressing the unused return value warning is a better solution in this case.

PR #7037, opened 2017-12-22
Fix #7036: Handler methods should be able to bind to the molgenis tok…

### LINKED ISSUE #7036: Handler methods should be able to bind to the molgenis token as a method parameter
So
```
@GetMapping(URI)
public String init(@TokenParam String token) {
...
}
```

fdlk:
…en as a method parameter

#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-data-rest/src/main/java/org/molgenis/data/rest/RestController.java
+++ b/molgenis-data-rest/src/main/java/org/molgenis/data/rest/RestController.java
@@ -24,7 +24,7 @@ import org.molgenis.security.core.PermissionService;
 import org.molgenis.security.core.token.TokenService;
 import org.molgenis.security.core.token.UnknownTokenException;
 import org.molgenis.security.settings.AuthenticationSettings;
-import org.molgenis.security.token.TokenExtractor;
+import org.molgenis.security.token.TokenParam;
 import org.molgenis.security.user.UserAccountService;
 import org.molgenis.util.ErrorMessageResponse;
 import org.molgenis.util.ErrorMessageResponse.ErrorMessage;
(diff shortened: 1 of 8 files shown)


molgenis-security/src/main/java/org/molgenis/security/token/TokenParam.java
```diff
@@ -0,0 +1,29 @@
...
+ * @GetMapping("/blah")
+ * public void getBlah(@TokenParam String token){
+ * ...
+ * }
+ * </pre>
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface TokenParam
```
dennishendriksen: Add documentation why this annotation is preferable above Spring data binder.

molgenis-data-rest/src/main/java/org/molgenis/data/rest/RestController.java
```diff
@@ -798,14 +798,8 @@ private boolean isUser2fa()
 
 	@PostMapping("/logout")
 	@ResponseStatus(OK)
-	public void logout(HttpServletRequest request)
+	public void logout(@TokenParam(required = true) String token, HttpServletRequest request)
```
dennishendriksen: ```POST http://localhost:8080/api/v1/logout``` with header ```x-molgenis-token``` <myToken> results in token being ```null```.

fdlk: Not good enough apperantly to make this a @Component.

dennishendriksen:
@fdlk approved, reviewed and tested. can be merged by you or someone else once Travis goes green.

PR #7050, opened 2018-01-09
Global exception handling framework

dennishendriksen:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-web/src/main/resources/templates/view-exception.ftl
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>MOLGENIS</title>
+        <meta charset="utf-8">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta http-equiv="X-UA-Compatible" content="chrome=1">
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+        <link rel="icon" href="<@resource_href "/img/favicon.ico"/>" type="image/x-icon">
```
bartcharbon: The ftl rendering errors because of the resource href.
changing it to:
href="/img/favicon.ico"
works, however the page contains than only the words "Unknown Error" I would expect at least the mention of a http status code (always 500 in this PR I think)

bartcharbon: will be fixed in follow up PR

PR #7056, opened 2018-01-11
M5514:  Add UnknownEntityTypeException and let the global exception handler handle it.

fdlk:
#### Checklist
- [x] Code reviewed
- [ nvt ] User documentation updated
- [ nvt ] (If you have changed REST API interface) view-swagger.ftl updated
- [ nvt ] Test plan template updated
- [x] Clean commits

### DIFF
--- a/molgenis-core-ui/src/main/resources/js/molgenis.js
+++ b/molgenis-core-ui/src/main/resources/js/molgenis.js
@@ -63,7 +63,10 @@
         $.each(alerts, function (i, alert) {
             if (i > 0)
                 items.push('<br/>');
-            items.push('<span>' + alert.message + '</span>');
+            var message = '<span>' + alert.message;
+            if (alert.code) message += ' (' + alert.code + ')';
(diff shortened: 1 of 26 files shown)


molgenis-core-ui/src/main/resources/js/molgenis.js
```diff
@@ -63,7 +63,10 @@
         $.each(alerts, function (i, alert) {
             if (i > 0)
                 items.push('<br/>');
-            items.push('<span>' + alert.message + '</span>');
+            var message = '<span>' + alert.message;
+            if (alert.code) message += ' (' + alert.code + ')';
+            message += '</span>';
+            items.push(message)
```
bartcharbon: The open PR from Dennis html escapes this, but I think that breaks the importer exception messages... Leave unescaped for now?

fdlk: We also still have open issue if we want to show it at all.

PR #7142, opened 2018-02-02
M4797 Dynamic repository decorators

bartcharbon:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


molgenis-data/src/main/java/org/molgenis/data/event/BootstrappingEvent.java
```diff
@@ -0,0 +1,16 @@
+package org.molgenis.data.event;
+
+public class BootstrappingEvent
```
tommydeboer: Maybe we can generify this a bit to support multiple bootstrapping 'states'? With an Enum for the states for example. `STARTED`, `POPULATING`, `FINISHED`.. something like that.

molgenis-data/src/main/java/org/molgenis/data/event/BootstrappingEventPublisher.java
```diff
@@ -0,0 +1,22 @@
...
+{
+	private final ApplicationEventPublisher publisher;
+
+	@Autowired
+	public BootstrappingEventPublisher(ApplicationEventPublisher publisher)
+	{
+		this.publisher = publisher;
+	}
+
+	public void createBoostrapEvent(boolean done)
```
tommydeboer: rename to `publishBootstrappingFinishedEvent()` and remove parameter to make intention clearer?

PR #7151, opened 2018-02-05
[Feature] Dynamic entities use parent entity factory

mark-de-haan:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data/src/main/java/org/molgenis/data/EntityManagerImpl.java
+++ b/molgenis-data/src/main/java/org/molgenis/data/EntityManagerImpl.java
@@ -81,13 +81,29 @@ public class EntityManagerImpl implements EntityManager
 			entityPopulator.populate(entity);
 		}
 
-		EntityFactory<? extends Entity, ?> entityFactory = entityFactoryRegistry.getEntityFactory(entityType);
+		EntityFactory<? extends Entity, ?> entityFactory = resolveEntityFactory(entityType);
 		if (entityFactory != null)
 		{
 			// create static entity (e.g. Tag, Language, Package) that wraps the constructed dynamic or partial entity.
 			return entityFactory.create(entity);
 		}
+
 		return entity;
+	}
+
+	private EntityFactory<? extends Entity, ?> resolveEntityFactory(EntityType entityType)
+	{
+		EntityFactory<? extends Entity, ?> entityFactory = entityFactoryRegistry.getEntityFactory(entityType);
+		if (entityFactory == null)
+		{
+			// In case of dynamic entities extending a StaticEntity (e.g. Questionnaires), use the parent entityFactory
+			EntityType parent = entityType.getExtends();
+			if (parent != null)
+			{
+				return resolveEntityFactory(parent);
+			}
+		}
+		return entityFactory;
 	}
 
 	@Override
--- a/molgenis-data/src/test/java/org/molgenis/data/EntityManagerImplTest.java
+++ b/molgenis-data/src/test/java/org/molgenis/data/EntityManagerImplTest.java
@@ -11,8 +11,10 @@ import java.util.Arrays;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static com.google.common.collect.Lists.newArrayList;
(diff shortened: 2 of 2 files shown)


molgenis-data/src/main/java/org/molgenis/data/EntityManagerImpl.java
```diff
@@ -81,15 +81,31 @@ private Entity create(EntityType entityType, Fetch fetch, CreationMode creationM
...
 		if (entityFactory != null)
 		{
 			// create static entity (e.g. Tag, Language, Package) that wraps the constructed dynamic or partial entity.
 			return entityFactory.create(entity);
 		}
+
 		return entity;
 	}
 
+	private EntityFactory<? extends Entity, ?> resolveEntityFactory(EntityType entityType)
```
dennishendriksen: Could you elaborate why the original Questionnaire entity factory is incorrect? This change feels wrong to me.

dennishendriksen: I think I see what is going wrong: you are missing a 'Questionnaire extends StaticEntity' class as well as a 'QuestionnaireFactory extends AbstractSystemEntityFactory' (see developer guidelines 7.4).

dennishendriksen: Request changes

Mark-de-Haan: Yes so what happens is: I have a Questionnaire.class with some getters and setters for label description etc.. + a Factory and metadata.

But QuestionnaireMetadata is abstract. So in practice I am dealing with the EntityType testQuestionnaire which extends from QuestionnaireMetadata. 

But when I am ordering EntityTypes from the dataservice which extends sys_Questionnaire, I want them to be of Questionnaire.class so I can make use of all the nice little getters and setters it offers me.

The problem I fix here:
When I try to retrieve testQuestionnaire as type Questionnaire.class, it will tell me that it is a DynamicEntity and I can not cast it. This is a special case in which the StaticEntity that is being extended does have a factory, but the dynamic entity does not.

Mark-de-Haan: > I think I see what is going wrong: you are missing a 'Questionnaire extends StaticEntity' class as well as a 'QuestionnaireFactory extends AbstractSystemEntityFactory'

I have these on a separate feature branch, but I put this up as a standalone PR

dennishendriksen: I think I understand. In that case wouldn't it make more sense:
- to apply your changes to the registry, since you always want to retrieve the 'best' abstract factory for your entity type.
- to look for entity factories recursively: e.g. if no factory can be found for Questionnaire, look for a factory for OwnedEntityType

Mark-de-Haan: I guess discussing this IRL might be easier, but to answer your question:

- public class Questionnaire extends StaticEntity
- public class QuestionnaireMetaData extends SystemEntityType
- public class QuestionnaireFactory extends AbstractSystemEntityFactory<Questionnaire, QuestionnaireMetaData, String>

dataService.findOne(entityTypeId, 
EQ(OWNER_USERNAME, getCurrentUsername()), Questionnaire.class)

Where entityTypeId is my testQuestionnaire (a regular entityType) which extends sys_Questionnaire

fdlk:
Tricky! Questionnaire is an abstract system entity type that'll be subclassed by non-system entity types at runtime. I think our current factory interface was not made with this case in mind.

Two of the three create() methods in the factory interface make no sense. Creating and populating a new instance using the factory is pointless, if you cannot specify the concrete subclass to use.

Having the factory wrap a retrieved dynamic entity with a static entity wrapper when you're reading from the repository does make sense, and is actually useful, since it allows you to use the typed getters and setters.

mark-de-haan:
Closing due to inactivity, and no real necessity for this feature at this time.

PR #7168, opened 2018-02-14
M5634 Security: use default AnonymousAuthenticationFilter

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


bartcharbon:
1) started with empty db and index
2) when I surf to molgenis, I can see the home plugin
3) login as admin and give the anonymous user permissions on some plugins and entities via the permissionmanager
4) logout -> login dialog is shown instead of the home plugin, log states "o.m.core.ui.MolgenisMenuController - main menu does not contain any (accessible) items"
5) login as admin and view the permissions in the permissionmanager: all plugins and entities that were given permission now show 2 lines:
1) one active row with "none"
2) "inherited from group: undefined" with the permissions given in step 3

PR #7243, opened 2018-03-22
Fix #7184 When I upload dataset with wrong mref value I get wrong error message

### LINKED ISSUE #7184: When I upload dataset with wrong mref value I get wrong error message
### How to Reproduce
Contact me for all data
Upload it

Interesting fact: I generated a test dataset, with one row and the same incorrect value in the mref field and here the error message is correct:
[eric_de_test_model_correct_error.xlsx](https://github.com/molgenis/molgenis/files/1744363/eric_de_test_model_correct_error.xlsx)

### Expected behavior
Error message like: incorrect value referenced in mref diagnosis_available

### Observed behavior
Error message:
```; nested exception is org.postgresql.util.PSQLException: ERROR: value too long for type character varying(255)```

tommydeboer:
#### Checklist
- [x] Code reviewed
- [na] User documentation updated
- [will squash] Clean commits


bartcharbon:
I would like to have some additional information added to the message. In PostgreSqlRepository.java we know for which attribute the error occurs

tommydeboer:
Yes, but this information is not available when the PSQL exception occurs.  We could add extra validation rules (extra overhead), but decided to go for the translation because of the rarity of this issue: it only occurs in a single dataset. Usually you would get a 'key doesn't exist for attribute x' error.

PR #7245, opened 2018-03-23
Fix #6969 Internal server error when idAttribute of refEntity is visible == false and no other labelAttribute specified

### LINKED ISSUE #6969: Internal server error when idAttribute of refEntity is visible == false and no other labelAttribute specified
### How to Reproduce
Create EMX model with one entity referencing to another, where idAttribute of the reference entity is set on visible == false (and no other attribute is specified as labelAttribute)

### Expected behavior
The importer will give a warning or error

### Observed behavior
The EMX is correctly uploaded, but you get an internal server error 500 when selecting the entity

-Marieke has an example if this is not clear enough ;-)

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/EntityTypeValidator.java
+++ b/molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/EntityTypeValidator.java
@@ -150,15 +150,15 @@ public class EntityTypeValidator
 			if (ownAttr == null)
 			{
 				throw new MolgenisValidationException(new ConstraintViolation(
-						format("Entity [%s] ID attribute [%s] is not part of the entity attributes", entityType.getId(),
-								ownIdAttr.getName())));
+						format("EntityType [%s] ID attribute [%s] is not part of the entity attributes",
+								entityType.getId(), ownIdAttr.getName())));
 			}
 
 			// Validate that ID attribute data type is allowed
 			if (!AttributeUtils.isIdAttributeTypeAllowed(ownIdAttr))
 			{
 				throw new MolgenisValidationException(new ConstraintViolation(
-						format("Entity [%s] ID attribute [%s] type [%s] is not allowed", entityType.getId(),
+						format("EntityType [%s] ID attribute [%s] type [%s] is not allowed", entityType.getId(),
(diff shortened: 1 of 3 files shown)


bartcharbon:
Since the check is in the "if (ownIdAttr != null)" branch of the code, it does not work correctly when abstract entities come into play.

For example:
I cannot upload a abstract entity with a hidden ID and a concrete entity which has a visible label and extends the abstract one...

molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/EntityTypeValidator.java
```diff
@@ -150,33 +150,41 @@ private static void validateOwnIdAttribute(EntityType entityType, Map<String, At
...
 			if (ownIdAttr.isNillable())
 			{
 				throw new MolgenisValidationException(new ConstraintViolation(
-						format("Entity [%s] ID attribute [%s] is not a non-nillable attribute", entityType.getId(),
+						format("EntityType [%s] ID attribute [%s] is not a non-nillable attribute", entityType.getId(),
 								ownIdAttr.getName())));
 			}
+
+			// Validate that there is a label if the ID attribute is hidden
+			if (!ownIdAttr.isVisible() && entityType.getLabelAttribute() == null)
```
bartcharbon: Since the check is in the "if (ownIdAttr != null)" branch of the code, it does not work correctly when abstract entities come into play.

For example:
I cannot upload an abstract entity with a hidden ID and a concrete entity which has a visible label and extends the abstract one.

molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/EntityTypeValidator.java
```diff
@@ -150,33 +150,41 @@ private static void validateOwnIdAttribute(EntityType entityType, Map<String, At
...
+			{
+				throw new MolgenisValidationException(new ConstraintViolation(
+						format("The ID attribute of EntityType [%s] can not be hidden because there is no label attribute specified",
```
bartcharbon: not sure if EntityType should be capitalized

PR #7283, opened 2018-04-05
Fix #7254: Cannot update attributes for an abstract entity that is ex…

### LINKED ISSUE #6729: When adding a compound attribute to a system entity the attribute is added twice
## How to reproduce

* Start the app with a clean database
* Stop the app
* In the DataExplorerSettings addGeneralSettings() method, add the following code at the end:
```java
addAttribute("testAttribute").setDataType(BOOL)
.setNillable(true)
.setDefaultValue("true")
.setLabel("Attribute to show issue");
```
* Start the app
* Go to the data explorer and open the settings

## Expected behaviour

There is one "Attribute to show issue" option with a Radio button

## Observed behaviour

"Attribute to show issue" is added twice
If you save, everything is completely broken

Note

It is triggered by the copy event in the PostgreSqlRepositoryCollectionDecorator.addAttribute() (line 58). We have fixed the issues in a not so elegant way on branch: https://github.com/sidohaakma/molgenis/tree/fix/6647.

We don't want to introduce this fix because of the impact on the codebase. We rather want to cleanup the way we interact with the database. We would like to introduce loose coupling between our own domain-model (Packages, EntityTypes, Entities and Attributes) and the databasemodel (Tables and Rows).

### LINKED ISSUE #6792: When adding a new setting, using it after restart will add a comma to all inputs
### How to Reproduce
- Create a new Java setting in DataExplorerSettings.java, boolean is easy
- Restart the app
- Go to the dataexplorer
- Open the settings with the cog and set the new setting to true
- Save and close

### Expected behavior
Dataexplorer works and setting is true when you open up the cog again

### Observed behavior

screen shot 2017-10-05 at 17 09 10
screen shot 2017-10-05 at 17 07 07

### LINKED ISSUE #6846: After adding a setting to Application Settings, there's an error on save
### How to Reproduce
1. Add an attribute to ```AppDbSettings.Meta```: ```addAttribute("test");```
2. Start up the app
3. Go to the settings plugin, open the Application Settings
4. Change something and click save

### Expected behavior
Settings are saved.

### Observed behavior
```Updating read-only attribute 'ga_acc_privacy_friendly_mgs' of entity 'sys_set_app' with id 'app' is not allowed.```

When restarting the app again, the problem goes away. Different but possibly related to #6792 and #6729

### LINKED ISSUE #7254: Cannot update attributes for an abstract entity that is extended by concrete entity
### How to Reproduce
Using the RD-Connect data model:
- Change the Ontology Term abstract entity to remove IRI from the lookup attributes

### Expected behavior
The entity is correctly updated

### Observed behavior
An error message: Duplicate key AttributeDescription{name=IRI}

fdlk:
…tended by concrete entity.

The EntityType copier does a shallow copy so it will not copy the complete entity type tree, only the entity type of the repository.
Fixed by giving a bit more information to the EntityTypeRegistry, which then can figure out more easily how to update the registry.

Fixes error where the registry registered a junction table name for xrefs but none exists
Fixes error where the registry only can be called inside a transaction.
Fixes two registered bugs caused by the attribute copier which mucked up the attribute tree for compound attributes:
Fixes #6792 When adding a new setting, using it after restart will add a comma to all inputs
Fixes #6729 When adding a compound attribute to a system entity the attribute is added twice

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


fdlk:
Sonar review comments are false positives, and I cannot reproduce them in IntelliJ nor when running maven sonar:sonar locally(!)

PR #7284, opened 2018-04-06
Fix #7280 Importing multiple attributes with the same name doesn't throw an error

### LINKED ISSUE #7280: Importing multiple attributes with the same name doesn't throw an error
### How to Reproduce
Import: 
[Chr6_vragen_emx_testimport_12_everything_nillable.xlsx](https://github.com/molgenis/molgenis/files/1457455/Chr6_vragen_emx_testimport_12_everything_nillable.xlsx)
I merged here the three parts of the chromosome6 questionnaire and set everything nillable. It imported fine, so I was happy. Then we had the fix for the nillable expression and I got an error. It turned out I did merge everything, but did not delete the 2 additional id attributes. 

### Expected behavior
An error telling me to only use the name of every attribute once per entitytype

### Observed behavior
Everything imports fine and I have issues much later then one would expect.

tommydeboer:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


molgenis-platform-integration-tests/src/test/java/org/molgenis/integrationtest/platform/importservice/EmxImportServiceIT.java
```diff
@@ -463,6 +465,15 @@ public void testDoImportUpdateEmxAsSuperuser(File file, File updateFile, Map<Str
 				entityValidationMethod };
 	}
 
+	@Test(expectedExceptions = MolgenisDataException.class, expectedExceptionsMessageRegExp = "Duplicate attribute name [type] for entity type [it_emx_duplicateAttr]")
```
fdlk: escape the [], it's a regex

PR #7323, opened 2018-04-23
Speed up javascript and magmascript evaluation

fdlk:
A couple of changes:
* Cache compiled expressions
* Don't keep threadlocal bindings between requests, the global namespace will get polluted in the long run (this is a bugfix not a speedup)
    - Create new empty bindings for each simple script evaluation
    - Copy existing bindings for magmascript and bind them to the magmascript $ function in java code. 
* Reuse entity bindings for evaluation of multiple Boolean expressions on same entity, assuming that they do not pollute the context
* Create MagmaScript javascript class to scope the static functions and speed up attribute instance creation both client and server side by putting class method functions on the prototype instead of each instance. 
* Move magma-specific scripts and bindings to the JsMagmaScriptEvaluator

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
+++ b/molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
@@ -7,13 +7,13 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
-import javax.script.Bindings;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 import static java.util.Collections.singletonList;
 import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toList;
 
 /**
  * Validates entities using JavaScript expressions. Expressions can use the Magma API.
@@ -58,14 +58,14 @@ public class ExpressionValidator
 			return Collections.emptyList();
 		}
 
-		Bindings bindings = jsMagmaScriptEvaluator.createBindings(entity);
+		return jsMagmaScriptEvaluator.evalList(expressions, entity)
+									 .stream()
+									 .map(this::convertToBoolean)
+									 .collect(toList());
+	}
 
-		List<Boolean> validationResults = new ArrayList<>(expressions.size());
-		for (String expression : expressions)
-		{
-			Object value = jsMagmaScriptEvaluator.eval(bindings, expression);
-			validationResults.add(value != null ? Boolean.valueOf(value.toString()) : null);
-		}
-		return validationResults;
+	private Boolean convertToBoolean(Object value)
+	{
+		return Optional.ofNullable(value).map(Object::toString).map(Boolean::valueOf).orElse(null);
 	}
 }
--- a/molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
+++ b/molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
@@ -58,7 +58,7 @@ public class ExpressionValidator
 			return Collections.emptyList();
 		}
 
-		return jsMagmaScriptEvaluator.evalList(expressions, entity)
+		return jsMagmaScriptEvaluator.eval(expressions, entity)
 									 .stream()
 									 .map(this::convertToBoolean)
 									 .collect(toList());
--- a/molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
+++ b/molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
@@ -7,6 +7,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
+import javax.script.Bindings;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -57,10 +58,12 @@ public class ExpressionValidator
 			return Collections.emptyList();
 		}
 
+		Bindings bindings = jsMagmaScriptEvaluator.createBindings(entity);
+
 		List<Boolean> validationResults = new ArrayList<>(expressions.size());
 		for (String expression : expressions)
 		{
-			Object value = jsMagmaScriptEvaluator.eval(expression, entity);
+			Object value = jsMagmaScriptEvaluator.eval(bindings, expression);
 			validationResults.add(value != null ? Boolean.valueOf(value.toString()) : null);
 		}
 		return validationResults;
--- a/molgenis-js/src/main/java/org/molgenis/js/magma/JsMagmaScriptEvaluator.java
+++ b/molgenis-js/src/main/java/org/molgenis/js/magma/JsMagmaScriptEvaluator.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.ZoneId;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
@@ -27,6 +28,7 @@ import static java.lang.String.format;
 import static java.util.Arrays.asList;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.MICROSECONDS;
+import static java.util.stream.Collectors.toList;
 import static org.molgenis.util.ResourceUtils.getString;
 
 /**
@@ -69,6 +71,29 @@ public class JsMagmaScriptEvaluator
 		jsScriptEngine.eval(magmaBindings, string);
 	}
 
+	public List<Object> evalList(Collection<String> expressions, Entity entity)
+	{
+		return evalList(expressions, entity, ENTITY_REFERENCE_DEFAULT_FETCHING_DEPTH);
+	}
+
+	public List<Object> evalList(Collection<String> expressions, Entity entity, int depth)
+	{
+		Stopwatch stopwatch = null;
+		if (LOG.isTraceEnabled())
+		{
+			stopwatch = Stopwatch.createStarted();
+		}
+
+		Bindings bindings = createBindings(entity, depth);
+		List<Object> result = expressions.stream().map(expression -> eval(bindings, expression)).collect(toList());
+		if (stopwatch != null)
+		{
+			stopwatch.stop();
+			LOG.trace("Script evaluation took {} µs", stopwatch.elapsed(MICROSECONDS));
+		}
+		return result;
+	}
+
 	public Object eval(String expression, Entity entity)
 	{
 		return eval(expression, entity, ENTITY_REFERENCE_DEFAULT_FETCHING_DEPTH);
@@ -93,18 +118,11 @@ public class JsMagmaScriptEvaluator
 	 * @param expression  JavaScript expression to evaluate
 	 * @return evaluated expression result, return type depends on the expression.
 	 */
-	public Object eval(Bindings bindings, String expression)
+	private Object eval(Bindings bindings, String expression)
 	{
-		Stopwatch stopwatch = null;
-		if (LOG.isTraceEnabled())
-		{
-			stopwatch = Stopwatch.createStarted();
-		}
-
-		Object value;
 		try
 		{
-			value = jsScriptEngine.eval(bindings, expression);
+			return jsScriptEngine.eval(bindings, expression);
 		}
 		catch (javax.script.ScriptException t)
 		{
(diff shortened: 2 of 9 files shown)


fdlk:
Test this please

molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
```diff
@@ -57,10 +58,12 @@ boolean resolveBooleanExpression(String expression, Entity entity)
 			return Collections.emptyList();
 		}
 
+		Bindings bindings = jsMagmaScriptEvaluator.createBindings(entity);
```
dennishendriksen: This introduces a dependency on the Java scripting API. Is this desirable?

molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
```diff
@@ -57,10 +58,12 @@ boolean resolveBooleanExpression(String expression, Entity entity)
...
+
 		List<Boolean> validationResults = new ArrayList<>(expressions.size());
 		for (String expression : expressions)
 		{
-			Object value = jsMagmaScriptEvaluator.eval(expression, entity);
+			Object value = jsMagmaScriptEvaluator.eval(bindings, expression);
```
dennishendriksen: According to the task for which this code was written evalList should be used instead of eval since this was measured to be a major performance bottleneck. Could you elaborate why this change wasn't made?

molgenis-js/src/main/java/org/molgenis/js/magma/JsMagmaScriptEvaluator.java
```diff
@@ -79,6 +124,38 @@ public Object eval(String expression, Entity entity, int depth)
...
+	}
+
+	/**
+	 * Creates magmascript bindings for a given Entity.
+	 *
+	 * @param entity the entity to bind to the magmascript $ function
+	 * @param depth  maximum depth to follow references when creating the entity value map
+	 * @return Bindings with $ function bound to the entity
+	 */
+	public Bindings createBindings(Entity entity, int depth)
```
dennishendriksen: According to IntelliJ this method can be private. I expected this to be called by ```AlgorithmServiceImpl```. Is the algorithm service benefiting from the performance improvements in this PR?

molgenis-js/src/main/java/org/molgenis/js/magma/JsMagmaScriptEvaluator.java
```diff
@@ -79,6 +124,38 @@ public Object eval(String expression, Entity entity, int depth)
...
+	{
+		Bindings bindings = new SimpleBindings();
```
dennishendriksen: Does this conform with:

> You can share a ScriptEngine and CompiledScript objects across threads. They are threadsafe. Actually, you should share them, as a single engine instance is a holder for a class cache and for JavaScript objects' hidden classes, so by having only one you cut down on repeated compilation.

> What you can't share is Bindings objects. The bindings object basically corresponds to the JavaScript runtime environment's Global object. The engine starts with a default bindings instance, but if you use it in multithreaded environment, you need to use engine.createBindings() to obtain a separate Bindings object for every thread -- its own global, and evaluate the compiled scripts into it. That way you'll set up isolated global scopes with the same code.

It seems like bindings content is shared between script executions?

molgenis-js/src/main/java/org/molgenis/js/nashorn/NashornScriptEngine.java
```diff
@@ -1,107 +1,68 @@
...
+		Object returnValue = compiledExpression.eval(bindings);
+		return convertNashornValue(returnValue);
 	}
 
-	public Object eval(String script) throws ScriptException
+	public ScriptObjectMirror newJSArray()
 	{
-		Bindings bindings = bindingsThreadLocal.get();
-		return scriptEngine.eval(script, bindings);
+		return (ScriptObjectMirror) ((JSObject) scriptEngine.get("Array")).newObject();
```
dennishendriksen: How about changing this to return List<Object> as to remove part of the magic from JsMagmaScriptEvaluator to this class?

molgenis-js/src/main/java/org/molgenis/js/nashorn/NashornScriptEngine.java
```diff
@@ -132,23 +92,15 @@ private Object convertNashornValue(Object nashornValue)
...
-				else if (((ScriptObjectMirror) nashornValue).containsKey(idValueKey))
-				{
-					// entity object returned from script
-					return ((ScriptObjectMirror) nashornValue).get(idValueKey);
-				}
-				else
-				{
-					throw new RuntimeException("Unable to convert [ScriptObjectMirror]");
-				}
+				else return ((ScriptObjectMirror) nashornValue).getOrDefault(KEY_ID_VALUE, null);
```
dennishendriksen: Why change throwing an exception to return ```null```?

molgenis-js/src/main/resources/js/script-evaluator.js
```diff
@@ -1,452 +1,444 @@
...
- * @param entity
- *            the entity
+ * @param script the script
+ * @param entity object mapping attribute name key to attribute value
  * @returns the evaluated script result
  */
 function evalScript (script, entity) {
-  function Attribute (value) {
+    // make functions available to the script in local scope
+    var $ = MagmaScript.$.bind(entity)
```
dennishendriksen: IntelliJ reports various warnings in the function related. Could you fix or suppress them?

molgenis-js/src/main/resources/js/script-evaluator.js
```diff
@@ -1,452 +1,444 @@
 /**
```
dennishendriksen: Do we have automated tests for all functions? Do these changes change the interface in the Java <-> Java or JavaScript <-> JavaScript scenarios? It is a bit hard to see what changed due to formatting changes.

molgenis-js/src/test/java/org/molgenis/js/JsMagmaScriptEvaluatorTest.java
```diff
@@ -155,7 +161,7 @@ public void testValueForNestedXrefDefaultDepth()
 
 		Object scriptExceptionObj = jsMagmaScriptEvaluator.eval("$('gender.xref.label').value()", person);
 		assertEquals(scriptExceptionObj.toString(),
-				"org.molgenis.script.core.ScriptException: <eval>:502 TypeError: Cannot read property \"label\" from undefined");
+				"javax.script.ScriptException: TypeError: Cannot read property \"label\" from undefined in <eval> at line number 431");
```
dennishendriksen: Why expose internal ScriptException instead of our ScriptException that doesn't depend on Java scripting API?

molgenis-js/src/test/java/org/molgenis/js/JsMagmaScriptEvaluatorTest.java
```diff
@@ -505,8 +511,25 @@ public void age()
 		Entity person = new DynamicEntity(personBirthDateMeta);
 		person.set("birthdate", now().atOffset(UTC).toLocalDate());
 
-		Object result = jsMagmaScriptEvaluator.eval("$('birthdate').age().value()", person, 3);
+		Bindings bindings = jsMagmaScriptEvaluator.createBindings(person);
```
dennishendriksen: Is suggest moving the performance test to a separate disabled test case that can be executed manually and actually fail when it doesn't complete on time.

molgenis-js/src/main/java/org/molgenis/js/magma/JsMagmaScriptEvaluator.java
fdlk: The bindings shouldn't be shared between threads. They are supposed to be shortlived and used only for a single set of script executions on the entity.

molgenis-js/src/test/java/org/molgenis/js/JsMagmaScriptEvaluatorTest.java
fdlk: agreed

molgenis-js/src/main/java/org/molgenis/js/magma/JsMagmaScriptEvaluator.java
fdlk: There are two methods in the mapping service.

### algorithm editing
While the user is typing on a single algorithm it regularly applies that single algorithm to a couple of entity rows from the appropriate source repository to get a column of target attribute values.

So it runs a single loop over all source repository entities. Reusing bindings is pointless here but the caching of the compiled algorithm already produces a noticeable speedup.

### running a mapping project
The mapping service then runs a triple loop.
* For each source repository
  * For each source repository entity
    * For each target attribute
      * it makes a call to the algorithm service

This also already benefits from the caching of the compiled algorithm but we can still significantly speed up the performance of running a mapping project if we delegate the inner loops to the algorithm service. That'll allow the algorithm service to cache the bindings between computations of the target attribute values for a single source row. 

I already took a brief look but the changes were nontrivial and don't benefit the Questionnaire use case. So I'd rather make that a separate issue and schedule it as a (probably high priority) bugfix.

molgenis-js/src/main/java/org/molgenis/js/nashorn/NashornScriptEngine.java
fdlk: The ScriptObjectMirror mirrors a native javascript array. Java-side you can convert it to list to manipulate it and fill it with other objects, but when passing it back to nashorn you need to provide this ScriptObjectMirror so that it ends up as its original self and for instance Arrays.isArray evaluates to true on the javascript-side.

molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
fdlk: I've switched to evalList method to hide the details.

molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
fdlk: I've switched to evalList method to hide the details.

molgenis-js/src/main/java/org/molgenis/js/magma/JsMagmaScriptEvaluator.java
fdlk: I've switched to evalList method to hide the details, so they won't be shared for sure now.

molgenis-js/src/main/java/org/molgenis/js/nashorn/NashornScriptEngine.java
fdlk: It happens at initialisation time when the libraries are loaded. They return the ScriptObjectMirror for a javascript function. So the choice here was to return the ScriptObjectMirror or to toss the mirror away.

molgenis-js/src/test/java/org/molgenis/js/JsMagmaScriptEvaluatorTest.java
fdlk: It was getting doubly wrappered and I didn't want to add a third wrapper. I've unwrapped it instead now.

fdlk:
Took a brief look at the questionnaire init performance but there I think the time is being taken up by sql queries and metadata lookups, I don't see much scripting activity going on. Do notice that it's creating a TON of transactions and if I add a @Transactional annotation to the retrieveEntityCollection method of RestControllerV2 execution time gets reduced significantly.

molgenis-data-validation/src/main/java/org/molgenis/data/validation/ExpressionValidator.java
dennishendriksen: Could you add unit test(s) for ExpressionValidator.resolveBooleanExpression?

molgenis-js/src/main/java/org/molgenis/js/magma/JsMagmaScriptEvaluator.java
dennishendriksen: Thanks for the explanation. Could you create a story in the backlog for this enhancement?

dennishendriksen: I've noticed that we're missing unit tests for script engine value conversion for attribute types BOOL, DATE_TIME, LONG. Lets create them and add in a separate PR.

fdlk: Created http://wiki.gcc.rug.nl/ticket/5830#ticket

PR #7324, opened 2018-04-24
M5415 App manager

mark-de-haan:
Documentation is WIP, opening PR for sonar information and travis build

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-app-manager/src/main/java/org/molgenis/app/manager/service/AppDeployService.java
```diff
@@ -0,0 +1,43 @@
...
+	void loadJavascriptResources(String uri, String fileName, HttpServletResponse response) throws IOException;
+
+	/**
+	 * Load css files from the file store
+	 *
+	 * @param uri      An uri belonging to an app
+	 * @param fileName File name of the requested css file
+	 * @param response a {@link HttpServletResponse} used to attach any found css files
+	 */
+	void loadCSSResources(String uri, String fileName, HttpServletResponse response) throws IOException;
```
fdlk: To keep this service independent of http servlet spec I'd let this method yield an InputStream and then let the controller copy it to the response. Also makes for easier testing.

molgenis-app-manager/src/main/java/org/molgenis/app/manager/service/impl/AppDeployServiceImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+	public AppDeployServiceImpl(DataService dataService)
+	{
+		this.dataService = requireNonNull(dataService);
+	}
+
+	@Override
+	public String configureTemplateResourceReferencing(String template, String baseUrl)
+	{
+		template = template.replaceAll("src=js/", "src=" + baseUrl + "js/");
```
fdlk: but what if they DO add "" in their html?

molgenis-app-manager/src/main/java/org/molgenis/app/manager/service/impl/AppDeployServiceImpl.java
```diff
@@ -0,0 +1,107 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+
+import static java.net.URLConnection.guessContentTypeFromName;
+import static java.util.Objects.requireNonNull;
+
+@Service
+public class AppDeployServiceImpl implements AppDeployService
+{
+	private static final String JS_FOLDER = "js";
```
fdlk: Determine mime type based on file extension (just like ordinary spring resources are served from classpath). 
If you pick app serve URI in a way that doesn't clash with the meta methods you can serve all folders.

molgenis-app-manager/src/main/frontend/README.md
```diff
@@ -0,0 +1,30 @@
+# molgenis-app-manager-ui
+
+> UI code for the app-manager plugin
+
+## Build Setup
+
+``` bash
+# install dependencies
+npm install
```
ConnorStroomberg: we recently updated this to yarn , it would be good if we update this

PR #7340, opened 2018-05-01
Suggestion how to fix the Questionnaire init performance

fdlk:
Needs a bit more work I suspect but a suggestion how to fix.

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/AttributeResponseV2.java
+++ b/molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/AttributeResponseV2.java
@@ -1,13 +1,12 @@
 package org.molgenis.data.rest.v2;
 
 import org.molgenis.core.ui.data.support.Href;
-import org.molgenis.data.DataService;
-import org.molgenis.data.Fetch;
-import org.molgenis.data.Range;
+import org.molgenis.data.*;
 import org.molgenis.data.meta.AttributeType;
 import org.molgenis.data.meta.model.Attribute;
 import org.molgenis.data.meta.model.EntityType;
 import org.molgenis.data.support.EntityTypeUtils;
+import org.molgenis.data.support.QueryImpl;
 import org.molgenis.security.core.UserPermissionEvaluator;
 
 import java.util.List;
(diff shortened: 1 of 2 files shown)


molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/AttributeResponseV2.java
```diff
@@ -74,7 +73,10 @@ public AttributeResponseV2(final String entityParentName, EntityType entityType,
 			if (includeCategories && (this.fieldType.equals(AttributeType.CATEGORICAL) || this.fieldType.equals(
 					AttributeType.CATEGORICAL_MREF)))
 			{
-				this.categoricalOptions = dataService.findAll(refEntity.getId())
+				String refEntityIdAttributeName = refEntity.getIdAttribute().getName();
+				Query<Entity> refEntityQuery = new QueryImpl<>().pageSize(1000)
```
dennishendriksen: What happens in case of more than a thousand categories?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/RestControllerV2.java
```diff
@@ -234,6 +234,7 @@ public void deleteEntityCollection(@PathVariable("entityTypeId") String entityTy
 	 * Retrieve an entity collection, optionally specify which attributes to include in the response.
 	 */
 	@GetMapping("/{entityTypeId}")
+	@Transactional(readOnly = true)
```
dennishendriksen: How about marking the DataServiceImpl.findAll with ```@Transactional(readOnly=true, propagation = MANDATORY)``` to enforce this behavior everywhere?

molgenis-data-rest/src/main/java/org/molgenis/data/rest/v2/AttributeResponseV2.java
ConnorStroomberg: We could just add this to the api specs, and even set the pageSize to something like 100. > 1000 results is not really a sensible usecase fore the 'includerefs' flag.

ConnorStroomberg:
@fdlk i propose we close this ? as we currently have no indication that users are having problems related to performance.

PR #7349, opened 2018-05-07
Suppress findbugs null argument warning on active menu item

bartcharbon:
Checked: where used the results are null-checked.
Code is untouched, I think the warning surfaces now due to the guava version bump

#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-core-ui/src/main/java/org/molgenis/core/ui/security/MolgenisUiMenuPermissionDecorator.java
+++ b/molgenis-core-ui/src/main/java/org/molgenis/core/ui/security/MolgenisUiMenuPermissionDecorator.java
@@ -86,6 +86,7 @@ public class MolgenisUiMenuPermissionDecorator implements UiMenu
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


molgenis-core-ui/src/main/java/org/molgenis/core/ui/security/MolgenisUiMenuPermissionDecorator.java
```diff
@@ -86,8 +86,10 @@ public boolean containsItem(final String itemId)
 	}
 
 	@Override
+	@javax.annotation.CheckForNull
```
dennishendriksen: What is the difference between javax.annotation.CheckForNull (which isn't used in the codebase)  and javax.annotation.Nullable (which is used in the codebase)?

dennishendriksen: https://stackoverflow.com/a/12301937/8579801

PR #7453, opened 2018-06-15
Fix Sonar Warnings

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/VcfRepository.java
```diff
@@ -34,98 +33,118 @@
...
-	 * behaviour. If you want to get the EntityType of this repository and you can't access getEntityType(),
-	 * convert the iterator to a PeekingIterator and peek the first Entity.
-	 */
 	@Override
 	public Iterator<Entity> iterator()
 	{
-		Iterator<VcfRecord> vcfRecordIterator = Iterators.unmodifiableIterator(vcfReaderFactory.get().iterator());
-		VcfToEntity vcfToEntity = vcfToEntitySupplier.get();
-		return Iterators.transform(vcfRecordIterator, vcfToEntity::toEntity);
+		throw new UnsupportedOperationException("Use forEachBatched instead of iterator");
```
dennishendriksen: This violates the Repository contract. How about storing a Closable locally for each call to iterator() and closing the Closables on calling Repository.close?

molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/VcfRepository.java
```diff
@@ -34,98 +33,118 @@
...
+		return Collections.emptySet();
 	}
 
 	@Override
-	public Set<RepositoryCapability> getCapabilities()
+	public long count()
 	{
-		return Collections.emptySet();
+		AtomicInteger counter = new AtomicInteger(0);
+		forEachBatched(batch -> counter.addAndGet(batch.size()), 1000);
```
dennishendriksen: Store magic number in private static final?

molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/importer/VcfImporterService.java
```diff
@@ -185,11 +183,12 @@ private EntityImportReport importVcf(Repository<Entity> inRepository, List<Entit
...
 			AtomicInteger vcfEntityCount = new AtomicInteger();
-			outRepository.add(StreamSupport.stream(inRepository.spliterator(), false).filter(entity ->
+			inRepository.forEachBatched(rowBatch ->
 			{
-				vcfEntityCount.incrementAndGet();
-				return true;
-			}));
+				outRepository.add(rowBatch.stream());
+				vcfEntityCount.addAndGet(rowBatch.size());
+			}, 1000);
```
dennishendriksen: Store magic number in private static final?

molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/VcfRepository.java
fdlk: I (We?)'d like to get rid of the calls to iterator() altogether and for VcfRepository we checked that none are left after this PR. The whole VcfReaderRegistry is a bug waiting to happen so I'd say good riddance to it. Would introducing a RepositoryCapability.ITERABLE help keep the contract safe? How much work till we can get rid of the method altogether?

molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/VcfRepository.java
dennishendriksen: @tommydeboer What do you think?

molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/importer/VcfImporterService.java
dennishendriksen: @tommydeboer what do you think?

molgenis-data-vcf/src/main/java/org/molgenis/data/vcf/VcfRepository.java
tommydeboer: There's [a WIP here](https://github.com/dennishendriksen/molgenis/tree/refactor/removeRepositoryIterator) to remove the Iterable contract from Repositories altogether. It's based on this PR. We can't be sure that we'll have time to finish it in the near future. 

We explored two options concerning this PR:
1. Merge as is. The ```iterator()``` will throw an ```UnsupportedOperationException()``` and we'll just have to make sure that nobody uses it. 
2. Suppress the Sonar warning for now and merge this fix later as part of the branch mentioned above. 

After some discussion we decided to go for option 1.

PR #7467, opened 2018-06-21
Fix #7459 RLS decorator silently ignores errors during stream update

### LINKED ISSUE #7291: EntityTypeRepositorySecurityDecorator filter streams for update instead of throwing exception
### How to Reproduce
- update a stream of entityTypes via the dataservice with a regular user
- make sure that this user has permission on some of these entities and no permissions on others

### Expected behavior
an exception is throws when an entity without permissions is encountered, just like what happens when you try to add/update one entity without the proper permissions.

### Observed behavior
operations succeeds for the entities with permission, others are silently ignored

NOTE: Filtering is a good way to use for counts and finds, but should not be used for add update and delete.

### LINKED ISSUE #7459: Streamed updates on RLS entity types silently ignore errors when some entities are inaccessible
Following the discussion in #7291:

### How to Reproduce
- Import it_emx_update
- Make TestUpdate row level secured
- Add user
- Give user permissions:
  - sys package (READ)
  - it_emx_update_TestUpdate (WRITE)
- Add token for user: "test"
- Login as user
- Add a row to TestUpdate with id: 2

Execute the following REST calls:
```
PUT localhost:8080/api/v2/it_emx_update_TestUpdate
{entities: [
{
    "id": "1",
    "label": "new label"
}
]}
```

```
PUT localhost:8080/api/v2/it_emx_update_TestUpdate
{entities: [
{
    "id": "1",
    "label": "newer label"
},
{
    "id": "2",
    "label": "newer label"
}
]}
```

```
PUT localhost:8080/api/v2/it_emx_update_TestUpdate
{entities: [
{
    "id": "1",
    "label": "newer label"
},
{
    "id": "99",
    "label": "newer label"
}
]}
```

### Expected behavior
Like with a non row level secured entity, the following error should be thrown:
```Cannot update [it_emx_update_TestUpdate] with id [x] because it does not exist```

### Observed behavior
In all cases a ```200 OK``` is returned. Inaccessible entities (id=1) aren't updated, but in one case the user's entity (id=2) **is** updated. That shouldn't happen.

tommydeboer:
Probably also fixes #7291 EntityTypeRepositorySecurityDecorator filter streams for add and update instead of throwing exception

#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-data-security/src/main/java/org/molgenis/data/security/owned/AbstractRowLevelSecurityRepositoryDecorator.java
+++ b/molgenis-data-security/src/main/java/org/molgenis/data/security/owned/AbstractRowLevelSecurityRepositoryDecorator.java
@@ -45,16 +45,14 @@ public abstract class AbstractRowLevelSecurityRepositoryDecorator<E extends Enti
 	public Iterator<E> iterator()
 	{
 		Iterable<E> iterable = () -> delegate().iterator();
-		return stream(iterable.spliterator(), false).filter(entity -> isActionPermitted(entity, READ))
-													.iterator();
+		return stream(iterable.spliterator(), false).filter(entity -> isActionPermitted(entity, READ)).iterator();
 	}
 
 	@Override
 	public void forEachBatched(Fetch fetch, Consumer<List<E>> consumer, int batchSize)
 	{
 		delegate().forEachBatched(fetch, entities -> consumer.accept(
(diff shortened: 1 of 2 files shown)


fdlk:
test this please

tommydeboer:
test this please

Issue #7429, opened 2018-06-08
App uri cannot end with /


fdlk:
### How to Reproduce
Upload an app with uri ending in `/`
[Archive.zip](https://github.com/molgenis/molgenis/files/2083544/Archive.zip)
Add it to the menu, give permissions to a user, open the app.

### Expected behavior
App opens or upload fails

### Observed behavior
> App with uri example1 does not exist. (AM06)

PR #7484, opened 2018-06-26
Fix #7432 #7429 App manager fixes

### LINKED ISSUE #7432: Putting apps in the menu relies on plugin ID hack
### How to Reproduce
Upload app, add it to the menu, open data explorer, select Plugin

### Expected behavior
Plugin ID does not contain /

### Observed behavior
Plugin ID equals `app/uri/`, where uri is the actual app id.
This is not okay, because the REST api cannot handle identifiers containing /

tommydeboer:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


bartcharbon:
- Exception AM07 still has a hardcoded "/app/" in the message.
- App metadata states " I.e. /app/{uri}" in the description for the URI attr
- Not sure if URI is still the right attr name
- adding a slash anywhere in the app name results in a broken app
#7429: App uri cannot end with / -> doesn't give a proper exception anymore. We might want to fix this issue in this PR.

tommydeboer:
> Exception AM07 still has a hardcoded "/app/" in the message.

That's actually correct because the menu forwards you to the actual location `/plugin/app/your-plugin`

>App metadata states " I.e. /app/{uri}" in the description for the URI attr
Not sure if URI is still the right attr name

I replaced the `uri` attribute with a `name` attribute.

PR #7511, opened 2018-07-04
Fix #7502: When I reload questionnaire page, the answers are gone

### LINKED ISSUE #7502: When I reload questionnaire page, the answers are gone
### How to Reproduce
Open questionnaire, fill in some answers, and reload the page or switch language.

### Expected behavior
Answers appear on reloaded page

### Observed behavior
Answers are gone

fdlk:
#### Checklist
- [ ] Code reviewed
- [x] Clean commits


molgenis-questionnaires/src/main/frontend/src/store/getters.js
```diff
@@ -49,12 +49,7 @@ const isVisible = (field: ChapterField, formData: Object) => {
   try {
     visible = field.visible(formData)
   } catch (e) {
-    console.error(`Error in getters, during evaluation of function 'visible' for field ${field.id}.
-      Where visible expression is: 
-        '${field.visible.toString()}'
-      And formData is: 
-        '${JSON.stringify(formData)}'
-      The default value of visible is set to false`, e)
+    console.log(`Setting ${field.id}.visible to false.`)
```
dennishendriksen: @sidohaakma is wondering why this is a console.log instead of a console.error

fdlk: When you start up the application with the vuex debugger open, you get about fifty of these messages which is overly disturbing. Therefore I toned down the red screaming messages a bit.

dennishendriksen: ok, but the question remains: why log instead of error?

fdlk: Because it is not always an error. Talked with @sidohaakma and we'll turn it into a log.warn.

tommydeboer:
There are other issues with the TestQuestionnaire! Some attributes don't show up because they aren't in a compound.

PR #7523, opened 2018-07-09
Feature/group permissions

fdlk:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-security/src/main/java/org/molgenis/data/security/auth/GroupService.java
```diff
@@ -86,6 +102,126 @@ public void persist(GroupValue groupValue)
 		dataService.add(ROLE, roles.stream());
 	}
 
+	/**
+	 * Grants default permissions on the root package and group to the roles of the group
+	 *
+	 * @param groupValue details of the group for which the permissions will be granted
+	 */
+	public void grantPermissions(GroupValue groupValue)
```
ConnorStroomberg: 'grantDefaultPermissions' seems like a more descriptive name. I would suggest moving the create actions from the controller to the service , this can then be part of the service method and be kept private ( inner workings of creating a group)

molgenis-data-security/src/main/java/org/molgenis/data/security/auth/GroupService.java
```diff
@@ -86,6 +102,126 @@ public void persist(GroupValue groupValue)
...
+	@RunAsSystem
+	public void removeMember(final Group group, final User user)
+	{
+		ArrayList<Role> groupRoles = newArrayList(group.getRoles());
+		final RoleMembership member = roleMembershipService.getMemberships(groupRoles)
+														   .stream()
+														   .filter(m -> m.getUser().getId().equals(user.getId()))
+														   .findFirst()
+														   .orElseThrow(() -> new UnknownEntityException(
+																   roleMembershipMetadata,
```
ConnorStroomberg: i would argue this was clearer before, just because every thing can be chained does not mean is should.

fdlk: https://dzone.com/articles/java-8-optional-replace-your-get-calls is what I was thinking about

molgenis-data-security/src/main/java/org/molgenis/data/security/auth/GroupService.java
fdlk: Method used to sit at the bottom of the class.
This PR moves it to the top, and makes it grant two permission sets, on both group and package, instead of just on the package.

fdlk: Would actually be nice if we can get rid of it here altogether and let another class worry about the permissions on the group? Can also get rid of the PermissionService as a constructor parameter here then.

molgenis-data-security/src/main/java/org/molgenis/data/security/auth/GroupService.java
ConnorStroomberg: yes, the article does not suggest to chain the everything, it closes with a comment about the needed functionality not being available until java 9, 

```Unfortunately this is yet to come It will be available in Java 9.

Until then we have to write something like:

if(optional.isPresent()){
    doSomething(optional.get());
} else {
    doSomethingElse();
}
There are cases in which you are allowed to use get() and isPresent() but use them with a grain of salt.```

ConnorStroomberg:
in the current state i am not able to create groups anymore :( , 

`Unable to find ACL information for object identity 'org.molgenis.data.security.GroupIdentity[Type: group; Identifier: foo]`

molgenis-data-security/src/main/java/org/molgenis/data/security/auth/GroupService.java
fdlk: I've split off the creation of the exception and reused the existing findRoleMembership method.

Issue #6731, opened 2017-09-27
Count permissions on System EntityTypes broken


fdlk:
### How to Reproduce
Give non-sysadmin user the necessary permissions to use the Data Explorer
Give the user COUNT permission on sys_sec_User EntityType
Give the user COUNT permissions on TypeTestRef EntityType
Log in as the user and open the Data Explorer.

### Expected behavior
I can select both User and TypeTestRef in the entity select box
When I select either User or TypeTestRef, the aggregates tab is shown

### Observed behavior
TypeTestRef works as expected but I cannot select User in the entity select box.
When I try to manually select it by surfing to `https://molgenis62.gcc.rug.nl/menu/main/dataexplorer?entity=sys_md_User` , I get redirected to the login dialog.

PR #7533, opened 2018-07-11
[FIX] Grant COUNT on sys package to USER role and READ on Security plugin to MANAGER role.

### LINKED ISSUE #7528: As non-su I cannot import data
### How to Reproduce
- create a group "test"
- create a user "tester"

### Expected behavior

### Observed behavior

fdlk:
Fixes #7528

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


dennishendriksen:
Regression of #6731 (see failing integration tests)? Permitting all users to count all system metadata opens unnecessary doors? Did you consider fixing #7528 by executing the mentioned check using elevated permissions?

PR #7537, opened 2018-07-12
Fix #7528: As non-su I cannot import data

### LINKED ISSUE #7528: As non-su I cannot import data

bartcharbon:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


fdlk:
I've created http://www.molgenis.org/ticket/5986#ticket to add a new permission set with READ_META only.

PR #7560, opened 2018-07-17
Dataexplorer: delete button for user with permission to delete instead of only for superusers

bartcharbon:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
+++ b/molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
@@ -326,6 +326,14 @@ $.when($,
                 }
             });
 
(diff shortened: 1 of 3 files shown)


molgenis-dataexplorer/src/main/resources/js/dataexplorer.js
```diff
@@ -326,6 +326,14 @@ $.when($,
                 }
             });
 
+            $.get(molgenis.getContextUrl() + '/deletePermission?entity=' + state.entity).done(function (data) {
```
fdlk: The data explorer already does a call to /meta which contains this information

PR #7569, opened 2018-07-18
Revert "Fix #7528: As non-su I cannot import data"

### LINKED ISSUE #7528: As non-su I cannot import data

fdlk:
This reverts commit e1ec52d55dc2c3d43497609961d95f1b254bb7fe.
We already give the read meta permission to all users instead.

#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-data/src/main/java/org/molgenis/data/util/MetaUtils.java
+++ b/molgenis-data/src/main/java/org/molgenis/data/util/MetaUtils.java
@@ -7,7 +7,6 @@ import javax.annotation.Nullable;
 
 import static org.molgenis.data.meta.model.EntityTypeMetadata.*;
 import static org.molgenis.data.system.model.RootSystemPackage.PACKAGE_SYSTEM;
-import static org.molgenis.security.core.runas.RunAsSystemAspect.runAsSystem;
 
 public class MetaUtils
 {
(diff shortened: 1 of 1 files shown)


molgenis-data/src/main/java/org/molgenis/data/util/MetaUtils.java
```diff
@@ -42,10 +41,7 @@ public static boolean isSystemPackage(@Nullable Package aPackage)
 		{
 			return false;
 		}
-		//Needs to run as system since you might not have permission to read the sys package
-		//Once we have READ_META permission checks on the packages this runAsSystem can be removed
-		//https://github.com/molgenis/molgenis/issues/7528
-		return runAsSystem(() -> aPackage.getId().equals(PACKAGE_SYSTEM) || (aPackage.getRootPackage() != null
-				&& aPackage.getRootPackage().getId().equals(PACKAGE_SYSTEM)));
+		return aPackage.getId().equals(PACKAGE_SYSTEM) || (aPackage.getRootPackage() != null
```
sidohaakma: Shouldn't this branch be tested in a JUnit test? There is also one for the EntityUtils class, but if you (as in the platform-team) have decided otherwise I could imagine this is not necessary.

sidohaakma:
New Jenkins, works like a charm at this moment ;-D.

molgenis-data/src/main/java/org/molgenis/data/util/MetaUtils.java
fdlk: I’m juist reverting the RunAsSystem.

PR #7597, opened 2018-07-20
Make SORTA a superuser plugin by default

tommydeboer:
Documentation updated for this PR and #7593 

I also looked at the testplans:
 - Questionnaires (now uses the Editor role)
 - One-Click-Importer (already used the Manager role, no change needed) 
 - SORTA (wasn't updated to new group security yet. no change needed)

#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/docs/user_documentation/admin-features/guide-ref-security.md
+++ b/docs/user_documentation/admin-features/guide-ref-security.md
@@ -67,25 +67,17 @@ The default permissions for these roles are as follows:
     * One click importer
   * Navigator
   * Data Explorer
-  * Data Integration
-    * Mapping service
-    * SORTA
   * Plugins
     * Search all
     * Job overview
-    * Questionnaires
     
 **Editor**:
 * You can edit data in the group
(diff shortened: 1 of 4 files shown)


dennishendriksen:
Hoe about adding a `@PreAuthorize("hasAnyRole('ROLE_SU')")` on all mapped controller methods to enforce superuser-only usage?

dennishendriksen:
@tommydeboer any clue why the coverage report looks like it does?

tommydeboer:
@dennishendriksen 
> How about adding a @PreAuthorize("hasAnyRole('ROLE_SU')") on all mapped controller methods to enforce superuser-only usage?

That might be a bit too strict I think. At least now you can still configure it so that normal users can use SORTA, if you really want to.

tommydeboer:
@dennishendriksen 
> any clue why the coverage report looks like it does?

No. The commit history is a bit messed up in this PR so I'll try cleaning that up first.

PR #7770, opened 2018-09-07
Fix #6928 Attribute.idAttribute update shouldn't be allowed

### LINKED ISSUE #6928: Attribute.idAttribute update shouldn't be allowed
### How to Reproduce
- In the metadata manager, create a new EntityType
- Add an attribute name = id_1, label = First ID
- set First ID as ID attribute and as Label attribute
- Save
<br></br>
- Add an attribute name = id_2, label = Second ID
- Set Second ID as ID attribute and as Label attribute
- Save
<br></br>
- Set First ID as nullable
- Save

### Expected behavior
Save is successful because First ID is no longer the ID attribute

### Observed behavior
`; nested exception is org.postgresql.util.PSQLException: ERROR: column "id_1" is in a primary key`

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


bartcharbon:
Shouldn't the dropdown be disabled if the update is not allowed? 
Like it is now for entities that inherit the ID from a parent (e.g. the "Amazon Bucket file ingest job execution")

PR #7825, opened 2018-09-25
Fix #7814 Metadata manager reference to no-reference type update

### LINKED ISSUE #7814: Metadata manager attribute reference to no-reference type update
### How to Reproduce
- load typetest
- change "xref" attribute to String
- save changes

### Expected behavior
- success

### Observed behavior
- error because the refEntity field for the attribute is not empty (and should be for a string), however the "refEntity" field gets hidden for strings, so I have to empty the refEnity field before changing the datatype

### Desired behaviour (I think)
The refEntity field should be cleared "onSave". 
Clearing it "onChange" would be annoying when you change datatype by accident.

dennishendriksen:
#### Checklist
- [x] Code reviewed
- (n.a.) User documentation updated
- [x] Clean commits

### DIFF
--- a/molgenis-metadata-manager/src/main/frontend/src/store/mutations.js
+++ b/molgenis-metadata-manager/src/main/frontend/src/store/mutations.js
@@ -67,6 +67,24 @@ const filterNonVisiblePackages = (packages: Array<EditorPackageIdentifier>) => {
 
 const compareByLabel = (a, b) => a.label && b.label ? a.label.localeCompare(b.label) : 0
 
+const isReferenceType = (type: string) => {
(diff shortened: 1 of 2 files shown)


molgenis-metadata-manager/src/main/frontend/src/store/mutations.js
```diff
@@ -144,6 +165,7 @@ export default {
       }
     }
     state.editorEntityType.attributes[index][key] = update.value
+    console.log(state.editorEntityType.attributes[index])
```
tommydeboer: Forgot to remove?

PR #7867, opened 2018-10-01
[FEAT] Add metrics for prometheus

fdlk:
#### Checklist
- [x] Code reviewed
- Test plan template updated
- [x] Clean commits


fdlk:
I think the test fails because another test case is being naughty and leaves an authentication on the test thread when done.

molgenis-bootstrap/src/main/java/org/molgenis/bootstrap/populate/UsersRolesPopulatorImpl.java
```diff
@@ -114,6 +115,11 @@ public void populate() {
         "Role containing permissions needed to manage groups. This role is included in all group manager roles.");
     manager.setIncludes(ImmutableList.of(editor));
 
+    Role metrics = roleFactory.create();
```
sidohaakma: Can you add a migration step for this role?

PR #7863, opened 2018-10-01
Fix #7862 Ontology-deletion non-practical

dennishendriksen:
#### Checklist
- [x] Functionality works & meets specifications
- [x] Code reviewed
- [x] Code unit/integration/system tested
- [NA] User documentation updated
- [NA] (If you have changed REST API interface) view-swagger.ftl updated
- [NA] Test plan template updated
- [x] Clean commits
- [x] Added Feature/Fix to release notes

PR #7871, opened 2018-10-03
Change schema update check and messages

dennishendriksen:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-migrate/src/main/java/org/molgenis/data/migrate/version/MolgenisUpgradeServiceImpl.java
```diff
@@ -38,13 +38,13 @@ public void addUpgrade(MolgenisUpgrade upgrade) {
   @Override
   public boolean upgrade() {
     int schemaVersion = versionService.getSchemaVersion();
-    if (schemaVersion < 19) {
+    if (schemaVersion < 32) {
```
tommydeboer: In the new migration step the version went from 31 to 33, see [this diff](https://github.com/molgenis/molgenis/pull/7863/files#diff-30655e6016df18b711aa540a26d0dbf8L20). So shouldn't this read `schemaVersion < 31`?

Now that I think of it... We've been at version 31 for a very very long time (since molgenis 2.0). But those versions aren't migratable. So actually this check should go something like `schemaVersion < 31 && molgenisVersion < 7.0`

dennishendriksen: MOLGENIS 7.2 has schema version 31, so indeed this should be ```<31```. As discussed in person we won't include the version check here.

PR #7919, opened 2018-10-29
Fix #7915 When I fill in an out of range number in questionnaire, I can proceed and submit

### LINKED ISSUE #7915: When I fill in an out of range number in questionnaire, I can proceed and submit (!)
### How to Reproduce
Upload 
[test.xlsx](https://github.com/molgenis/molgenis/files/2515058/test.xlsx)
Make Row level secured
Go to questionnaires
Fill in:
medium
medium
9

Validation fails for 9 (because it should be within range 0-6)
Click on next
And submit

### Expected behavior
Next doesn't work because validation fails and I cannot submit

### Observed behavior
I can click next and submit the invalid value

Note: probably something with combination nillable and rangeMin/rangeMax on int

ConnorStroomberg:
When I fill in an out of range number in questionnaire, I can proceed and submit

Use equal-or-larger instead of function definition

#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-questionnaires/src/main/frontend/src/store/getters.js
+++ b/molgenis-questionnaires/src/main/frontend/src/store/getters.js
@@ -36,7 +36,7 @@ const isChapterComplete = (chapter: Object, formData: Object): boolean => {
     const valid = child.validate(formData)
 
     if (filledInValue) {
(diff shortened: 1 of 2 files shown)


sidohaakma:
this this please!

PR #7943, opened 2018-11-12
[Fix] Make request body optional when updating entity.attribute via rest api

### LINKED ISSUE #7942: Update single attribute put/post is rejected if body is empty or null
### How to Reproduce
Send a post to the rest api endpoint 
http://[server]/api/v1/[entity-id]/[attribute-id]
method: PUT or Post
Content-Type: application/json'
With empty request body 

### Expected behavior
The attribute of the referenced entity is cleared / set to null.

### Observed behavior
The request is rejected with the message that the request is not valid, due to missing request body.

ConnorStroomberg:
This allows clearing the attribute value

Fix #7942 

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-api-tests/src/test/java/org/molgenis/api/tests/rest/v1/RestControllerV1APIIT.java
+++ b/molgenis-api-tests/src/test/java/org/molgenis/api/tests/rest/v1/RestControllerV1APIIT.java
@@ -614,7 +614,7 @@ public class RestControllerV1APIIT {
         .log()
         .all()
         .statusCode(OKE)
-        .body("label", nullValue());
+        .body("xstringnillable", nullValue());
   }
 
   @Test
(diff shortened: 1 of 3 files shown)


molgenis-api-tests/src/test/java/org/molgenis/api/tests/rest/v1/RestControllerV1APIIT.java
```diff
@@ -591,6 +592,31 @@ public void testUpdateAttributePut() {
...
+    given()
+        .log()
+        .all()
+        .header(X_MOLGENIS_TOKEN, testUserToken)
+        .get(API_V1 + "V1_API_TypeTestAPIV1/1/xstringnillable")
+        .then()
+        .log()
+        .all()
+        .statusCode(OKE)
+        .body("xstringnillable", nullValue());
```
fdlk: should be before the then I think?

ConnorStroomberg: ?? no ? send out the 'get' to check if update was succes, then when the get returns check the result

fdlk:
@ConnorStroomberg and I took a look and we think you can already upload a request body containing
`null`

PR #7950, opened 2018-11-14
Fix #7948 Job failure progress messages not coded

### LINKED ISSUE #7948: Job failure progress messages not coded
### How to Reproduce
- Throw an exception with ```getLocalizedMessage``` implementation (e.g. CodedRuntimeException) in a job

### Expected behavior
- Job fails with localized progress message with the language set to the current language of the job owner
- Job log contains the exception message based on ```Exception.getMessage()```

### Observed behavior
- Job fails with progress message based on ```Exception.getMessage()```

dennishendriksen:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-jobs/src/main/java/org/molgenis/jobs/JobExecutionTemplate.java
```diff
@@ -31,36 +33,45 @@
       Progress progress,
       Authentication authentication,
       TransactionOperations transactionOperations) {
-    return runWithAuthentication(
-        authentication, () -> tryCallInTransaction(job, progress, transactionOperations));
+    // FIXME determine locale using job username
```
tommydeboer: Resolve `FIXME`

molgenis-jobs/src/main/java/org/molgenis/jobs/NontransactionalJob.java
```diff
@@ -25,6 +26,12 @@ public NontransactionalJob(Progress progress, Authentication authentication) {
 
   @Override
   public T call() {
-    return jobExecutionTemplate.call(this, progress, authentication);
+    // FIXME determine locale using job username
```
tommydeboer: Resolve `FIXME`

molgenis-jobs/src/main/java/org/molgenis/jobs/ProgressImpl.java
```diff
@@ -90,12 +90,12 @@ public void success() {
...
         jobExecution.getLog());
     update();
-    JobExecutionContext.unset();
+    JobExecutionHolder.unset();
   }
 
   @Override
   public void failed(Exception ex) {
-    JOB_EXECUTION_LOG.error("Failed. " + ex.getMessage(), ex);
+    JOB_EXECUTION_LOG.error("Failed. " + ex.getLocalizedMessage(), ex);
```
tommydeboer: The error code should be logged as well here.

tommydeboer:
image
Not sure what to expect here, but have you considered showing the localized message above the loading bar?

molgenis-jobs/src/main/java/org/molgenis/jobs/ProgressImpl.java
dennishendriksen: ```java.lang.Exception``` doesn't have a code

dennishendriksen:
> image
> Not sure what to expect here, but have you considered showing the localized message above the loading bar?

I didn't make changes to this job UI component.

tommydeboer:
> I didn't make changes to this job UI component.

I know, I'm just wondering if you should.

molgenis-jobs/src/main/java/org/molgenis/jobs/ProgressImpl.java
tommydeboer: The whole point of the custom exceptions was to have the codes visible so we could easily figure out what went wrong, even in cases where it says "エンティティタイプテストの不明な属性テスト". 

I think in cases of `CodedRuntimeException` it should show the code.

tommydeboer: Maybe its just weird to log localized exceptions messages in the first place... 🤔

PR #8031, opened 2018-12-04
chore: No longer run findbugs analysis

fdlk:
SonarQube is the only rule engine you need

#### Checklist
- [x] Code reviewed
- [NA] Code unit/integration/system tested
- [x] Clean commits
- [NA] Added Feature/Fix to release notes

### DIFF
--- a/molgenis-data/src/main/java/org/molgenis/data/DataConverter.java
+++ b/molgenis-data/src/main/java/org/molgenis/data/DataConverter.java
@@ -3,7 +3,6 @@ package org.molgenis.data;
 import static java.util.Collections.singletonList;
 import static java.util.stream.StreamSupport.stream;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.time.Instant;
 import java.time.LocalDate;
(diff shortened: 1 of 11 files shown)


molgenis-data/src/main/java/org/molgenis/data/DataConverter.java
```diff
@@ -20,9 +19,6 @@
 import org.springframework.core.convert.ConverterNotFoundException;
 import org.springframework.core.convert.support.DefaultConversionService;
 
-@SuppressFBWarnings(
-    value = "NP_BOOLEAN_RETURN_NULL",
-    justification = "We want to return Boolean.TRUE, Boolean.FALSE or null")
```
fdlk: This one already has a @SuppressWarning at the actual method

tommydeboer: @fdlk What does Sonar say about this one?

fdlk: Nothing, I think because we suppress the sonar equivalent

PR #8057, opened 2018-12-07
Fix #7908 StackOverflowError when moving a package to its child

### LINKED ISSUE #7908: StackOverflowError when moving a package to its child
### How to Reproduce
1. Go to Package in the Data Explorer
1. Add a package _parent_
2. Add a package _child_, with package _parent_ as its parent.
3. Edit package _parent_ and set itss parent to package _child_

### Expected behavior
An error explaining that you can't move packages to its sub-packages.

### Observed behavior
```
Package changes could not be saved.
Handler dispatch failed; nested exception is java.lang.StackOverflowError
```

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/PackageValidator.java
+++ b/molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/PackageValidator.java
@@ -1,7 +1,11 @@
 package org.molgenis.data.validation.meta;
 
+import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
+import static org.molgenis.security.core.runas.RunAsSystemAspect.runAsSystem;
 
+import java.util.HashSet;
+import java.util.Set;
 import org.molgenis.data.meta.model.Package;
 import org.molgenis.data.meta.system.SystemPackageRegistry;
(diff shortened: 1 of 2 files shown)


molgenis-data-validation/src/main/java/org/molgenis/data/validation/meta/PackageValidator.java
```diff
@@ -42,4 +47,20 @@ private void validatePackageAllowed(Package aPackage) {
   private static void validatePackageName(Package aPackage) {
     NameValidator.validatePackageId(aPackage.getId());
   }
+
+  private static void validatePackageParent(Package aPackage) {
```
tommydeboer: I think you can use  `PackageUtils.contains()` here to check if the current package contains the package you want to set as a parent.

dennishendriksen: Cool, wasn't aware that such a method already existed. As mentioned offline the contains algorithm prunes the children instead of the parent. Lets check outside of this PR if we can merge the algorithms.

PR #8076, opened 2018-12-13
Fix #8074 Security manager: show group on member add

### LINKED ISSUE #8074: Security manager: show group on member add
### How to Reproduce
Go to the security manager, create a group and click 'add member'

### Expected behavior
I want to see the name of the group on the 'add member' screen

### Observed behavior
Group name disappeared and I got very confused

dennishendriksen:
#### Checklist
- [x] Code reviewed
- User documentation updated
- Test plan template updated
- [x] Clean commits

### DIFF
--- a/molgenis-security-ui/src/main/frontend/src/components/MemberAdd.vue
+++ b/molgenis-security-ui/src/main/frontend/src/components/MemberAdd.vue
@@ -5,6 +5,21 @@
 
     <div class="row mb-3  ">
       <div class="col">
+        <nav aria-label="breadcrumb">
+          <ol class="breadcrumb">
+            <li class="breadcrumb-item">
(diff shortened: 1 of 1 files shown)


marijevdgeest:
I think appending the information to the task label ("Add member to <group name>", that's the label right?) would also do the trick. If you prefer that over a breadcrumb I would be equally happy.

dennishendriksen:
Personally I prefer the breadcrumb to give context about the group to which you want to add a user since it displays the context in the same manner as the screen that I navigated from. With the label-with-group-name solution the label 'jumps up' when selecting the 'Add Member' button.

8074_ux

Does the use of a breadcrumb imply the use of subgroups? How should the label look like in this case?
@mswertz any preferences for you?

PR #8078, opened 2018-12-13
Nullable annotations on core data interfaces & classes

dennishendriksen:
This PR is expected to result in an increase of Sonar issues.

#### Checklist
- [x] Code reviewed
- [NA] Code unit/integration/system tested
- [x] Clean commits
- [NA] Added Feature/Fix to release notes

### DIFF
--- a/molgenis-amazon-bucket/src/main/java/org/molgenis/amazon/bucket/meta/AmazonBucketJobExecution.java
+++ b/molgenis-amazon-bucket/src/main/java/org/molgenis/amazon/bucket/meta/AmazonBucketJobExecution.java
@@ -1,6 +1,7 @@
 package org.molgenis.amazon.bucket.meta;
 
 import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
 import org.molgenis.data.Entity;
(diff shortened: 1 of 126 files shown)


dennishendriksen:
> Is it necessary to use both @Nullable and @CheckForNull? IntelliJ can handle both.

@tommydeboer Yes, although IntelliJ checks Nullable, Sonar only checks CheckForNull. Also see the Nullable documentation:
```
This annotation is useful mostly for overriding a {@link Nonnull} annotation.
 * Static analysis tools should generally treat the annotated items as though they
 * had no annotation, unless they are configured to minimize false negatives.
 * Use {@link CheckForNull} to indicate that the element value should always be checked
 * for a null value.
```

tommydeboer:
@dennishendriksen What I meant was, do we still need to use `@Nullable`?

dennishendriksen:
@tommydeboer Third-party dependencies like Google AutoValue use @Nullable instead of @CheckForNull. To keep in line with both worlds, different IDEs and different static code analysis tools I think it is best to add both.

PR #8113, opened 2019-01-08
Metadata tests 

tommydeboer:
- DynamicDecorator 
- Scripts 
- Two Facto Authentication
- Security 
- Mapping Service 
- Jobs
- Index

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-security/src/test/java/org/molgenis/data/security/auth/TokenTest.java
```diff
@@ -1,43 +1,20 @@
...
-
-  @Test
-  public void testIsExpiredNoExpirationDate() throws Exception {
-    assertFalse(token.isExpired());
-  }
-
-  @Test
-  public void testIsExpiredExpirationDateInFuture() throws Exception {
-    token.setExpirationDate(now().plus(1, MINUTES));
-    assertFalse(token.isExpired());
```
bartcharbon: Wouldn't we like to keep the specific tests? this behaviour is not tested in the new test framework

molgenis-security/src/main/java/org/molgenis/security/twofactor/model/UserSecretFactory.java
```diff
@@ -6,8 +6,9 @@
 
 @Component
 public class UserSecretFactory
-    extends AbstractSystemEntityFactory<UserSecret, UserSecretMetaData, String> {
-  UserSecretFactory(UserSecretMetaData userSecretMetaData, EntityPopulator entityPopulator) {
+    extends AbstractSystemEntityFactory<UserSecret, UserSecretMetadata, String> {
+
```
bartcharbon: Assuming this is the result of a "refactor" action I'm not sure how this newline ended up here, but I think we want to remove it right?

molgenis-data-security/src/test/java/org/molgenis/data/security/auth/TokenTest.java
tommydeboer: Sharp! Not sure what I was thinking when I removed the original file...

PR #8131, opened 2019-01-11
Make header of datarowedit plugin h1 rather than h5 to fix broken styling

marikaris:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data-row-edit/src/main/frontend/src/components/DataRowEdit.vue
+++ b/molgenis-data-row-edit/src/main/frontend/src/components/DataRowEdit.vue
@@ -15,7 +15,7 @@
 
     <div v-if="showForm">
 
-      <h5>{{dataTableLabel}}</h5>
+      <h1>{{dataTableLabel}}</h1>
 
       <form-component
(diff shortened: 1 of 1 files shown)


molgenis-data-row-edit/src/main/frontend/src/components/DataRowEdit.vue
```diff
@@ -15,7 +15,7 @@
 
     <div v-if="showForm">
 
-      <h5>{{dataTableLabel}}</h5>
+      <h1>{{dataTableLabel}}</h1>
```
ConnorStroomberg: This seems strange to me as this is the label of a form, a suspect we should reserve h1 for main page headers ?
maybe jou could add screen shot to make clear the issue this PR is trying to solve

ConnorStroomberg: i see what you mean, but i'm thinking why is the 'xcomound label' that big ..

PR #8136, opened 2019-01-11
Feature/5941 c7 themes

marikaris:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


molgenis-core-ui/src/main/frontend/yarn-error.log
```diff
@@ -0,0 +1,6618 @@
+Arguments:
```
ConnorStroomberg: i dont think we should add the error log ( maybe added by accident?) , suggest to remove it

PR #8150, opened 2019-01-15
fix(header): Add polyfill to all pages.

fdlk:
Some apps don't render properly on old browsers. Add the complete babel polyfill to all pages.

N.B. This has as a consequence that apps cannot also import the polyfill, themselves.
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/docs/developer_documentation/app-development.md
+++ b/docs/developer_documentation/app-development.md
@@ -77,7 +77,8 @@ It contains the following parameters
 | version | string |yes |The version of your app. Used to manage app version, updating etc etc.. |
 | apiDependency | string |yes | The version of the MOLGENIS REST api you used in your app. Can be used to give warnings about possible incompatibility with a specific version of MOLGENIS |
 | name | string |yes | The name of the app. May not contain '/'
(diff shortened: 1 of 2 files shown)


fdlk:
Babel preset env polyfilling allows you to specify the range of browsers you want to serve and it inspects the code for each app, allowing you to limit the polyfills you use per app which results in smaller bundle sizes. Current strategy for molgenis therefore is that version 2 pages provide their own polyfill.

But! This is quite hard to get right for a couple of reasons.
* The polyfills that change global object prototypes can only be run once and babel gets very angry when they get run twice.
* Babel does not inspect node_modules code but dependencies of your app may be depending on polyfills. For instance, vuex depends on promise. You need to specify those as well.
* We have three vue apps on the version 2 pages: menu app and footer. I tried packaging the biobank explorer app with the polyfill and then the menu fails to run because it is loaded before the app.
* Babel is changing the way the preset env specifies its requirements to allow dependencies to generate a series of specific polyfills they need to work
* Babel 7 has removed object prototype polyfills altogether

PR #8159, opened 2019-01-16
[FEATURE] added war file argument to docker-compose. updated readme

MaxPostema:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-app/docker-compose.yml
+++ b/molgenis-app/docker-compose.yml
@@ -1,7 +1,11 @@
 version: "3.1"
 services:
     molgenis:
-      image: registry.molgenis.org/molgenis/molgenis-app:${TAG:-latest}
+      build:
+        context: .
+        args:
+          WAR_FILE: target/molgenis-app-*.war
+
       ports:
         - 8081:8080
       environment:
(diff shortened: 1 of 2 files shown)


molgenis-app/docker-compose.yml
```diff
@@ -1,7 +1,11 @@
 version: "3.1"
 services:
     molgenis:
-      image: registry.molgenis.org/molgenis/molgenis-app:${TAG:-latest}
+      build:
+        context: .
+        args:
+          WAR_FILE: target/molgenis-app-*.war
```
fdlk: I currently use this to locally run preview versions of molgenis by specifying a tag in the registry.

MaxPostema: @fdlk I added an env variable with default value and updated documentation.
Will this work with your workflow?

fdlk: In the target dir you'll only find the version you've checked out and built, so I don't see a use case where you need to specify a different version for the war file in the target dir.

I used this docker-compose to specify a specific docker image image tag instead, but I believe that the changes made to the molgenis/docker repository will allow me to use the docker-compose specified there now.

PR #8199, opened 2019-01-29
Fix #4331 Fix #4332 Fix #4516 Remove unused VCF writer

### LINKED ISSUE #4331: VCF writer declares String in the header when the values are Floats or Integers
For some VCF annotators, including ExAC, the data types are declared wrong. Examples:

```
##INFO=<ID=EXAC_AF,Number=.,Type=String,Description="The ExAC allele frequency">
```

This should be 'Float'.

```
##INFO=<ID=EXAC_AC_HOM,Number=.,Type=String,Description="The ExAC homozygous alternative genotype count">
##INFO=<ID=EXAC_AC_HET,Number=.,Type=String,Description="The ExAC heterozygous genotype count">
```

These should be 'Integer'.

### LINKED ISSUE #4332: VCF writer declares "Number=.", even when there is only 1 value
VCF writer will always write out "Number=.". (=the number of values listed inside a 'single value', where "." means 'any') This is not wrong by itself, but since MOLGENIS now expects a list, it will convert ANY data type (Integer, Float, etc) as String because it cannot store lists as a value. This results in being unable to properly sort and filter on numerical data (CADD scores, allele frequencies etc).

However, when Number=1 is declared, it will respect the original data type and keep numbers as numbers. In many cases, VCFs with only 1 alternative allele are uploaded and thus any values are 'single values'. For example, say we have a VCF file where there are NO alternative alleles present, so data that looks like this for every line: ref=A, alt=G, EXAC_AF=0.0012, CADD_SCALED=24.62. Here, EXAC_AF and CADD_SCALED may be declared as:

```
##INFO=<ID=EXAC_AC_HOM,Number=1,Type=Integer,Description="abc">
##INFO=<ID=CADD_SCALED,Number=1,Type=Float,Description="abc">
```

Resulting in correct import, instead of:

```
##INFO=<ID=EXAC_AC_HOM,Number=.,Type=Integer,Description="abc">
##INFO=<ID=CADD_SCALED,Number=.,Type=Float,Description="abc">
```

Resulting in a bad import. The secondary issue is how MOLGENIS handles lists internally, because it SHOULD work always correctly with "Number=." declared, but for these cases a workaround solution could be really helpful.

### LINKED ISSUE #4516: Molgenis CmdLineAnnotator should print its version in the output VCF header
It would be very helpful to know which version of the CmdLineAnnotator was used to process data. Newer versions may be better at mapping variants, so knowing that an older version was used on your data may trigger you to re-run with a newer version. Now, you don't know which version you used.

Suggestion, put an extra line in the header to denote the version, for example:
```
##MolgenisCmdLineAnnotator="1.16.2 (build 2016-02-15), by Jenkins automation server at http://molgenis.org/jenkins/"
```

Question: what happens when you use different MolgenisCmdLineAnnotator versions on the same file?

dennishendriksen:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-vcf/src/test/java/org/molgenis/data/vcf/utils/test/VcfUtilsTest.java
```diff
@@ -13,7 +13,6 @@
 import static org.molgenis.data.vcf.model.VcfAttributes.POS;
 import static org.molgenis.data.vcf.model.VcfAttributes.QUAL;
 import static org.molgenis.data.vcf.model.VcfAttributes.REF;
-import static org.molgenis.data.vcf.utils.VcfWriterUtils.EFFECT;
```
bartcharbon: Entire setup of this test is obsolete, only test that is executed (createId) requires 8 lines of setup. I will merge this PR anyway. I'll create a cleanup PR for this test.

PR #8210, opened 2019-02-01
Fix #6025 LookupAttributeIndex differs depending on the way they are …

### LINKED ISSUE #6025: LookupAttributeIndex differs depending on the way they are created
### How to Reproduce
Import an entity with an attribute for which label is true and lookup is false
Check the lookupAttributeIndex
delete
Import the same entity with the attribute label is true and now lookup is also true
Check the lookupAttributeIndex

### Expected behavior
Indices are the same since both times the entity has exactly 1 lookup attr

### Observed behavior
In the one case the index is 0 and in the other index = 1

bartcharbon:
…created

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-data-import/src/main/java/org/molgenis/data/importer/emx/IntermediateParseResults.java
```diff
@@ -95,14 +96,22 @@ public void addAttributes(String entityTypeId, List<EmxAttribute> emxAttrs) {
...
+    if (ownIdAttr != null && ownIdAttr.isVisible()) {
+      ownIdAttr.setLookupAttributeIndex(lookupAttributeIndex);
+      lookupAttributeIndex++;
+    }
+    Attribute ownLabelAttr = entityType.getOwnLabelAttribute();
+    if (ownLabelAttr != null
+        && ownIdAttr != null
+        && !ownIdAttr.getName().equals(ownLabelAttr.getName())) {
+      ownLabelAttr.setLookupAttributeIndex(lookupAttributeIndex);
+      lookupAttributeIndex++;
```
tommydeboer: Sonar reports an error on this line: 
- Remove this useless assignment to local variable "lookupAttributeIndex"

PR #8258, opened 2019-03-08
[FEAT] Update OpenCPU setting to can make use of cloud instance

sidohaakma:
When you target a cloud instance, there will be indirection between servers such as loadbalancers and stuff. When you add a port number in the URL these services will not accept it.

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-r/src/main/java/org/molgenis/r/RScriptExecutor.java
+++ b/molgenis-r/src/main/java/org/molgenis/r/RScriptExecutor.java
@@ -207,19 +207,21 @@ public class RScriptExecutor {
   private URI getOpenCpuUri() {
     try {
       URI uri;
-      if(openCpuSettings.getPort() == null) {
-        uri = new URIBuilder()
-            .setScheme(openCpuSettings.getScheme())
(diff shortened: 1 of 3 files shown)


molgenis-r/src/main/java/org/molgenis/r/RScriptExecutor.java
```diff
@@ -206,12 +206,24 @@ private URI getScriptGetFileResponseUri(String openCpuSessionKey, String fileId)
...
-          .setScheme(openCpuSettings.getScheme())
-          .setHost(openCpuSettings.getHost())
-          .setPort(openCpuSettings.getPort())
-          .setPath(openCpuSettings.getRootPath())
-          .build();
+      URI uri;
+      if (openCpuSettings.getPort() == null) {
+        uri =
+            new URIBuilder()
+                .setScheme(openCpuSettings.getScheme())
```
fdlk: Most of this can be shared between both cases.

sidohaakma:
Just add -1 as port number and it will work.

PR #8314, opened 2019-03-28
Fix #8286 Log PostgreSqlExceptionTranslator exceptions of unexpected class

### LINKED ISSUE #8286: Exception gets eaten by postgresql exception translator
### How to Reproduce
Inspect: https://github.com/molgenis/molgenis/blob/6b342e01a4916e7d5205783af8c99b3d2f61d6cb/molgenis-data-postgresql/src/main/java/org/molgenis/data/postgresql/PostgreSqlExceptionTranslator.java#L89

```
Caused by: java.lang.RuntimeException: Unexpected exception class [SQLException]
	at org.molgenis.data.postgresql.PostgreSqlExceptionTranslator.doTranslate(PostgreSqlExceptionTranslator.java:92)
	at org.molgenis.data.postgresql.PostgreSqlExceptionTranslator.doTranslate(PostgreSqlExceptionTranslator.java:61)
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:72)
	at org.springframework.jdbc.core.JdbcTemplate.translateException(JdbcTemplate.java:1402)

```

https://github.com/spring-projects/spring-framework/blob/0204b082b1d4b008fc6dc94a0085f1a3baf03ca1/spring-jdbc/src/main/java/org/springframework/jdbc/support/AbstractFallbackSQLExceptionTranslator.java#L95
says that a method should return null, if it doesn't recognize the exception

### Expected behavior
I get detailed information about my exception

### Observed behavior
I only get its name (SQLException), which is not very informative

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [na] Code unit/integration/system tested
- [na] User documentation updated
- [x] Clean commits

### DIFF
--- a/molgenis-data-postgresql/src/main/java/org/molgenis/data/postgresql/PostgreSqlExceptionTranslator.java
+++ b/molgenis-data-postgresql/src/main/java/org/molgenis/data/postgresql/PostgreSqlExceptionTranslator.java
@@ -85,6 +85,7 @@ class PostgreSqlExceptionTranslator extends SQLErrorCodeSQLExceptionTranslator
       sqlException = sqlException.getNextException();
     }
     if (!(sqlException instanceof PSQLException)) {
+      LOG.error("", sqlException);
       throw new RuntimeException(
           format("Unexpected exception class [%s]", sqlException.getClass().getSimpleName()));
     }


molgenis-data-postgresql/src/main/java/org/molgenis/data/postgresql/PostgreSqlExceptionTranslator.java
```diff
@@ -85,6 +85,7 @@ private MolgenisDataException doTranslate(SQLException sqlException) {
       sqlException = sqlException.getNextException();
     }
     if (!(sqlException instanceof PSQLException)) {
+      LOG.error("", sqlException);
```
fdlk: The contract with the framework is to return null and let some fallback exception handler kick in.

PR #8329, opened 2019-04-02
Fix #8292 Account unreachable when 'Forgot password' mail not sent

### LINKED ISSUE #8292: Account unreachable when 'Forgot password' mail not send
### How to Reproduce
- Disable mail server, e.g. by turning it off or invalidating settings
- As admin user select 'Forgot password?' and enter admin mail address
- Select 'Reset password'

### Expected behavior
- Error stating that password reset failed due to a server issue
- I can still sign in with my old password

### Observed behavior
- I can't sign in with my old password anymore
image

In the case that the system has only one superuser and the password reset was requested by ANY user the system becomes unmanageable.

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-security/src/main/resources/l10n/security_en.properties
+++ b/molgenis-security/src/main/resources/l10n/security_en.properties
@@ -1,4 +1,4 @@
-SEC01=
+SEC01=Password reset failed.
 SEC02=The password reset link has expired.
 SEC03=The password reset link is invalid.
 SEC04=The password reset link is invalid.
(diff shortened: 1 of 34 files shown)


molgenis-security/src/main/resources/l10n/security_en.properties
```diff
@@ -0,0 +1,4 @@
+SEC01=
```
bartcharbon: missing message

PR #8445, opened 2019-04-18
Fix/8443 context is not a plugin

connoratrug:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-security/src/main/java/org/molgenis/security/MolgenisWebAppSecurityConfig.java
```diff
@@ -404,7 +407,7 @@ protected void configure(AuthenticationManagerBuilder auth) {
       authenticationProvider.setPreAuthenticationChecks(userDetailsChecker());
       auth.authenticationProvider(authenticationProvider);
     } catch (Exception e) {
-      throw new RuntimeException(e);
+      throw new WebAppSecurityConfigException(e);
```
sidohaakma: This does not need to be translated. This RuntimeException will be picked up by the GlobalExceptionHandling and will return something like "Something went wrong". (Conferred with platform team).

PR #8446, opened 2019-04-19
[FIX] RPM molgenis-server.properties generation is updated

sidohaakma:
- Interactive-mode is deleted, this is not a good practise as RPM's should be non-interactive always
- Added tomcat-user to molgenis group to be able to write to /home/molgenis
- Added chmod 770 (read/write for molgenis group) to /home/molgenis

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-app/pom.xml
+++ b/molgenis-app/pom.xml
@@ -70,7 +70,7 @@
           <mappings>
             <mapping>
               <directory>/usr/local/share/molgenis/war</directory>
-              <filemode>700</filemode>
+              <filemode>755</filemode>
               <username>molgenis</username>
               <groupname>molgenis</groupname>
               <sources>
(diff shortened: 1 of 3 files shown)


molgenis-app/pom.xml
```diff
@@ -70,7 +70,7 @@
           <mappings>
             <mapping>
               <directory>/usr/local/share/molgenis/war</directory>
-              <filemode>700</filemode>
+              <filemode>755</filemode>
```
fdlk: should be 555, probably?

molgenis-app/rpm/preinstall.bash
```diff
@@ -6,11 +6,25 @@ echo "########################################"
...
+
+TOMCAT_INSTALLED==$(rpm -qa | grep tomcat)
+if [[ ! -z ${TOMCAT_INSTALLED} ]]
+then
+  echo "[INFO] Tomcat is installed"
+  echo "[INFO] Add tomcat-user to molgenis-group to be able to write in MOLGENIS-homedir"
+  usermod -g molgenis tomcat
+  echo "[INFO] Allow users other than the user molgenis to write in molgenis-homedir"
+  echo "[INFO] If they are part of the molgenis-group"
+  chmod -R 770 /home/molgenis
```
fdlk: Shouldn't the RPM be simply put in the tomcat webapps dir?

PR #8571, opened 2019-07-09
[CHORE] rewrite /@molgenis-ui to /ui

sidohaakma:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-app/docker-compose.yml
+++ b/molgenis-app/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.1"
 services:
     nginx:
-      image: nginx:stable
+      image: molgenis/nginx:stable
       volumes:
         - ./nginx.conf:/etc/nginx/nginx.conf
         - ./redirect-map.conf:/etc/nginx/redirect-map.conf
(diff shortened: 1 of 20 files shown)


molgenis-core-ui/src/main/java/org/molgenis/core/ui/MolgenisWebAppConfig.java
```diff
@@ -134,7 +134,7 @@ public void addResourceHandlers(ResourceHandlerRegistry registry) {
         .setCachePeriod(cachePeriod);
     registry
         .addResourceHandler(PATTERN_MOLGENIS_UI)
-        .addResourceLocations("/@molgenis-ui/", "classpath:/@molgenis-ui/")
+        .addResourceLocations("/ui/", "classpath:/@molgenis-ui/")
```
fdlk: Looks like they need to match, this does not serve the resources

fdlk: (at least I do not get a menu on the preview server)

molgenis-app/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3.1"
 services:
     nginx:
-      image: nginx:stable
+      image: molgenis/nginx:stable
```
connoratrug: is this a different image ? , or the same from other repo ?

sidohaakma: This is our own located on hub.docker.com on the molgenis organisation. The configuration is located in the molgenis-frontend repository. The version needs to be updated.

dennishendriksen:
can be closed according to @sidohaakma. could you add a comment explaining why for future reference?

PR #8611, opened 2019-08-15
Data API v3: disallow entity type/attribute create/update/delete

dennishendriksen:
#### Checklist
- [x] Code reviewed
- User documentation updated
- Test plan template updated
- [x] Clean commits

### DIFF
--- a/molgenis-api-data/src/main/java/org/molgenis/api/data/v3/DataServiceV3Impl.java
+++ b/molgenis-api-data/src/main/java/org/molgenis/api/data/v3/DataServiceV3Impl.java
@@ -2,6 +2,9 @@ package org.molgenis.api.data.v3;
 
 import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.toList;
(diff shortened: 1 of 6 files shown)


molgenis-api-data/src/main/java/org/molgenis/api/data/v3/DataServiceV3Impl.java
```diff
@@ -244,6 +248,10 @@ public void deleteAll(String entityTypeId, @Nullable @CheckForNull Query query)
   }
 
   private Repository<Entity> getRepository(String entityTypeId) {
+    if (entityTypeId.equals(ENTITY_TYPE_META_DATA) || entityTypeId.equals(ATTRIBUTE_META_DATA)) {
+      throw new RepositoryNotCapableException(entityTypeId, WRITABLE);
```
bartcharbon: Also happens for "GET" endpoints, "WRITABLE" is not correct in those cases

PR #8617, opened 2019-08-20
Fix dataApiV3 previous/self/next urls

bartcharbon:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-api-tests/src/test/resources/org/molgenis/api/tests/data/v3/deleteResourceCollection.json
+++ b/molgenis-api-tests/src/test/resources/org/molgenis/api/tests/data/v3/deleteResourceCollection.json
@@ -1,6 +1,6 @@
 {
   "links": {
-    "self": "${baseUri}/api/data/v3_MyDataset"
+    "self": "${baseUri}/api/data/v3_MyDataset?size=100&page=0"
   },
   "items": [],
   "page": {
--- a/molgenis-api/src/main/java/org/molgenis/api/convert/UrlParameterUtils.java
+++ b/molgenis-api/src/main/java/org/molgenis/api/convert/UrlParameterUtils.java
@@ -1,0 +1,152 @@
+package org.molgenis.api.convert;
+
+import static org.molgenis.api.model.Sort.Order.Direction.ASC;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.molgenis.api.model.Query;
+import org.molgenis.api.model.Query.Operator;
+import org.molgenis.api.model.Selection;
+import org.molgenis.api.model.Sort;
+import org.molgenis.util.UnexpectedEnumException;
+
+public class UrlParameterUtils {
+
+  public static final String PAGE = "page";
+  public static final String SIZE = "size";
+  public static final String QUERY = "q";
+  public static final String SORT = "sort";
+  public static final String SELECTION = "selection";
+  public static final String FILTER = "filter";
+
+  private UrlParameterUtils() {
+    // emptu constructor to hide default.
+  }
+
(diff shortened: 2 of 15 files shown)


molgenis-api/src/main/java/org/molgenis/api/convert/UrlParameterUtils.java
```diff
@@ -0,0 +1,152 @@
...
+
+    return params;
+  }
+
+  public static String createUrlParamString(
+      Selection filter, Selection expand, Sort sort, int size, int number, Optional<Query> query) {
+    Map<String, String> paramMap = createUrlParams(filter, expand, sort, size, number, query);
+    return paramMap.entrySet().stream()
+        .filter(entry -> !entry.getValue().isEmpty())
+        .map(entry -> entry.getKey() + "=" + entry.getValue())
```
dennishendriksen: ```entry.getValue()``` requires escaping

molgenis-api/src/main/java/org/molgenis/api/convert/UrlParameterUtils.java
```diff
@@ -0,0 +1,152 @@
...
+  static String queryToUrlParam(Query query) {
+    return queryToUrlParam(query, false);
+  }
+
+  static String queryToUrlParam(Query query, boolean hasParent) {
+    String param;
+    Operator queryOperator = query.getOperator();
+    switch (queryOperator) {
+      case EQUALS:
+        param = query.getItem() + "==" + query.getStringValue();
```
dennishendriksen: ```query.getStringValue()``` might double quoting, same goes for other cases in this switch.

molgenis-api/src/main/java/org/molgenis/api/convert/UrlParameterUtils.java
```diff
@@ -0,0 +1,152 @@
...
+        break;
+      default:
+        throw new UnexpectedEnumException(queryOperator);
+    }
+    return param;
+  }
+
+  static String sortToUrlParam(Sort sort) {
+    return sort.getOrders().stream()
+        .map(order -> (order.getDirection() == ASC ? "+" : "-") + order.getItem())
```
dennishendriksen: ```+``` should be removed due to changes in #8621

molgenis-api/src/main/java/org/molgenis/api/convert/UrlParameterUtils.java
```diff
@@ -0,0 +1,152 @@
...
+                .collect(Collectors.joining(","));
+        param = hasParent ? "(" + orQuery + ")" : orQuery;
+        break;
+      default:
+        throw new UnexpectedEnumException(queryOperator);
+    }
+    return param;
+  }
+
+  static String sortToUrlParam(Sort sort) {
```
dennishendriksen: Move to SortConverter so that we 'from' and 'to' converters in the same class? Same for *ToUrlParam converters.

molgenis-api-tests/src/test/resources/org/molgenis/api/tests/data/v3/deleteResourceCollection.json
```diff
@@ -1,6 +1,6 @@
 {
   "links": {
-    "self": "${baseUri}/api/data/v3_MyDataset"
+    "self": "${baseUri}/api/data/v3_MyDataset?size=100&page=0"
```
dennishendriksen: How about either:
- using exactly the URL that the user entered
- not including query parameters when the value is the default value

dennishendriksen:
When using ```ServletUriComponentsBuilder.fromCurrentRequest()``` instead of ```ServletUriComponentsBuilder.fromCurrentRequestUri()``` you only need to replace the page query parameter for previous/next.

PR #8639, opened 2019-08-29
UniqueId unit tests

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-data/src/test/java/org/molgenis/data/util/UniqueIdTest.java
+++ b/molgenis-data/src/test/java/org/molgenis/data/util/UniqueIdTest.java
@@ -1,0 +1,20 @@
+package org.molgenis.data.util;
+
+import static org.testng.Assert.assertNotEquals;
+
+import org.testng.annotations.Test;
+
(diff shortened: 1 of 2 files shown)


molgenis-data/src/test/java/org/molgenis/data/util/UniqueIdTest.java
```diff
@@ -0,0 +1,20 @@
...
+import static org.testng.Assert.assertNotEquals;
+
+import org.testng.annotations.Test;
+
+public class UniqueIdTest {
+  @Test
+  public void testGetId() {
+    byte[] id0 = new UniqueId().getId();
+    byte[] id1 = new UniqueId().getId();
+    assertNotEquals(id0, id1);
```
fdlk: Each instance of UniqueId has an internal counter that increments when an id is generated and resets each millisecond.

`id0` and `id1` will be equal when both getId calls fall within the same millisecond.
This is because both instances of UniqueId have their counter set to 0.

PR #8688, opened 2019-09-25
Fix one-to-many tests

bartcharbon:
#### Checklist
- [N.A.] Functionality works & meets specifications
- [x] Code reviewed
- [N.A.] Code unit/integration/system tested
- [x] Clean commits


molgenis-data/src/main/java/org/molgenis/data/DataService.java
```diff
@@ -218,28 +218,32 @@
   <E extends Entity> E findOne(String entityTypeId, Query<E> q, Class<E> clazz);
 
   /**
-   * Adds an entity to it's repository
+   * Adds an entity to it's repository Note: when the entityType contains one-to-many attributes,
```
dennishendriksen: Better?

Note: the caller is responsible for updating the other side of bidirectional relationships (e.g. one-to-many attributes).

PR #8711, opened 2019-10-10
Export DATE_TIME in UTC

fdlk:
We do not store timezone information for DATE_TIME so it is more accurate to export in UTC.
As a bonus this makes the export timezone independent.
Also switches from SimpleDateFormatter to the 310 formatter which is thread-safe.

#### Checklist
- [x] Code reviewed
- User documentation updated
- Test plan template updated
- [x] Clean commits


sidohaakma (approved):

sidohaakma:
[Export DateTime in UTC](https://trello.com/c/owbTs5Xq/494-export-datetime-in-utc)

PR #8722, opened 2019-10-15
M5207 Remove unused code

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-api-data/src/main/java/org/molgenis/api/data/v1/AttributeResponse.java
```diff
@@ -274,6 +265,7 @@ public Boolean getLookupAttribute() {
     return lookupAttribute;
   }
 
+  @SuppressWarnings("unused")
   public Boolean isAggregatable() {
```
dennishendriksen: IntelliJ reports that the following methods are unused as well:
```java
public List<String> getEnumOptions()
public String getMappedBy()
public Object getDefaultValue()
public Range getRange()
public String getExpression()
public String getNullableExpression()
public String getVisibleExpression()
public String getValidationExpression()
```
Manually inspecting the usages of other getters results in none of them being used. Is there a reason for only suppressing this method?

Considering that this class is in practice only used the GSON message converter for serialization we could drop all the accessors according to the [GSON user guide](https://github.com/google/gson/blob/master/UserGuide.md#finer-points-with-objects).

What do you think?

molgenis-api-data/src/main/java/org/molgenis/api/data/v1/SortV1.java
```diff
@@ -17,10 +17,8 @@
 
 import java.io.Serializable;
```
dennishendriksen: FYI The class docs state that ```Copy of org.springframework.data.domain.Sort for backwards compatibility```. I checked whether the [current implementation](https://github.com/spring-projects/spring-data-commons/blob/master/src/main/java/org/springframework/data/domain/Sort.java) is still similar to ours, because in that case I would favor suppressions. But since the current implementation is quite different from ours I prefer your solution.

molgenis-api-data/src/main/java/org/molgenis/api/data/v2/AttributeResponseV2.java
```diff
@@ -249,6 +249,7 @@ public Boolean getLookupAttribute() {
     return lookupAttribute;
   }
 
+  @SuppressWarnings("unused")
```
dennishendriksen: See AttributeResponse comments.

molgenis-api-data/src/main/java/org/molgenis/api/data/v2/CopyEntityRequestV2.java
```diff
@@ -12,6 +12,7 @@
   @NotNull
   public abstract String getNewEntityName();
 
+  @SuppressWarnings("unused")
```
dennishendriksen: Both IntelliJ and Sonarlint are not reporting issues in my IDE?

Similar to the AttributeResponse comments I think that we can remove all create methods and builders from AutoValue objects unless they are used in unit tests (which they should).

tommydeboer: As discussed, figuring out which methods can be removed or should be suppressed is work for another time. For now I'll remove the annotations

dennishendriksen:
@tommydeboer could you fix the merge conflict?

tommydeboer:
PR will be split up

PR #8735, opened 2019-10-24
M6763 Asynchronous EntityType/Attribute delete endpoints

tommydeboer:
**Suggestion:**
This PR adds a `deleteEntityTypes(Collection<String> entityTypeIds)` method to the `MetaDataService` because that made more sense in this context. The already existing `deleteEntityType(Collection<EntityType> entityTypes)` is only used in one place during bootstrapping. Seeing as how most delete methods are 'by ID', I'd prefer it if we drop the latter and use the new method instead. That would however change the interface and I'm unsure of the impact outside of this repo (gut feeling: low impact). What do you think?


dennishendriksen:
@tommydeboer good idea. how about a seperate pr to the master with the metadata service changes?

tommydeboer:
@dennishendriksen Done: https://github.com/molgenis/molgenis/pull/8739

PR #8739, opened 2019-10-30
Add MetaDataService.deleteEntityTypes(Collection<String>)

tommydeboer:
#### Checklist
- [x] Code reviewed
- [n.a.] Migration step added in case of breaking change
- [n.a.] User documentation updated
- [n.a.] (If you have changed REST API interface) view-swagger.ftl updated
- [n.a.] Test plan template updated
- [x] Clean commits
- [n.a.] Added Feature/Fix to release notes


molgenis-data-platform/src/main/java/org/molgenis/data/meta/system/SystemEntityTypePersister.java
```diff
@@ -106,17 +106,18 @@ private void persistMetadataMetadata() {
   /** Package-private for testability */
   void removeNonExistingSystemEntityTypes() {
     // get all system entities
-    List<EntityType> removedSystemEntityMetas =
+    List<String> removedSystemEntityIds =
         dataService
```
dennishendriksen: Since you only need the id adding a fetch will speed things up a bit.

PR #8755, opened 2019-11-12
PATCH entitytype endpoint

bartcharbon:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


tommydeboer:
**Issue:**
Trying to set labelAttribute/lookupAttributes to non-existing attributes doesn't give an error:

tommydeboer:
**Issue:**
`IdModificationException` not localized? (on preview server):
```json
 "type": "https://preview-pr-8755-16.dev.molgenis.org/problem",
  "title": "Bad Request",
  "status": 400,
  "detail": "#MAPI03#",
  "errorCode": "MAPI03",
```

molgenis-api-metadata/src/test/resources/l10n/api-metadata_en.properties
```diff
@@ -1 +1,4 @@
-MAPI01=Query ''{0}'' didn't lead to any results
\ No newline at end of file
+MAPI01=Query ''{0}'' didn''t lead to any results.
+MAPI02=An entityType cannot have zero attributes.
+MAPI03=It is not allowed to edit the identifier of an attribute.
+MAPI04=Field ''{0}'' is unknown for ''{1}'' requests.
```
tommydeboer: I was trying to update the `backend` and `tags` fields and I think this message is a bit confusing: `"Field 'tags' is unknown for 'entityType' requests."`

In these cases I'd expect something like `"Updating field 'backend' of 'entityType' is not allowed"`
In the cases where the field really doesn't exist maybe an `UnknownAttributeException` would suffice.

_BTW:_ This exception also isn't localised on the preview server, haven't tried locally yet.

tommydeboer:
Trying (wrong input)
```
{
  "label": "test"
}
```
results in a 500:
```json
"type": "https://preview-pr-8755-16.dev.molgenis.org/problem",
  "title": "Internal Server Error",
  "status": 500,
  "detail": "Null label"
```

molgenis-api-metadata/src/test/resources/l10n/api-metadata_en.properties
bartcharbon: Let's discuss this, since we designed the api not to know those fields its hard to make an exception in the exception that gets thrown here...

bartcharbon: Nevermind, we can't get away without specific code for these fields in other places as well, I'll create an specific exception for this

tommydeboer:
**Discuss:**
The YAML for the PATCH entityType endpoint shows the `EntityTypeCreate` schema. For the user this is sufficient but technically it's incorrect. Should we keep? Or how to document otherwise?

tommydeboer:
**Issue:**
Trying to update the `abstract_` field does create a job but then fails because the `isAbstract` attribute is `readOnly`.

I think this should fail before creating the job, and this field shouldn't be mentioned in the YAML.

tommydeboer:
**Issue:**
Values of the error message are flipped around:

```json
"Field 'entityType' is unknown for 'lookupAttribute' requests."
```

tommydeboer:
**Issue:**
Trying
```json
{
	"attributes":[{
		"id": "test",
		"name": "test",
		"type": "string"
	}]
}
```
gives:
`"Unknown attribute 'aaaac3uijhwgrhq5tlg7qpqaam' of entity type 'testAutoId'."`

tommydeboer:
**Issue:**
Using [it_emx_addAttrToAbstract.xlsx](https://github.com/molgenis/molgenis/files/3868999/it_emx_addAttrToAbstract.xlsx), when trying to set the `idAttribute` of `it_emx_addAttrToAbstract_Child2` like this:

```
{
	"idAttribute":"aaaac3uitukxfhq5tlg7qpqafe"
}
```

I get:
`"Unknown lookup attribute(s) 'aaaac3uitukxfhq5tlg7qpqaei' for EntityType 'it_emx_addAttrToAbstract_Child2'."`

tommydeboer:
`500 Internal Server Error` when trying using an invalid data type:
```json
{
	"attributes":[
	{
		"id": "aaaac3uijhwgrhq5tlg7qpqaam",
		"name": "id",
		"type": "string"
	},{
		"id": "aaaac3uijhwgrhq5tlg7qpqaaq",
		"name": "firstName",
		"type": "boolean"
	}]
} 
```

(`boolean` should be `bool`)

PR #8788, opened 2019-12-02
M6763 Fix naming inconsistency

tommydeboer:
#### Checklist
- [spec sync later] Functionality works & meets specifications
- [x] Code reviewed
- [N.A.] User documentation updated
- [spec sync later] (If you have changed REST API interface) view-swagger.ftl updated
- [x] Clean commits


molgenis-api-metadata/src/main/java/org/molgenis/api/metadata/v3/AttributeV3Mapper.java
```diff
@@ -222,7 +222,7 @@ private Attribute toAttribute(
     }
     attribute.setName(attributeRequest.getName());
     attribute.setEntity(entityType);
-    Integer sequenceNumber = attributeRequest.getSequenceNumber();
```
dennishendriksen: would be nice to have a unit test for the non-null sequence number case.

PR #8796, opened 2019-12-06
Fix #8797: OpenID connect redirect_uri sometimes doesn't take X-Forwarded-* headers into account

### LINKED ISSUE #8797: OpenID connect redirect_uri sometimes doesn't take X-Forwarded-* headers into account
### How to Reproduce
In short: try to use OpenID Connect behind a proxy.
It will fail because the redirect_uri is incorrect.

More detailed:
Start server on localhost (or on the kubernetes cluster).
Configure an OIDC client
Set authentication settings: allow signup, no moderation, use the OIDC client

Send a curl with forward header to the OIDC authorization
```
curl --verbose -H "X-Forwarded-Proto: https" -H "X-Forwarded-Host: oidc-test.example.org" http://localhost/oauth2/authorization/<oidc client>
```
(Or on the cluster, just open the URL in the browser)
Examine the redirect_uri in the Location header.

### Expected behavior
The redirect_uri is `https://oidc-test.example.org/oauth2/authorization/<oidc client>` (or on the cluster, the server URL)

### Observed behavior
*sometimes* the redirect_uri is `http://localhost/oauth2/authorization/<oidc client>` (on localhost) or
`http://oidc-test-molgenis:8080/login/oauth2/code/<oidc client>` (on cluster)

fdlk:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-security/src/main/java/org/molgenis/security/WebAppSecurityInitializer.java
+++ b/molgenis-security/src/main/java/org/molgenis/security/WebAppSecurityInitializer.java
@@ -1,7 +1,7 @@
 package org.molgenis.security;
 
 import static javax.servlet.DispatcherType.ASYNC;
-import static javax.servlet.DispatcherType.FORWARD;
+import static javax.servlet.DispatcherType.ERROR;
 import static javax.servlet.DispatcherType.REQUEST;
 
 import java.util.EnumSet;
@@ -18,6 +18,6 @@ public class WebAppSecurityInitializer extends AbstractSecurityWebApplicationIni
     Dynamic forwardedHeaderFilter =
         servletContext.addFilter("forwardedHeaderFilter", ForwardedHeaderFilter.class);
     forwardedHeaderFilter.setAsyncSupported(true);
-    forwardedHeaderFilter.addMappingForUrlPatterns(EnumSet.of(REQUEST, FORWARD, ASYNC), false, "*");
+    forwardedHeaderFilter.addMappingForUrlPatterns(EnumSet.of(REQUEST, ERROR, ASYNC), false, "*");
   }
 }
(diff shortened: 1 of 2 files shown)


molgenis-security/src/main/java/org/molgenis/security/WebAppSecurityInitializer.java
```diff
@@ -1,5 +1,23 @@
...
+public class WebAppSecurityInitializer extends AbstractSecurityWebApplicationInitializer {
 
-public class WebAppSecurityInitializer extends AbstractSecurityWebApplicationInitializer {}
+  @Override
+  protected void beforeSpringSecurityFilterChain(ServletContext servletContext) {
+    // add filters
+    Dynamic forwardedHeaderFilter =
+        servletContext.addFilter("forwardedHeaderFilter", ForwardedHeaderFilter.class);
+    forwardedHeaderFilter.setAsyncSupported(true);
+    forwardedHeaderFilter.addMappingForUrlPatterns(EnumSet.of(REQUEST, FORWARD, ASYNC), false, "*");
```
dennishendriksen: Why did you add 'FORWARD'?

fdlk: Oops, that's a leftover from previous attempt to fix it. I noticed the filter wasn't getting executed and thought perhaps the inner workings of the OIDC filter included a forward request.

I just now checked how spring boot registers the filter: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/ServletWebServerFactoryAutoConfiguration.java

They use request, async, error. So forward not needed, removing it.
Shall we add error to match what spring boot does?

dennishendriksen: Good idea.

PR #8805, opened 2019-12-11
Remove code duplication: page calculation logic

tommydeboer:
#### Checklist
- [x] Code reviewed
- [N.A.] User documentation updated
- [x] Clean commits

### DIFF
--- a/molgenis-api/src/main/java/org/molgenis/api/model/response/PageResponse.java
+++ b/molgenis-api/src/main/java/org/molgenis/api/model/response/PageResponse.java
@@ -18,12 +18,11 @@ public abstract class PageResponse {
 
   public abstract int getNumber();
 
-  public static PageResponse create(
-      int newSize, int newTotalElements, int newTotalPages, int newNumber) {
+  public static PageResponse create(int newSize, int newTotalElements, int newNumber) {
     return builder()
         .setSize(newSize)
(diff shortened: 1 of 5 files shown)


molgenis-api-data/src/main/java/org/molgenis/api/data/v3/EntityController.java
```diff
@@ -43,7 +43,7 @@
   private final DataServiceV3 dataServiceV3;
   private final EntityMapper entityMapper;
 
-  EntityController(DataServiceV3 dataServiceV3, EntityMapper entityMapper) {
+  public EntityController(DataServiceV3 dataServiceV3, EntityMapper entityMapper) {
```
dennishendriksen: Why?

molgenis-api/src/main/java/org/molgenis/api/model/response/PageResponse.java
```diff
@@ -53,4 +53,8 @@ public PageResponse build() {
       return pageResponse;
     }
   }
+
+  private static int getTotalPages(int pageSize, int totalElements) {
```
dennishendriksen: Did you consider making totalPages a derived property? (https://github.com/google/auto/blob/master/value/userguide/howto.md#memoize)

PR #8829, opened 2020-01-07
Fix #8675 Various performance improvements

### LINKED ISSUE #8675: Importing pointing study data slow
### How to Reproduce
Import an emx csv.zip containing data plus metadata for a nontrivial model like bbmri eric or the pointing study. I think what they have in common is a lot of small code lists.

### Expected behavior
It works

### Observed behavior
I do not have the patience to wait it out. Does not look like it will ever finish.
If you split it into two files, one for metadata and one for data, metadata takes a while but can be imported and data finishes near instantly

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [NA] Migration step added in case of breaking change
- [x] Clean commits

### DIFF
--- a/molgenis-data/src/main/java/org/molgenis/data/Fetch.java
+++ b/molgenis-data/src/main/java/org/molgenis/data/Fetch.java
@@ -5,6 +5,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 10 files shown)


molgenis-data/src/main/java/org/molgenis/data/Fetch.java
```diff
@@ -19,10 +20,22 @@
...
+    this(false);
+  }
+
+  /**
+   * Creates an empty valid Fetch. Can be used to improve performance in case of system entity type
+   * fetches defined by the system.
+   *
+   * @param validated whether or not the fetch has been validated
+   */
+  public Fetch(boolean validated) {
```
tommydeboer: For clarity it's a bit nicer to make this one private and use a factory method instead.

PR #8830, opened 2020-01-07
Suppress warnings: unused Decorator Factories

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-app-manager/src/main/java/org/molgenis/app/manager/decorator/AppsRepositoryDecoratorFactory.java
```diff
@@ -10,8 +10,10 @@
 import org.springframework.stereotype.Component;
 
 @Component
+@SuppressWarnings("unused")
```
dennishendriksen: I prefer to switch the order of annotations for consistency purposes:
```java
@SuppressWarnings("unused")
@Component
```
This is also what IntellliJ does with the 'Suppress for class' action.

PR #8833, opened 2020-01-07
Remove unused code in molgenis-semantic-* modules

tommydeboer:
#### Checklist
- [N.A.] Functionality works & meets specifications
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-semantic-mapper/src/main/java/org/molgenis/semanticmapper/algorithmgenerator/categorymapper/CategoryMapper.java
+++ b/molgenis-semantic-mapper/src/main/java/org/molgenis/semanticmapper/algorithmgenerator/categorymapper/CategoryMapper.java
@@ -15,7 +15,6 @@ public abstract class CategoryMapper {
   public abstract Category findBestCategoryMatch(
       Category sourceCategory, List<Category> targetCategories);
 
-  @SuppressWarnings("unused")
   public abstract CategoryMatchQuality<?> applyCustomRules(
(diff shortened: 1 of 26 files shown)


molgenis-semantic-mapper/src/main/java/org/molgenis/semanticmapper/algorithmgenerator/categorymapper/CategoryMapper.java
```diff
@@ -15,6 +15,7 @@ public CategoryMapper(List<CategoryRule> rules) {
   public abstract Category findBestCategoryMatch(
       Category sourceCategory, List<Category> targetCategories);
 
+  @SuppressWarnings("unused")
```
dennishendriksen: Used by LexicalCategoryMapper.findBestCategoryMatch according to IntelliJ

molgenis-semantic-mapper/src/main/java/org/molgenis/semanticmapper/service/MappingService.java
```diff
@@ -88,6 +88,7 @@ long applyMappings(
    * @param clonedMappingProjectName name of the cloned {@link MappingProject}
    * @return cloned {@link MappingProject}
    */
+  @SuppressWarnings("unused")
```
dennishendriksen: I think this is a valid warning and should not be suppressed. Instead the code calling this method should at least do something with id of the cloned project (not as part of this PR).

PR #8844, opened 2020-01-15
M7121 Unify restrictions on package ids, group names and role names 

tommydeboer:
#### Checklist
- [x] Code reviewed
- [ ] Clean commits


molgenis-data-migrate/src/main/java/org/molgenis/data/migrate/bootstrap/MolgenisUpgradeBootstrapper.java
```diff
@@ -32,7 +31,6 @@ public void bootstrap() {
     upgradeService.addUpgrade(new Step36EnableDataRowEdit(dataSource));
     upgradeService.addUpgrade(new Step37AddSettingsPluginToMenu(dataSource));
     upgradeService.addUpgrade(new Step38AddAnonymouseRoleToDb(dataSource));
-    upgradeService.addUpgrade(new Step39CreateRootPackageGroups(dataSource));
```
dennishendriksen: Why? According to http://wiki.gcc.rug.nl/ticket/6336 we ```Wrote a migration step that creates groups for packages-without-parent ```. I don't see how this works after your changes.

molgenis-security-core/src/test/java/org/molgenis/security/core/GroupValueFactoryTest.java
```diff
@@ -80,9 +80,9 @@ void testCreateRoleName(String groupName, String roleLabel, String roleName) {
 
   static Object[][] roleNameProvider() {
     return new Object[][] {
-      new Object[] {"bbmri-eric", "Manager", "BBMRI_ERIC_MANAGER"},
-      new Object[] {"bbmri-eric", "Manager", "BBMRI_ERIC_MANAGER"},
-      new Object[] {"patientenfederatie-nederland", "Member", "PATIENTENFEDERATIE_NEDERLAND_MEMBER"}
+      new Object[] {"bbmri_eric", "Manager", "bbmri_eric_MANAGER"},
+      new Object[] {"bbmri_eric", "Manager", "bbmri_eric_MANAGER"},
```
bartcharbon: In both the old and new situation this line seems duplicated

PR #8881, opened 2020-02-03
chore: docker stack config from env file

jvanveen:
The idea is to make the Docker config more generic by using an env file. The extra host host.docker.internal is added manually. Afaik, this should be available in Docker by default, but it wasn't on Linux. Docker host ip from the container on Linux is 172.17.0.1. Not sure if this is always the case. Using network mode 'host' while in Linux is quite useful (performance & less effort to get it working while having ufw switched on). I removed the shell scripts, since they are not necessary when using the .env file and docker-compose commands.

#### Checklist
- [x] Code reviewed
- [x] Clean commits


jvanveen:
There is somewhat of a consensus that `host.docker.internal` should be the default hostname that points to the host, instead of using mac/windows/linux specific hostnames. This is not implemented in Docker at the moment. Using the extra_hosts option and a variable is a workable solution i suppose. The ip can be resolved with(also in the docs and in the .env.example file): 
`docker inspect -f '{{range .NetworkSettings.Networks}}{{.Gateway}}{{end}}' nginx`

jvanveen:
Extra feedback:
* Add sane default values when no .env is available
* Also apply to integration-tests docker-compose setup

molgenis-app/dev-env/docker-compose.yml
```diff
@@ -1,11 +1,15 @@
...
+    image: ${MOLGENIS_FRONTEND:-molgenis/molgenis-frontend:latest}
+    container_name: frontend
     ports:
       - 80:80
+    network_mode: ${NETWORK_MODE:-bridge}
     volumes:
-      - ${BACKEND}:/etc/nginx/proxy.d/backend.conf
+      - ./backend.conf:/etc/nginx/proxy.d/backend.conf
+    extra_hosts:
+      - host.docker.internal:${MOLGENIS_HOST:-172.17.0.1}
```
fdlk: This does not work for me on mac. If I remove this line it works fine.
I believe the MOLGENIS_HOST cannot be set reliably on mac based on what I find on https://docs.docker.com/docker-for-mac/networking/: 
> The host has a changing IP address (or none if you have no network access). From 18.03 onwards our recommendation is to connect to the special DNS name host.docker.internal, which resolves to the internal IP address used by the host.

fdlk: I shelled into the frontend container and asked it to resolve host.docker.internal. It resolves to 192.168.65.2 and if I set MOLGENIS_HOST=192.168.65.2 everything works.
However, I still suspect that this IP is random.

jvanveen: I didn't like the extra_hosts option either. The whole reason for going that way was that docker-compose doesn't deal with variable interpolation in mounted volume config files(backend.conf) and that NGINX doesn't support config directives from the environment. I think I found an acceptable workaround that does the same with a little bit of bash magic. Instead of using a backend.conf, the file is being interpolated inside the container using envsubst and stored to backend.conf.

PR #8900, opened 2020-02-07
Build performance improvements

dennishendriksen:
Travis: ~1 minute faster

#### Checklist
- [x] Code reviewed
- User documentation updated
- Test plan template updated
- [x] Clean commits

### DIFF
--- a/.ci/jobs/api-tests.sh
+++ b/.ci/jobs/api-tests.sh
@@ -4,6 +4,9 @@
 export MINIO_ACCESS_KEY=molgenis
 export MINIO_SECRET_KEY=molgenis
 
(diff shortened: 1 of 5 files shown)


sidohaakma:
@dennishendriksen: Can we look at this together? Some of the changes override the default MAVEN configuration. We need to decide what to set in which part of the pipeline. I will show you where we need to take into account that the MAVEN build docker sets default configuration as well. We can determine together what the way forward should be.

dennishendriksen:
> @dennishendriksen: Can we look at this together?

Sure. Does that explain why the docker step is failing?

PR #8913, opened 2020-02-11
Fix #4080 Update active session authentication roles

### LINKED ISSUE #4080: Updating permissions for users requires them to log out and in
### To reproduce:
1. Make a new user
2. Log in with that user (don't log out!)
3. In a separate session as an admin, give the user permissions for the data explorer
4. As the user, refresh the page.
### Expected:

The user can now see the Data Explorer.
### Actual:

The Data Explorer is not visible until the user logs out and in. This goes for entity permissions as well.

dennishendriksen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-data-security/src/main/java/org/molgenis/data/security/SessionSecurityContextUpdaterImpl.java
+++ b/molgenis-data-security/src/main/java/org/molgenis/data/security/SessionSecurityContextUpdaterImpl.java
@@ -1,0 +1,88 @@
+package org.molgenis.data.security;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import org.molgenis.data.security.auth.User;
+import org.molgenis.data.transaction.TransactionListener;
+import org.molgenis.data.transaction.TransactionManager;
+import org.molgenis.security.core.AuthenticationAuthoritiesUpdater;
+import org.molgenis.security.core.PrincipalSecurityContextRegistry;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.security.core.userdetails.UserDetails;
+import org.springframework.security.core.userdetails.UserDetailsService;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SessionSecurityContextUpdaterImpl
+    implements SessionSecurityContextUpdater, TransactionListener {
+
+  private final PrincipalSecurityContextRegistry principalSecurityContextRegistry;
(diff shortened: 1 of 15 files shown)


dennishendriksen:
@fdlk @sidohaakma could you add your notes from the phone call to this PR so that we don't forget about them when looking into this PR in the future (e.g. keycloak, worries regarding security vulnerabilities, spring deals with this for you)

dennishendriksen:
@fdlk @sidohaakma *bump*

fdlk:
We don't have notes from the phonecall but some considerations:

Existing behavior:
* For the release the bug to solve was smaller, namely that if the session auth roles are out of date (existing behavior) molgenis suddenly started showing error page (new bug, easy to fix)

Correctness:
* When we start a job in the user's name it runs with a security context not registered in the session registry and it will not get updated.
* When multiple threads finish adding roles (think e.g. a commander script that triggers quick successive role membership updates), you get nasty race conditions that mean the end result may be incorrect
* Adding nonstandard security code is tricky, may introduce vulnerabilities that we cannot oversee.

Heavily relies on existing session mechanism and existing architecture:
* JWTs can contain roles and by definition cannot be updated (signed and immutable, live in the client)

dennishendriksen:
> When we start a job in the user's name it runs with a security context not registered in the session registry and it will not get updated.

This is not related to the issue reported in #4080. The fix in this PR shouldn't change the behavior for jobs.

> When multiple threads finish adding roles (think e.g. a commander script that triggers quick successive role membership updates), you get nasty race conditions that mean the end result may be incorrect

How to reproduce? Alternatively, can you point out which code change introduces the race conditions?

> Adding nonstandard security code is tricky, may introduce vulnerabilities that we cannot oversee.

I don't think this fix introduces non-standard security code. It is using code that we have been using in MOLGENIS for quiet some time with no known issues (keep in mind that #4080 is a regression). The existing code is based on solutions used by others to deal with the same issue.

> JWTs can contain roles and by definition cannot be updated (signed and immutable, live in the client)

MOLGENIS is not using any roles that might be stored in a JWT ...

molgenis-data-security/src/main/java/org/molgenis/data/security/SessionSecurityContextUpdaterImpl.java
```diff
@@ -0,0 +1,88 @@
...
+      usernames.clear();
+    }
+  }
+
+  @Override
+  public void doCleanupAfterCompletion(String transactionId) {
+    actionThreadLocal.remove();
+  }
+
+  private void resetAuthorities(String username) {
```
fdlk: I think that this may overlap with concurrent updates to the authorities by other threads

PR #8952, opened 2020-03-16
Feat/expose selected theme

connoratrug:
#### Checklist
- [x] Code reviewed
- [x] Clean commits


molgenis-core-ui/src/main/java/org/molgenis/core/ui/MolgenisWebAppConfig.java
```diff
@@ -255,6 +255,11 @@ public PropertiesMessageSource uiFormMessageSource() {
     return new PropertiesMessageSource("ui-form");
   }
 
+  @Bean
+  public PropertiesMessageSource coreUiMessageSource() {
+    return new PropertiesMessageSource("core-ui");
```
connoratrug: Should the framework have thrown an error when a referenced messageSource was not found ?

fdlk: It falls back to the more technical message format then.

PR #8960, opened 2020-03-31
chore: add default tomcat environment variables

fdlk:
#### Checklist
- Code reviewed
- User documentation updated
- Test plan template updated
- [x] Clean commits

### DIFF
--- a/molgenis-app/dev-env/intellij-tomcat-env
+++ b/molgenis-app/dev-env/intellij-tomcat-env
@@ -1,0 +1,12 @@
+environment=development
+opencpu.uri.host=localhost
+elasticsearch.transport.addresses=localhost:9300
+db_uri=jdbc:postgresql://localhost/molgenis
+db_user=molgenis
+db_password=molgenis
+admin.password=admin
+MINIO_BUCKET_NAME=molgenis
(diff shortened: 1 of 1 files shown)


molgenis-app/tomcat-env
```diff
@@ -0,0 +1,12 @@
+environment=production
+opencpu.uri.host=localhost
+elasticsearch.transport.addresses=localhost:9300
+db_uri=jdbc:postgresql://localhost/molgenis
+db_user=molgenis
+db_password=molgenis
+admin.password=admin
```
sidohaakma: There is a property file in the rpm directory with the same values. Maybe we need to make a choice how to add the properties to the deployment: https://github.com/molgenis/molgenis/blob/master/molgenis-app/rpm/templates/molgenis-server.properties. Or do we need 2?

fdlk: This is intended for configuration of intellij on a dev machine, so you need not manually twiddle with two different lists of env boxes. Sounds like I need a better name and/or move the file around a bit :D

fdlk: Better now?

sidohaakma: Ah right, to copy you mean! Nice.

PR #8973, opened 2020-05-12
M7343: As search user I want case-insensitive keyword search

fdlk:
In EMX, you can now tag ENUM, HYPERLINK, EMAIL and STRING attributes (i.e. max length 255) with predefined tags `token` and `case-sensitive`.

# no tag
Nothing changed

# token
The attribute is analyzed using the ElasticSearch keyword analyzer, with a normalizer that lowercases and ascii-folds the values.
* SEARCH on the field remains case-insensitive.
* SEARCH will only match the complete token, no partial matches

# token, case-sensitive
The attribute is analyzed using the ElasticSearch keyword analyzer, but no normalizer is applied.
* SEARCH on the field field becomes case-sensitive
* SEARCH will only match the complete token, no partial matches

# where can I see this
You can test this in the data explorer filters, those use search on the attribute if it is a string.
The top-left search box does a search on the _all field, this is unchanged.

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-app/dev-env/docker-compose.yml
+++ b/molgenis-app/dev-env/docker-compose.yml
@@ -7,7 +7,6 @@ services:
     container_name: frontend
     ports:
       - 80:80
-    network_mode: ${NETWORK_MODE:-bridge}
     volumes:
       - ./backend.conf.template:/etc/nginx/proxy.d/backend.conf.template
     command: >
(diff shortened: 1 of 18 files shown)


fdlk:
not so sure anymore who wants the case-sensitive set to true

PR #9013, opened 2020-05-29
[FEAT] Hikari connection pool

fdlk:
* Switch to Hikari CP
* Scale down connections when idle
* Add connection pool metrics

#### Checklist
- [x] Code reviewed
- User documentation updated
- Test plan template updated
- [x] Clean commits

### DIFF
--- a/molgenis-data/pom.xml
+++ b/molgenis-data/pom.xml
@@ -90,9 +90,8 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>com.mchange</groupId>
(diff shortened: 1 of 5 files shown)


sidohaakma:
Have you experienced the lightning fastness of the connection pool? ;-D.

fdlk:
I ran some tests and the difference is not so big, I think. I ended up reading https://github.com/brettwooldridge/HikariCP/wiki/%22My-benchmark-doesn't-show-a-difference.%22

PR #9036, opened 2020-06-25
fix: do not crash row-edit if data-explorer is not in menu

jvanveen:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-data-row-edit/src/main/resources/templates/view-data-row-edit.ftl
+++ b/molgenis-data-row-edit/src/main/resources/templates/view-data-row-edit.ftl
@@ -14,8 +14,8 @@
     window.__INITIAL_STATE__ = {
         baseUrl: '${baseUrl}',
         lng: '${lng}',
-        fallbackLng: '${fallbackLng}',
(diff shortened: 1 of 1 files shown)


molgenis-data-row-edit/src/main/resources/templates/view-data-row-edit.ftl
```diff
@@ -15,7 +15,7 @@
         baseUrl: '${baseUrl}',
         lng: '${lng}',
         fallbackLng: '${fallbackLng}',
-        dataExplorerBaseUrl: '${dataExplorerBaseUrl}'
+        dataExplorerBaseUrl: <#if dataExplorerBaseUrl??>\'${dataExplorerBaseUrl}\'<#else>null</#if>
```
connoratrug, line 18: dataExplorerBaseUrl is used in the data-row-edit see https://github.com/molgenis/molgenis-frontend/blob/d2b22c15f58fc9ac67e5473d0bec1107bc51b945/packages/data-row-edit/src/main.js#L29 
could still have it set to null , but maybe this leeds to strange behaviour, did you check ? 
It seems to be set as a catch all, so i could still be ok that in case of missing dataExplorerBaseUrl it jumps to application home path

PR #9089, opened 2020-09-22
Fix #9011 Updating one value via REST API v2 fails for reference attributes

### LINKED ISSUE #9011: REST api v2 updating one value fails if attribute is of data type xref/mref/categorical/categoricalmref
The REST api V2 functionality Update One Value for multiple IDs does work in case the attribute is of type int or string, but fails when the attribute is of type xref/categorical/mref/categoricalmref. The error messages that occurs is: "lazyEntity instead of String" 
When changing the input from for example a string value ('A') to a list ['A'] or dictionary {'A'} the errors are "ArrayList? instead of String" and "LinkedTreeMap? in stead of string".

FROM REST api V2 Documentation:

One value
If you only want to change one attribute without needing to provide all other attributes:
Request
```PUT ​http://molgenis.mydomain.example/api/v2/person/age
Content-Type: application/json
{
   entities: [
    {
      "id": 1,
      "age": "1"
    },
    {
      "id": 2,
      "age": "2"
    }
  ]
```
### How to Reproduce
With the attached EMX and python script the error can be reproduced.

### Expected behavior
No error is returned and the update is performed
### Observed behavior
An error is returned

### python and emx file to reproduce:
[issue_rest-api-v2-updating-one-xref-mref-categorical-categoricalmref.zip](https://github.com/molgenis/molgenis/files/4701063/issue_rest-api-v2-updating-one-xref-mref-categorical-categoricalmref.zip)

tommydeboer:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-api-data/src/main/java/org/molgenis/api/data/v2/RestControllerV2.java
+++ b/molgenis-api-data/src/main/java/org/molgenis/api/data/v2/RestControllerV2.java
@@ -103,6 +103,7 @@ import org.springframework.web.util.UriComponentsBuilder;
 @RequestMapping(RestControllerV2.BASE_URI)
 @Timed(value = "rest.v2", description = "Timing information for the REST API v2.", histogram = true)
 public class RestControllerV2 {
+
   private static final Logger LOG = LoggerFactory.getLogger(RestControllerV2.class);
 
   static final int MAX_ENTITIES = 1000;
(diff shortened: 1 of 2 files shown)


fdlk:
[Fix #9011](https://trello.com/c/kTDX931G/467-fix-9011)

fdlk (approved):
Nice catch!

PR #9116, opened 2020-10-29
feat: Show number of active user sessions in the user manager plugin

connoratrug:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-core-ui/src/main/java/org/molgenis/core/ui/admin/usermanager/UserManagerServiceImpl.java
+++ b/molgenis-core-ui/src/main/java/org/molgenis/core/ui/admin/usermanager/UserManagerServiceImpl.java
@@ -9,19 +9,19 @@ import java.util.stream.Stream;
 import org.molgenis.data.DataService;
 import org.molgenis.data.UnknownEntityException;
 import org.molgenis.data.security.auth.User;
+import org.molgenis.security.core.SecurityContextRegistry;
 import org.springframework.security.access.prepost.PreAuthorize;
-import org.springframework.security.core.session.SessionRegistry;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
 @Service
 public class UserManagerServiceImpl implements UserManagerService {
   private final DataService dataService;
-  private final SessionRegistry sessionRegistry;
+  private final SecurityContextRegistry securityContextRegistry;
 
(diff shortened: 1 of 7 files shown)


molgenis-security/src/main/java/org/molgenis/security/MolgenisWebAppSecurityConfig.java
```diff
@@ -137,6 +140,10 @@ protected void configure(HttpSecurity http) throws Exception {
     requestCache.setRequestMatcher(createWebRequestMatcher());
     http.requestCache().requestCache(requestCache);
 
+    http.sessionManagement()
+        .maximumSessions(MAXIMUM_CONCURRENT_USER_SESSIONS)
```
fdlk, line 144: That'll break concurrent use by functional / admin/admin accounts.

connoratrug: was wondering about that , why would it break ?

connoratrug: switched to using molgenis SecurityContextRegistry, removes need to set max sessions

molgenis-core-ui/src/main/java/org/molgenis/core/ui/admin/usermanager/UserManagerServiceImpl.java
```diff
@@ -41,6 +44,12 @@ public void setActivationUser(String userId, Boolean active) {
     this.dataService.update(USER, user);
   }
 
+  @Override
+  @PreAuthorize("hasAnyRole('ROLE_SU','ROLE_MANAGER')")
```
fdlk, line 48: This check will only go off on a server where a non-su non-manager user is given plugin permissions on the user manager. In that rare case it'll cause a nasty 500 error breaking the complete screen.
I'd say simply remove the check.

fdlk (changes requested):

molgenis-core-ui/src/main/java/org/molgenis/core/ui/admin/usermanager/UserManagerServiceImpl.java
fdlk: Nevermind, the rest of the methods have the same annotation.

fdlk (approved):

PR #9120, opened 2020-11-09
[FEAT] Show if user has active session

connoratrug:
Add a column to the user-manger users table indicating whether or not the user has a active session. 
 
#### Checklist
- [x] Code reviewed
- [x] Clean commits


molgenis-core-ui/src/main/java/org/molgenis/core/ui/admin/usermanager/UserManagerServiceImpl.java
```diff
@@ -50,6 +54,20 @@ public long getActiveSessionsCount() {
...
+  @Override
+  @PreAuthorize("hasAnyRole('ROLE_SU','ROLE_MANAGER')")
+  public List<String> getActiveSessionUserNames() {
+    return securityContextRegistry
+        .getSecurityContexts()
+        .map(SecurityContext::getAuthentication)
+        .filter(Objects::nonNull)
+        .map(Authentication::getPrincipal)
+        .filter(
+            principal -> principal instanceof org.springframework.security.core.userdetails.User)
```
fdlk, line 66: I think principal.getName is probably better, it's not safe to assume too much about the Authentication object

connoratrug: But what if principal is of different type but still has a getName ? , then the type check in the PR seems saver to me, as its skips it in case of unexpected type

connoratrug: getPrincipal returns a Object so there is no guarantee you get de a Principal with a getName

fdlk: Authentication is a principal, so you can call getName on it. They've done their best already to make sure they return the correct name.

connoratrug: Updated the PR as suggested, seems to work :). But whole thing depends on us using the Principal.name as userName. But i guess that kind of true for all implementations where we join the list of users with the list of authentications

fdlk (approved):

PR #9124, opened 2020-12-04
Fix #9078 ACEs not removed when deleting roles

### LINKED ISSUE #9078: ACL tables not updated when deleting a group (+ roles)
### How to Reproduce
1. Add a group _test_ 
2. Inspect the `acl_sid` table in the database:

|id|principal|sid|
|---|-----|-----------------|
|...|...|...|
|7  |true |admin            |
|8  |false|ROLE_test_MANAGER|
|9  |false|ROLE_test_EDITOR |
|10 |false|ROLE_test_VIEWER |

3. Give role _test_VIEWER_ read permission on the Metadata Manager
3. Delete the group
4. Create the group _test_ again
5. Inspect the Plugin permissions for the role _test_VIEWER_

### Expected behavior
1. The new group's role does not have permission on the Metadata Manager.
2. Deleting a group also deletes its roles and the associated sids are removed from the ACL tables.

### Observed behavior
1. The new group's role **does** have permission on the Metadata Manager, even though it's a new role.
2. The ACL tables were not updated and the sids are still there (before adding the new group):

|id|principal|sid|
|---|-----|-----------------|
|...|...|...|
|7  |true |admin            |
|8  |false|ROLE_test_MANAGER|
|9  |false|ROLE_test_EDITOR |
|10 |false|ROLE_test_VIEWER |

### ⚠️ Warning 
Because this is an old bug (molgenis 7-ish), it's very possible that there are now Molgenis instances that have a database with an inconsistent state. To be precise: every instance where a group was deleted without re-adding it again. No cause for great alarm yet, I think, but we should remember this.

tommydeboer:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-data-migrate/src/main/java/org/molgenis/data/migrate/bootstrap/MolgenisUpgradeBootstrapper.java
+++ b/molgenis-data-migrate/src/main/java/org/molgenis/data/migrate/bootstrap/MolgenisUpgradeBootstrapper.java
@@ -17,11 +17,13 @@ import org.molgenis.data.migrate.version.Step40AddRoleSystem;
 import org.molgenis.data.migrate.version.Step41Reindex;
 import org.molgenis.data.migrate.version.Step42RemoveFormsUrlRow;
 import org.molgenis.data.migrate.version.Step43SetUsernameAttributeName;
+import org.molgenis.data.migrate.version.Step44CascadeDeleteSids;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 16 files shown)


tommydeboer:
[Fix #9078 ACLs not removed when deleting a group](https://trello.com/c/eYHLSyTU/656-fix-9078-acls-not-removed-when-deleting-a-group)

fdlk (approved):

PR #9136, opened 2021-01-11
Fix #9135: Validation expressions in old forms no longer work

### LINKED ISSUE #9135: Validation expressions in old forms no longer work
### How to Reproduce
1. In the Data Explorer Settings, set the **Edit the data row using the forms edit plugin** setting to _False_
2. Go to an EntityType with validation expressions (for example **Attribute**)
3. Change a field with a validation expression on it (for example **Nillable**)

### Expected behavior
If the values are correct according to the expression, it shows no error.

### Observed behavior
As soon as you change the value, it will always show the error, even when you put it back as it was.

fdlk:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


jelmerveen (approved):
We do need to test a preview, but I am unable to deploy one

PR #9137, opened 2021-01-12
Fix #6360 Separate sign-up form and OIDC settings

### LINKED ISSUE #6360: Google sign-in only works as long as user signup is enabled and moderation is disabled
### How to Reproduce
Open app settings, enable google sign-in

### Expected behavior
* I can enable google sign-up in combination with moderation
* Accounts that have signed up using google can still log in even if I disable signup

### Observed behavior
* I can only allow google sign-in in combination with signup enabled and moderation disabled

tommydeboer:
_Note_: validation expression on the authentication settings doesn't work yet because of the issues with validation expressions.

#### Checklist
- [x] Code reviewed
- [x] Clean commits


molgenis-data-migrate/src/main/resources/step46-disableInactiveOidcClients.sql
```diff
@@ -0,0 +1,14 @@
+DO $$
+DECLARE
+    settings public."sys_set_auth#98c4c015"%ROWTYPE;
```
fdlk, line 3: help!

fdlk (changes requested):
Missing source code for Step46DisableInactiveOidcClients

fdlk (approved):

PR #9147, opened 2021-01-18
Feat: use bootstrap glyphicons in attribute tree

fdlk:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


connoratrug:
New icon look worse then old icons imo

also worse ux; icons do not seem actionable, no pointer hover action


Screenshot 2021-01-20 at 10 08 06

Suggest to close pr and keep as is, then replace with new version of data-explorer

PR #9157, opened 2021-02-05
feat(api-fair): name the blank nodes with the valueof the Entity ID

svituz:
In the api-fair module, when two properties point to the same xref object, the RDF representation creates two BNode that, as a matter of fact, represent the same object. 
For example, if for some reason the `fdp_Dataset` `rights` attribute is associated with two tags (e.g., `dct:rights`, `dct:accessRights`), the current behavior creates two blank nodes. 

```
:dataset_1 a dcat:Dataset, dcat:Resource;
    dct:rights [
        a dct:RightsStatement;
        dct:description "This resource has access restriction" .
    ];
    dct:accessRights [
        a dct:RightsStatement;
        dct:description "This resource has access restriction" .
    ].
```
This PR changes this giving the BNode a name corresponding to the entityId. In this way, the properties refer to the same object.
The above example in the new version is:

```
:dataset_1 a dcat:Dataset, dcat:Resource;
    dct:rights _:restricted;
    dct:accessRights _:restricted.

_:restricted a dct:RightsStatement;
    dct:description "This resource has access restriction" .
```

#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-api-fair/src/main/java/org/molgenis/api/fair/controller/EntityModelWriter.java
+++ b/molgenis-api-fair/src/main/java/org/molgenis/api/fair/controller/EntityModelWriter.java
@@ -264,7 +264,7 @@ public class EntityModelWriter {
               .toUriString();
       return valueFactory.createIRI(iri);
     }
-    return valueFactory.createBNode();
+    return valueFactory.createBNode(entity.getIdValue().toString());
   }
 
   UriComponentsBuilder getServletUriComponentsBuilder() {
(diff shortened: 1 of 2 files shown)


fdlk (changes requested):
Hi @svituz, excellent point, thanks for contributing!
Entity IDs are only unique within their entity type. So to create an ID that's guaranteed unique within the entire document you'll need to combine `entity.getEntityType().getId()` with `entity.getIdValue()`.

svituz:
Hi @fdlk, you're totally right. I changed it and added a specific test

fdlk (approved):
Works like a charm! Thanks for the contribution

PR #9163, opened 2021-02-15
Fix #9084 Dash not allowed in data api filters

### LINKED ISSUE #9084: data api v3 cant return data from entities that have a attribute with a minus character
### How to Reproduce

Give an entity an Attribute with a minus character in its name. like "intro-en".
Request the entity via data api v3.

See for example:
`https://master.dev.molgenis.org/@molgenis-ui/data-explorer/dist/index.html#/quest_nillable_report_header`
That tries to do a:
`https://master.dev.molgenis.org/api/data/quest_nillable_report_header?size=100&expand=&filter=id,intro,intro-en,intro-nl,intro-fr
`
### Expected behavior
API v2: no problem
API v3: no problem

### Observed behavior
API v2: no problem
API v3: "My god it burns, the world is on fire" status: 400 - Bad Request

tommydeboer:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/molgenis-api/src/main/resources/SelectionParser.jj
+++ b/molgenis-api/src/main/resources/SelectionParser.jj
@@ -22,11 +22,12 @@ TOKEN: {
 	| <#LETTER: ["a"-"z","A"-"Z"]>
   |	<#DIGIT: ["0"-"9"]>
   |	<UNDERSCORE: "_">
+  | <DASH: "-">
   |	<HASH: "#">
 }
 
 TOKEN: {
-		<IDENTIFIER: (<LETTER> | <DIGIT> | <UNDERSCORE> | <HASH>)+>
+		<IDENTIFIER: (<LETTER> | <DIGIT> | <UNDERSCORE> | <DASH> | <HASH>)+>
 }
 
 Selection parse() : {


fdlk (approved):

molgenis-api/src/main/resources/SelectionParser.jj
```diff
@@ -22,11 +22,12 @@ TOKEN: {
 	| <#LETTER: ["a"-"z","A"-"Z"]>
   |	<#DIGIT: ["0"-"9"]>
   |	<UNDERSCORE: "_">
+  | <DASH: "-">
   |	<HASH: "#">
 }
 
 TOKEN: {
-		<IDENTIFIER: (<LETTER> | <DIGIT> | <UNDERSCORE> | <HASH>)+>
+		<IDENTIFIER: (<LETTER> | <DIGIT> | <UNDERSCORE> | <DASH> | <HASH>)+>
```
connoratrug, line 30: i thought the dash was only allowed in-between other tokens

tommydeboer: @connoratrug Technically, a dash is only allowed at the end of an attribute name if it is followed by two letters. There are more attribute naming rules that are not being checked by the above expression: names can't start with an underscore or hash for example... 

Invalid names won't break anything, the API knows how to deal with them. That's why I decided to keep this change small.

PR #9179, opened 2021-03-09
feat/openid groups

fdlk:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


molgenis-api-identities/src/main/java/org/molgenis/api/identities/IdentitiesApiController.java
```diff
@@ -217,6 +229,89 @@ public void updateMember(
...
+    final String voGroupName = addMemberCommand.getVOGroup();
+    final String roleName = addMemberCommand.getRoleName();
+    final Role role = roleService.getRole(roleName);
+    final VOGroup voGroup = voGroupService.getGroup(voGroupName);
+
+    groupService.addVOGroupMembership(group, voGroup, role);
+
+    URI location =
+        ServletUriComponentsBuilder.fromCurrentRequest()
+            .path("/{voGroupName}")
```
fdlk, line 264: Mhmm... now the name ends up in the URL after all. Or use ID?

fdlk: Discussed with @sidohaakma: let's use the ID instead

sidohaakma: We do it now!

.gitignore
```diff
@@ -13,7 +13,6 @@ test-output/
 derby.log
 .metadata/
 .RData
-data
```
fdlk, line 16: oof!

sidohaakma (approved):

PR #9193, opened 2021-03-29
chore: build image using jib

fdlk:
Depends on https://github.com/molgenis/molgenis-ops-docker/pull/74
Building using jib means we bypass the docker socket, so need less privileges to build.
Also, it can pick up registry credentials from maven settings, so should become immune to the pull rate limit.
Finally, the image should become smarter with more cacheable layers

#### Checklist
- [x] Code reviewed
- User documentation updated
- Test plan template updated
- [x] Clean commits


molgenis-app/pom.xml
```diff
@@ -40,25 +38,38 @@
...
         </executions>
         <configuration>
-          <repository>${dockerfile.repository}</repository>
-          <tag>${dockerfile.tag}</tag>
-          <buildArgs>
-            <WAR_FILE>${dockerfile.warfile}</WAR_FILE>
-          </buildArgs>
-          <useMavenSettingsForAuth>true</useMavenSettingsForAuth>
+          <from>
+            <image>registry://registry.molgenis.org/molgenis/molgenis-base:PR-74</image>
```
fdlk, line 54: Is the base image introduced in https://github.com/molgenis/molgenis-ops-docker/pull/74

sidohaakma (approved):

PR #9204, opened 2021-04-13
feat: configure max file size as runtime property

fdlk:
Allows you to override max file upload size with system property `max.file.mb`.
The default remains 150MiB

In Tomcat:
* can be added to `JAVA_OPTS` or `CATALINA_OPTS` (`-Dmax.file.mb=512`)
* or specified in `catalina.properties`
* or as a parameter in `context.xml` 

In Docker:
* The container sets a context parameter from environment variable `MOLGENIS_MAX_FILE_MB`

In an app war:
* The default can still be overridden by subclassing the WebAppInitializer

#### Checklist
- [x] Code reviewed
- [NA] Migration step added in case of breaking change
- [x] Clean commits


tommydeboer (changes requested):
I think some documentation would be nice

tommydeboer:
As discussed with @fdlk: this is documented in docker-compose, molgenis/docker, etc.

tommydeboer (approved):

PR #9239, opened 2021-05-10
Fix #9237 Remove unnecessary synchronizations

### LINKED ISSUE #9237: Opening this table in DE2 completely breaks the application
### How to Reproduce
1. Import [genturis_v2.xlsx](https://github.com/molgenis/molgenis/files/6426245/genturis_v2.xlsx)
6. If needed, remove the DE1 from the menu
7. Go to the Navigator
8. Open `Radboud university medical center > Subject` 

### Expected behavior
Table opens.

### Observed behavior
The request times out and after that the whole application hangs indefinitely.

tommydeboer:
#### Checklist
- [ ] Code reviewed
- [ ] Clean commits

### DIFF
--- a/molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorRegistryImpl.java
+++ b/molgenis-data/src/main/java/org/molgenis/data/decorator/DynamicRepositoryDecoratorRegistryImpl.java
@@ -71,7 +71,7 @@ public class DynamicRepositoryDecoratorRegistryImpl implements DynamicRepository
    * repository.
    */
   @Override
-  public synchronized Repository<Entity> decorate(Repository<Entity> repository) {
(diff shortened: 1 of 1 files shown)


tommydeboer:
[Fix #9237 & #9240 Fix deadlock](https://trello.com/c/1X9haOMs/92-fix-9237-9240-fix-deadlock)

fdlk (changes requested):
The mutable part of the registries is the private HashMap, which is not thread-safe in the sense that you cannot read from it while it's being written.
Leaving the synchronized flag on the method that mutates the map means there won't be two concurrent writes. But we already saw they're not written concurrently.
So I'd think either trust our analysis and also remove the second synchronized flag. OR make it a ConcurrentHashMap.

fdlk (approved):

PR #9260, opened 2021-06-14
[FEAT] visible expression overrides nullable expression

fdlk:
Right now, the nullableExpression column must contain a negation of the visibleExpression to express that the value is nullable when the attribute is invisible. This is a lot of superfluous work to fill, so let's instead combine the values of both expressions in code.

This is how it works:

* The database only writes a `NOT NULL` constraint if the `nullable` boolean attribute is `false` (unchanged)
* In EMX, if `nullable` is set to `false`, and a `visibleExpression` is set, writes the `false` to the `nullableExpression` instead, so the value does not get a `NOT NULL` constraint. :new:
* When validating an attribute value, the nullable expression constraint is violated if
  * the value is `null` (unchanged)
  * the nullable expression is present and does not evaluate to `true` (unchanged)
  * not (the visible expression is present and evaluates to `false`) :new:
* If an attribute has value `nullable` (without the expression) set to `true` then
  * `nullableExpression` must be empty (unchanged) and
  * `visibleExpression` must be empty :new:

#### Checklist
- [x] Code reviewed
- [-] Migration step added in case of breaking change
- [x] Clean commits


tommydeboer:
image
The examples in the Metadata Manager still show Magma

tommydeboer (approved):

PR #9299, opened 2021-08-18
fix #9300: the menu will now calculate the space it needs 

### LINKED ISSUE #9300: Menu overlaps plugins when zoomed in or on small screen
### How to Reproduce
1. Go to the Data Explorer
2. Zoom in until contents of the menu don't fit on the screen anymore

### Expected behavior
The whole plugin is visible. I can still see the Data Explorer settings cogwheel ⚙️ 

### Observed behavior
image

MaxPostema:
#### Checklist
- [x] Code reviewed
- User documentation updated
- Test plan template updated
- [X] Clean commits
- [X] Added Feature/Fix to release notes

### DIFF
--- a/molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
+++ b/molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
@@ -137,22 +137,24 @@
     </#if>
 
 <#-- Start application content -->
-<script>
-    // Fix bootstrap 3 menu when the menu gets to big and needs 2 lines.
-    function calcHeaderHeight(){
-        let pageContext = document.getElementsByClassName("mg-page-content")[0];
-        let navBar = document.getElementsByClassName("navbar-fixed-top")[0];
-        let logoTopHeight = ${app_settings.logoTopMaxHeight};
-        let logoTopHref = ${app_settings.logoTopHref?js_string};
-        let height = 50;
+<#if (!version?? ||version == 1)>
+    <script>
+        // Fix bootstrap 3 menu when the menu gets to big and needs 2 lines.
+        function calcHeaderHeight(){
+            let pageContext = document.getElementsByClassName("mg-page-content")[0];
+            let navBar = document.getElementsByClassName("navbar-fixed-top")[0];
+            let logoTopHeight = ${app_settings.logoTopMaxHeight};
+            let logoTopHref = ${app_settings.logoTopHref?js_string};
+            let height = 50;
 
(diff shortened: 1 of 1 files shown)


molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
```diff
@@ -137,8 +137,27 @@
     </#if>
 
 <#-- Start application content -->
-<div class="container-fluid mg-page-content"
-     style="margin-top: <#if app_settings.logoTopHref?? && (!version?? ||version == 1)>${app_settings.logoTopMaxHeight + 50}<#else>50</#if>px;">
+<#if (!version?? ||version == 1)>
+    <script>
+        // Fix bootstrap 3 menu when the menu gets to big and needs 2 lines.
+        function calcHeaderHeight(){
+            let pageContext = document.getElementsByClassName("mg-page-content")[0];
```
jelmerveen, line 144: If you use var, it's also compatible with IE11

molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
```diff
@@ -137,8 +137,27 @@
...
+            let navBar = document.getElementsByClassName("navbar-fixed-top")[0];
+            let height = 50;
+
+            <#if app_settings.logoTopHref?has_content>
+                let logoTopHeight = ${app_settings.logoTopMaxHeight};
+                if(logoTopHref) height += logoTopHeight; // Correct height for header image
```
jelmerveen, line 150: Where does logoTopHref come from? should this be prefixed with app_settings. ?

molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
```diff
@@ -137,8 +137,27 @@
...
+            </#if>
+            
+            if(navBar.getBoundingClientRect().height>50) height += 50; // Correct height for double sized menu
+            pageContext.style.marginTop = height + "px";
+        }
+        window.addEventListener('resize', calcHeaderHeight);
```
jelmerveen, line 156: A good practise is to remove the event listener before adding it, to prevent multiple eventlisteners to be added for the same function (memory leak hazard)

molgenis-core-ui/src/main/resources/templates/molgenis-header.ftl
jelmerveen: Maybe add id selector instead, it's faster and more precise

jelmerveen (approved):

PR #9354, opened 2021-11-26
feat: endpoints for deleting sequences and getting the current value

tommydeboer:
#### Checklist
- [x] Code reviewed
- [x] Clean commits

### DIFF
--- a/docs/guide-pseudonymisation.md
+++ b/docs/guide-pseudonymisation.md
@@ -42,8 +42,8 @@ Here are the tags for the sequence we're looking at:
 | hasIDDigitCount7 | hasIDDigitCount | http://purl.obolibrary.org/obo/IAO_0000596 | 7     |
 | hasIDPrefixGen   | hasIDPrefix     | http://purl.obolibrary.org/obo/IAO_0000599 | GEN-  |
 
-Now it's time to add the tags to your id-attribute. You can do this in the Data Explorer (by going 
-to the Attribute entity type and editing your attribute), or you can make it part of the model in
+Now it's time to add the tags to your id-attribute. You can do this in the Data Explorer (by opening 
+the Attribute table and editing your attribute), or you can make it part of the model in
 your EMX file:
 
 | name      | dataType | nillable | idAttribute | tags                            |
--- a/docs/guide-pseudonymisation.md
+++ b/docs/guide-pseudonymisation.md
@@ -63,9 +63,19 @@ with a sequence there are two endpoints available:
 #### `GET <host>/plugin/metadata-manager/sequences`
 Returns the sequences present in the database.
 
+> Note: Keep in mind that the sequence identifiers contain number signs (#). These need to be URL 
+> escaped with `%23` in the upcoming endpoints.
+
+#### `GET <host>/plugin/metadata-manager/sequences/<sequence_id>`
+Returns the current value of the sequence.
+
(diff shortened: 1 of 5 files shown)


tommydeboer:
[Endpoints: delete sequence + get sequence value](https://trello.com/c/KnYvPAFD/455-endpoints-delete-sequence-get-sequence-value)

docs/guide-pseudonymisation.md
```diff
@@ -63,9 +63,19 @@ with a sequence there are two endpoints available:
 #### `GET <host>/plugin/metadata-manager/sequences`
 Returns the sequences present in the database.
 
+> Note: Keep in mind that the sequence identifiers contain number signs (#). These need to be URL
```
jelmerveen, line 66: why do we do this?

docs/guide-pseudonymisation.md
```diff
@@ -63,9 +63,19 @@ with a sequence there are two endpoints available:
...
+
+#### `GET <host>/plugin/metadata-manager/sequences/<sequence_id>`
+Returns the current value of the sequence.
+
+#### `DElETE <host>/plugin/metadata-manager/sequences/<sequence_id>`
+Deletes a sequence in the database. Since sequences are not automatically deleted when an entity type is deleted,
+you can use this endpoint to clean up. This endpoint can also be used to "reset" a sequence: a new sequence
+starting at 1 will be created as soon as a new row is added.
+
 #### `POST <host>/plugin/metadata-manager/sequences/<sequence_id>?value=123`
```
jelmerveen, line 77: make the title reflect the correct endpoint if # is enforced, so e.g value=%23123

molgenis-data-postgresql/src/test/java/org/molgenis/data/postgresql/JdbcSequencesTest.java
```diff
@@ -32,8 +33,50 @@ void testGetSequences() {
 
   @Test
   void testSetValue() {
-    jdbcSequences.setValue("abc", 123);
-    verify(jdbcOperations).queryForMap("SELECT setval(?, ?)", "abc", 123L);
+    when(jdbcOperations.queryForList("SELECT sequence_name FROM information_schema.sequences"))
+        .thenReturn(List.of(Map.of("sequence_name", "seq1")));
+    jdbcSequences.setValue("seq1", 123);
```
jelmerveen, line 38: how come value does not have a # here?

docs/guide-pseudonymisation.md
tommydeboer: PostgreSQL has a restriction on the length of table (and sequence) names, which is a maximum of 32 characters. We don't have this restriction on entity type identifiers and attribute names. So to make sure that every identifier fits in the database, we truncate the identifier and append a hash of the identifier to make sure that it's unique. In the case of the sequence identifiers, a hash of the entity type id + attribute name is used. The format we use is:

`<truncated_id>#<hash>`

jelmerveen: ah ok, maybe that is good to add then. max size 32 characters, else you need to url encode

jelmerveen (approved):

tommydeboer:
[[PID] Assign PIDs to existing biobanks](https://trello.com/c/71jeJOVZ/483-pid-assign-pids-to-existing-biobanks)

Issue #8505, opened 2019-05-20
Inconsistent search operator behaviour


marikaris:
### How to Reproduce
Import 
[test_hgvsc.csv.zip](https://github.com/molgenis/molgenis/files/3197293/test_hgvsc.csv.zip) via the quick importer
Go to dataexplorer
Go to your table
Type in the searchbar: NM_144670.4:c.483+7G>A
Underlying query is: `q=*=q=%27NM_144670.4:c.483%2B7G%3EA%27`
One variant appears
Now click hgvs_c in the filter tree and search for: NM_144670.4:c.483+7G>A
Underlying query is: `q=hgvs_c=q=%27NM_144670.4:c.483%2B7G%3EA%27`

### Expected behavior
Query `*=q=` gives same result as `hgvs_c=q=`

### Observed behavior
All (6) variants are shown

mswertz:
please add this example to the design spec for search improvements

dennishendriksen:
Assuming that you are looking for rows with ```hgvs_c``` attribute value being **exactly** ```NM_144670.4:c.483+7G>A``` then the wrong query operator is used. The correct operator here is ```==``` for an exact match: ```hgvs_c==NM_144670.4:c.483+7G>A```.

Besides this point I also think that it is unexpected for the ```*``` and ```hgvs_c``` to have different search results. This is what happens under the hood:

Elasticsearch query for ```q=*=q=NM_144670.4:c.483+7G>A```:
```json
  "match_phrase" : {
    "_all" : {
      "query" : "NM_144670.4:c.483+7G>A",
      "slop" : 10,
      "boost" : 1.0
    }
  }
```

Elasticsearch query for q=hgvs_c=q=NM_144670.4:c.483+7G>A:
```json
  "match" : {
    "hgvsc_b9d0a67a" : {
      "query" : "NM_144670.4:c.483+7G>A",
      "operator" : "OR",
      "prefix_length" : 0,
      "max_expansions" : 50,
      "fuzzy_transpositions" : true,
      "lenient" : false,
      "zero_terms_query" : "NONE",
      "boost" : 1.0
    }
  }
```

[...]

PR #9385, opened 2022-03-17
Fix #9386: introduce custom _all field for searchAll behaviour

### LINKED ISSUE #9386: Search-all slow, inconsistent with previous versions and in some cases completely broken
When using the search bar in the Data Explorer, generally you will find that response times have deteriorated. Some other issues include:

**BBMRI**: Searching for a biobank ID results in many many results instead of just the one biobank (like it did in 9.0)
**Chr6**: Searching anything will result in an error that the max query clause count has been reached (defaults to 1024). This value can be bumped by setting the `indices.query.bool.max_clause_count` property but in this case the query will take more than 30 seconds and it will time out.

Context: sinds ES 7 the `_all` field has been removed and we've replaced this feature by using multi-match queries.

tommydeboer:
In ElasticSearch 7 the `_all` field has been removed. In the upgrade to ES7 done in #8998, the search-all behaviour was refactored to use a multi-match query instead of the `_all` field. We have since found out that this does not perform well enough and in some cases even breaks (see issue #9386).

This PR reimplements the search-all behaviour by introducing a custom `_all` field. All other fields copy their values to this field using the `copy_to` parameter. Search behaviour should now be identical to pre-10.0 versions again. That means that issue #8505 will probably be reopened.

#### Checklist
- [x] Code reviewed
- User documentation updated
- Test plan template updated
- [x] Clean commits


tommydeboer:
[[Search] Implement _all field](https://trello.com/c/nm2Fak5H/684-search-implement-all-field)

PR #9408, opened 2023-03-13
feat(api-fair): adds management for ONE_TO_MANY data type when creating the RDF model

svituz:
Very simple PR that adds ONE_TO_MANY datatype handling when creating the RDF model.

It is handled as an MREF attribute

#### Checklist
- [ ] Code reviewed
- [ ] Clean commits


svituz:
Changed the previous implementation as agreed to have a v2 version that handles ONE_TO_MANY attributes serialization. In order not to change the old implementation, I copied the FAIRController and the EntityModelWriter to have a completely separated endpoint.

The new endpoint respond to requests at `/api/fdp/v2`

svituz:
This PR is not necessary anymore

